PR #58, opened 2013-10-26
Chocolatey Packaging

cbadke:
Add scripts to build chocolatey install packages. One package installs to the global modules location in c:\Windows the other installs to the installing user's Documents module location.

If this is accepted, I assumed that @lzybkr would want to upload and own the package on Chocolatey.org. If not, I can upload the packages if I notice a new release in Github.

I would also suggest that Githubs release functionality be used to store release binaries rather than checking in the built packages directly in the repo.


vorou:
@lzybkr did you abandon the idea? Every now and then I found myself typing in `cinst PSReadLine` before I remember there is no package.

gwojan:
@vorou Try using PowerShellGet from the WMF 5.0 preview. You can download the experimental builds of OneGet for Windows 7.

`Find-Module psreadline | Install-Module`

It works pretty well. :smile:

vorou:
I'm sure it works, but I'm also sure there are much more people who are already using chocolatey. And it seems weird to install a new package manager just to install ReadLine. I mean, it's OK to do it one time (if you only need to install it on a single box, you could go well w/ just downloading the file manually, after all) but it's a pain to repeat it on each box.

But since the issue isn't getting any attention, I guess I'm the only one with that problem.

@lzybkr are you still against the chocolatey? If not, and you just don't have the time/motivation, I could upload the package myself.

lzybkr:
Out of laziness, I'd prefer publishing in just one place.  With the introduct of the PowerShell gallery, I was somewhat compelled to publish in more than one place to help bootstrap/test PowerShellGet.

When OneGet/PowerShellGet has support for PowerShell V3/V4, I will consider publishing on the PowerShell gallery exclusively - partly out of laziness, partly to have a single set of instructions for installation.

Because OneGet support both chocolately and the PowerShell gallery, I'd really prefer not publishing a chocolately package because some people will see two packages and wonder which one is the official one.

PR #225, opened 2015-04-01
Add option to trim .\ prefix from directory and filename completions

gfody:
".\" being prepended to directory and file names was causing trouble for a lot of the msys utils that I use from powershell. I should wrap them all with scripts that use cygpath but for the majority that just take a single file or directory name this option to trim the ".\" from completion was more convenient.

It only trims if you didn't explicitly type .\, e.g:
trimdotslash

### DIFF
--- a/PSReadLine/Cmdlets.cs
+++ b/PSReadLine/Cmdlets.cs
@@ -90,6 +90,11 @@ namespace PSConsoleUtilities
         public const bool DefaultShowToolTips = false;
 
         /// <summary>
+        /// Trim "./" prefix when auto completing directories or files.
+        /// </summary>
+        public const bool DefaultTrimDotSlash = false;
+
+        /// <summary>
         /// When ringing the bell, what frequency do we use?
         /// </summary>
         public const int DefaultDingTone = 1221;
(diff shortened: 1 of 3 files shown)


lzybkr:
I'd probably accept something like this in TabExpansion++, but I'm not sure it belongs in PSReadline.

My thinking is - there are already too many options.  If we look at some *nix shells like zsh, I think it's easy to go too far in adding options like this and have a very hard to configure system.

Also, you can fully implement this functionality in TabExpansion++, there is nothing special w.r.t. command line editing here.  And if you add this to TabExpansion++, then it will work in other hosts like the ise.

I might be open to adding an to filter items using a ScriptBlock, similar to the AddToHistoryHandler - that would enable your scenario and others that haven't been considered yet while adding just one option, but still, this filter probably belongs in TabExpansion++ so it works in other hosts.

gfody:
Ah, I didn't know about TabExpansion++ I will check it out.

For this, I will remove the option and just have my fork behave this way - I only made it optional in the hopes that I could kick it upstream. This and one other change to replace backslashes with forward slashes whenever forward slashes are present (and backslashes are not) make ps very usable as an msys shell. I'll have a crack at implementing them both in TabExpansion++

PR #235, opened 2015-05-14
Adding in an F7 handler to bring up command history in a grid.

andrewducker:
See issue at https://github.com/lzybkr/PSReadLine/issues/233
(With many thanks to user "nightroman" for doing most of the heavy lifting!)

### DIFF
--- a/PSReadLine/SamplePSReadlineProfile.ps1
+++ b/PSReadLine/SamplePSReadlineProfile.ps1
@@ -22,9 +22,9 @@ Set-PSReadlineKeyHandler -Key DownArrow -Function HistorySearchForward
 # really easy to find a command you previously used.
 # Either press F7 to bring up your entire history, or type part of the 
 # command line and then press F7 to bring up just the commands that match.
-Set-PSReadlineKeyHandler -Key F7 
-                         -BriefDescription History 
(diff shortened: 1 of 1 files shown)


andrewducker:
Apologies - serves me right for editing the file directly in GitHub, rather than doing things properly!
Should now be fixed.

CollinChaffin:
Hi, forgive me but on multiple systems I have yet to ever see F7 bring up anything at all when psreadline is loaded.  It actually was driving me nuts for some time until one by one I went through disabling each module to determine that psreadline was removing the default F7 history popup.

So I'm confused I see older docs stating that F7 functionality was never changed, which certainly is not my experience, and now with 1.2 changlog a mention of this "new" F7 history "emulation".

So, with any version of psreadline can I have my F7 default history popup or similar?  If so, which version and how do I activate it?

lzybkr:
This PR is just a sample, you have to add the code to your profile if you want the key binding do anything.

At some point, I'm sure I will add it for real, but it won't use Out-GridView so that it can work purely in the console with no new window popping up.

nightroman:
+1 for native support. Out-GridView is a just workaround. But it shows how useful and needed such a feature. It saves a lot of time.

PR #255, opened 2015-06-25
Set-PSReadlineKeyHandler -Key F7: command history using Out-GridView.

nightroman:
- Properly reads multiline commands from the history file.
- Allows selection and insertion of several commands.


nightroman:
Ah, got it. "For a sample" is the key. Yes, I completely agree.

nightroman:
Submitted #257 with two tweaks, including the subtle output issue. If you want, please use `Write-Output`. I do not like to use it explicitly. It can be even redefined, in theory.

PR #344, opened 2016-02-11
Generate help from platyPS markdown

vors:


lzybkr:
Awesome, this will make it easy to update the help.  It also made it easy for me to see some updates are needed for some defaults that have changed.

PR #345, opened 2016-02-11
Add help output for the module as PSReadLine.txt artifact

vors:
I published platyPS v 0.3.0 which should fix Update-Help interactive problem https://ci.appveyor.com/project/lzybkr/psreadline/build/1.0.14


vors:
"Get-Help calling Update-Help and CI is non-interactive" problem is fixed in platyPS 0.3.1

PR #407, opened 2016-07-16
SamplePSReadlineProfile::SmartInsertQuote: don't put double quotes wh…

Laoujin:
…en there is an oneven amount of them on the current line

I always found this annoying... not sure if it is the case for most peoples workflow...

### DIFF
--- a/PSReadLine/SamplePSReadlineProfile.ps1
+++ b/PSReadLine/SamplePSReadlineProfile.ps1
@@ -127,7 +127,12 @@ Set-PSReadlineKeyHandler -Key '"',"'" `
     $cursor = $null
     [Microsoft.PowerShell.PSConsoleReadLine]::GetBufferState([ref]$line, [ref]$cursor)
 
-    if ($line[$cursor] -eq $key.KeyChar) {
+    $quoteNumber = Select-String -InputObject $line -Pattern $key.KeyChar -AllMatches
+    if ($quoteNumber.Matches.Count % 2 -eq 1) {
+        # Oneven amount of quotes, put just one quote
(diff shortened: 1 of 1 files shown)


lzybkr:
Indeed, it is annoying.  This is certainly simple - I had something a little more sophisticated in mind, but I never got around to it, so let's just go with this for now.

PSReadLine/SamplePSReadlineProfile.ps1
```diff
@@ -127,7 +127,12 @@ Set-PSReadlineKeyHandler -Key '"',"'" `
     $cursor = $null
     [Microsoft.PowerShell.PSConsoleReadLine]::GetBufferState([ref]$line, [ref]$cursor)
 
-    if ($line[$cursor] -eq $key.KeyChar) {
+    $quoteNumber = Select-String -InputObject $line -Pattern $key.KeyChar -AllMatches
+    if ($quoteNumber.Matches.Count % 2 -eq 1) {
+        # Oneven amount of quotes, put just one quote
+        [PSConsoleUtilities.PSConsoleReadline]::Insert($key.KeyChar)
```
rkeithhill, line 133: `PSConsoleUtilities` should be changed to `Microsoft.PowerShell`.  Also, Select-String is a bit overkill.   I think this should do:

``` powershell
    $quoteCount = @([regex]::Matches($line, $key.KeyChar)).Length
    if ($quoteCount % 2 -eq 1) {
        # On odd number of quotes, insert just one quote
        [Microsoft.PowerShell.PSConsoleReadline]::Insert($key.KeyChar)
```

PR #412, opened 2016-07-24
PSConsoleUtilities namespace renamed to Microsoft.PowerShell

Laoujin:
What kind of logic did you have in mind regarding the quote insertion?
I have something like it's "good enough" but I'm willing to tinker a bit more if you have any ideas for further improvements...


rkeithhill:
I  was thinking you could use something like `$quoteCount = @([regex]::Matches($line, $key.KeyChar)).Length` instead of the call to the Select-String command.

Laoujin:
I was talking about the quote insertion... :)

Not sure, why is the regex better than Select-String. My idea would be that even if Select-String is 10 times slower, we're still talking about a few milliseconds.

rkeithhill:
Yeah, the Select-String perf is down in the low 10's milliseconds so no need to change.

lzybkr:
Oh, the only "smart" thing I had in mind that differs from what you did was to not insert paired quotes when already inside a string, e.g. I want:

``` powershell
# A single single quote
"'"
# A single double quote
'"'
```

I might be special though, so maybe it's not worth implementing

PR #414, opened 2016-08-01
SampleCode QuoteInsertion: insert only a single quote when inside the other quotes

Laoujin:
well, you're not the only one: because it always doubled something like  `"this isn''t working"`, I had the quote insertion turned off for single quotes.

now, this implementation is also pretty naieve, the following won't work for example:
`git ci -m"this doesn't work" | echo '`

It thinks it shouldn't double the single quote because there is an orphan single quote before it. But that one is in between double quotes so it shouldn't take it into account.

Anyhow, I hope that that this implementation will be wrong in more "complex" scenarios only...


lzybkr:
I ended up writing something that I think is a bit better because it uses the tokens to decide when it's in a string, see https://github.com/lzybkr/PSReadLine/commit/c4325d097bd851df4dc909e930eb911e8a22f986

PR #432, opened 2016-08-28
removed reference to PSReadline.dll-help.xml in csproj file

SteveL-MSFT:
enables building without error in VS


lzybkr:
That's not the right change - the file is generated by PlatyPS - the fix should be to build that file.

SteveL-MSFT:
I see, so PlatyPS builds the xml from a md file.

PR #463, opened 2016-11-05
Remove PSReadline.dll-help.xml from csproj.

jimmehc:
PR #344 deleted this file, but didn't remove the reference from the
csproj, which broke building from VS.


jimmehc:
Oh, it's just generated now.  Nevermind...

PR #480, opened 2016-12-07
Use InvokeBuild to build

lzybkr:


vors:
choco support is gone completely?

lzybkr:
Publishing to 2 different places is annoying enough, I'm not publishing to choco too.

rkeithhill:
I don't believe chocolatey is cross-platform.  I've been pushing folks to grab modules from the PowerShell Gallery.

Issue #485, opened 2016-12-23
Incremental Search for MenuComplete


MVKozlov:
still can't use Menucomplete because it not have incremental search and **set** first available command  instead of _suggesting_ it

Is any plans for it since 2014 ?

lzybkr:
Most folks find `MenuComplete` plenty useful as is, but I agree that it can be improved. This is a side project for me, so if someone doesn't do it first, I'll get to it when I get to it.

MVKozlov:
I found dirty way to do Incremental search but it is need more efficient way for repeating MenuSelect call 
May be you can help with it ?
Here is the diff, just a few lines added
https://gist.github.com/MVKozlov/04253d87086298e90fecef5ba784c65e

PR #515, opened 2017-03-07
Added incremental MenuComplete support

MVKozlov:
#485 
Add new Option and support for Incremental MenuComplete (Default = $false for seamless intergration)

Mimic PowerTab behaviour on MenuComplete - Enter finishes completition, Space finishes completition and add a space after.

### DIFF
--- a/PSReadLine/Completion.cs
+++ b/PSReadLine/Completion.cs
@@ -492,6 +492,7 @@ namespace Microsoft.PowerShell
                     {
                         undo = true;
                         processingKeys = false;
+                        _visualSelectionCommandCount = 0;
                     }
                     else if (Options.IncrementalMenuComplete && (nextKey == Keys.Space || nextKey == Keys.Enter)) {
                         if (nextKey == Keys.Space)
@@ -507,12 +508,9 @@ namespace Microsoft.PowerShell
                         PrependQueuedKeys(nextKey);
                         processingKeys = false;
                         if (Options.IncrementalMenuComplete) {
-                            //TODO: here we need to request another MenuSelect call. Need more efficient way!
-                            // Dumb variant - add another Keys.CtrlSpace key to key queue
-                            if (nextKey != Keys.Backspace) {
-                                //_queuedKeys.Enqueue(Keys.CtrlSpace);
+                            //here we need to request another MenuSelect call.
+                            if (nextKey != Keys.Backspace)
                                 _InvokeMenuCompleteCounter = 2;
-                            }
                         }
                     }
 
--- a/PSReadLine/Completion.cs
+++ b/PSReadLine/Completion.cs
@@ -697,6 +697,8 @@ namespace Microsoft.PowerShell
                                     prependNextKey = true;
                                     // we exit loop with current completion up to cursor
                                     truncateCurrentCompletion = true;
+                                    if (userInitialCompletionLength == 0)
+                                        undo = true;
                                 }
                             }
                             else
@@ -740,7 +742,7 @@ namespace Microsoft.PowerShell
                         }
                         if (!processingKeys) // time to exit loop
                         {
-                            if (truncateCurrentCompletion)
+                            if (truncateCurrentCompletion && ! undo)
                             {
                                 CompletionResult r = new CompletionResult(matches[selectedItem].CompletionText.Substring(0, _current - completions.ReplacementIndex));
                                 DoReplacementForCompletion(r, completions);
@@ -757,7 +759,7 @@ namespace Microsoft.PowerShell
                                 _mark = savedUserMark;
                             }
                             // without render all key chords that just move cursor leave selection visible, but it can be wrong
-                            Render();
+                            if (!undo) Render();
                             if (prependNextKey)
                             {
                                 _current -= cursorAdjustment;
--- a/PSReadLine/Completion.cs
+++ b/PSReadLine/Completion.cs
@@ -452,13 +452,13 @@ namespace Microsoft.PowerShell
                 // Showing the menu may have scrolled the screen or moved the cursor, update initialY to reflect that.
                 _initialY -= (previousMenuTop - menuAreaTop);
                 PlaceCursor();
-                previousMenuTop = menuAreaTop;
 
                 int previousItem = selectedItem;
 
                 bool processingKeys = true;
                 while (processingKeys)
                 {
+                    previousMenuTop = menuAreaTop;
                     var nextKey = ReadKey();
                     if (nextKey == Keys.RightArrow)
                     {
--- a/PSReadLine/Completion.cs
+++ b/PSReadLine/Completion.cs
@@ -423,7 +423,9 @@ namespace Microsoft.PowerShell
                     {
                         var index = row + (DisplayRows * col);
                         if (index >= matches.Count)
+                        {
                             break;
+                        }
                         var match = matches[index];
                         if (match.CompletionText.Equals(currentCompletionText)) currentCompletionIndex = index;
                         var item = HandleNewlinesForPossibleCompletions(match.ListItemText);
@@ -461,7 +463,9 @@ namespace Microsoft.PowerShell
                 if (item.ListItemText.StartsWith(completionFilter, StringComparison.OrdinalIgnoreCase) ||
                     GetUnquotedText(item.CompletionText, consistentQuoting).StartsWith(completionFilter, StringComparison.OrdinalIgnoreCase)
                    )
+                {
                     result.Add(item);
+                }
             }
             return result;
         }
@@ -528,7 +532,9 @@ namespace Microsoft.PowerShell
                 bool ambiguous;
                 var userCompletionText = GetUnambiguousPrefix(matches, useCompletionText: true, ambiguous: out ambiguous);
                 if (userCompletionText.Length == 0)
+                {
                     userCompletionText = GetUnambiguousPrefix(matches, useCompletionText: false, ambiguous: out ambiguous);
+                }
                 // remove possible first quote
                 if (userCompletionText.Length > 0 &&
                     ( IsSingleQuote(userCompletionText[0])
@@ -562,7 +568,9 @@ namespace Microsoft.PowerShell
                             || IsDoubleQuote(matches[selectedItem].CompletionText[0])
                             )
                            )
+                        {
                             curPos++;
+                        }
                         // set mark to the end of UserCompletion but in real completion (because of .\ and so on)
                         _mark = completions.ReplacementIndex + curPos + userCompletionText.Length;
                         DoReplacementForCompletion(matches[selectedItem], completions);
@@ -583,7 +591,9 @@ namespace Microsoft.PowerShell
                         _initialY -= (previousMenuTop - menuAreaTop);
                         // return cursor in place if screen was scrolled
                         if (previousMenuTop != menuAreaTop)
+                        {
                             PlaceCursor();
+                        }
 
                         if (previousItem != -1 && blanklinesCount > 0)
                         {
@@ -623,7 +633,9 @@ namespace Microsoft.PowerShell
                         // Search for possible unAmbiguous common prefix. ...
                         string unAmbiguousText = GetUnambiguousPrefix(matches, useCompletionText: true, ambiguous: out ambiguous);
                         if (unAmbiguousText.Length == 0)
+                        {
                             unAmbiguousText = GetUnambiguousPrefix(matches, useCompletionText: false, ambiguous: out ambiguous);
+                        }
                         int userComplPos = unAmbiguousText.IndexOf(userCompletionText, StringComparison.OrdinalIgnoreCase);
                         // ... If found - advance IncrementalCompletion ...
                         if (unAmbiguousText.Length > 0 && userComplPos >= 0 && unAmbiguousText.Length > (userComplPos + userCompletionText.Length))
@@ -698,7 +710,9 @@ namespace Microsoft.PowerShell
                                     // we exit loop with current completion up to cursor
                                     truncateCurrentCompletion = true;
                                     if (userInitialCompletionLength == 0)
+                                    {
                                         undo = true;
+                                    }
                                 }
                             }
                             else
@@ -726,7 +740,9 @@ namespace Microsoft.PowerShell
                                     // we exit loop with current completion up to cursor
                                     truncateCurrentCompletion = true;
                                     if (userInitialCompletionLength == 0)
+                                    {
                                         undo = true;
+                                    }
                                 }
                             }
                         }
@@ -761,7 +777,10 @@ namespace Microsoft.PowerShell
                                 _mark = savedUserMark;
                             }
                             // without render all key chords that just move cursor leave selection visible, but it can be wrong
-                            if (!undo && !keepSelection) Render();
+                            if (!undo && !keepSelection)
+                            {
+                                Render();
+                            }
                             if (prependNextKey)
                             {
                                 _current -= cursorAdjustment;
--- a/PSReadLine/Completion.cs
+++ b/PSReadLine/Completion.cs
@@ -439,6 +439,7 @@ namespace Microsoft.PowerShell
                 _visualSelectionCommandCount += 1;
 
                 string userCompletionText = _buffer.ToString().Substring(completions.ReplacementIndex, _current - completions.ReplacementIndex);
+                int userInitialCompletionLength = userCompletionText.Length;
                 DoReplacementForCompletion(matches[0], completions);
 
                 // Recompute end of buffer coordinates as the replacement could have
@@ -518,7 +519,7 @@ namespace Microsoft.PowerShell
                         _visualSelectionCommandCount = 0;
                         _singleton._mark = savedUserMark;
                     }
-                    else if (nextKey == Keys.Backspace)
+                    else if (nextKey == Keys.Delete)
                     {
                         // Esc alternative:
                         // stay string beginning as in current replacement but don't replace to full string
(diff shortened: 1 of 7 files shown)


lzybkr (changes requested):
I think this is a great proof of concept.

I don't see why this is an option - I've tried to avoid adding unnecessary options, and incremental feels like it's something everybody would want, nobody would turn it off.

Using the point and mark feels wrong - that's a user's bookmark.

Most importantly - invoking `MenuComplete` again is probably not acceptable. It might feel fast enough in some cases, but re-invoking tab completion on each incremental update will be painful in some cases.

MVKozlov:
This is an option mostly because it change behavoiur of Enter key - it can be a pain for other users ( not me )

I use point and mark because it a common way to suggest things in other editors and because every cursor move drop this mark, but I agree that saving it before and restoring after will be a good thing. btw, now this mark dropped on every completition

and yes, I still can't feel that I fully understand how Tab Completition work and I decided: lets PSReadLine think for me :) I will see what I can do with it with minimal changes

MVKozlov:
Still leave Option intact because of reasons above. If other changes will be approved it can be removed faster than inserted :)

Still use mark for suggestions because can't find how original version can use it, Marking always dropped after using menu.

Now CompletionMatches filtered by remove because of strange error with CommandCompletion constructor. This code stay commented out.

MVKozlov:
@lzybkr , rfc for latest commit :)
mostly I want to know your opinions about commented out code with `new CommandCompletion()`, why the compiler say 'Invalid Constructor'
the powershell can create such object
```
D:\> $filtered = New-Object 'System.Collections.ObjectModel.Collection[System.Management.Automation.CompletionResult]'
D:\> $compl = New-Object System.Management.Automation.CommandCompletion $filtered,0,0,10
D:\> $compl
CurrentMatchIndex ReplacementIndex ReplacementLength CompletionMatches
---------------- ---------------- ----------------- -----------------
               0                0                10 {}
D:\>
```

lzybkr:
Sorry, I've been pretty busy and haven't had a chance to look closely at this yet.

If you are getting a compiler error, it's most likely because I build against the PowerShell V3 reference assemblies and a new constructor was added since then.

I will consider dropping V3 support in the next release though - because I want to make the module portable between Windows PowerShell and PowerShell Core - and that means I'll need to require .Net 4.5 at a minimum.  Windows PowerShell V3 does not require .Net 4.5, so it might be confusing if PSReadline started requiring a newer runtime.

MVKozlov:
latest commit use CopyTo() instead of building new CommandCompletion()
this is more universal and does not use expensive Remove()
may be tomorrow I try to save/restore user mark
and what with option ? remove it ?

lzybkr:
I strongly prefer no option - so we'll refine the experience so it feels natural.

MVKozlov:
here it is

MVKozlov:
There still problems with keys like **Home/End/PgUp/PgDn**.

I don't know, may be I need to rework current concept into handling all keys inside `PossibleCompletionsImpl()` without leaving until **Enter/Space/Esc** pressed and so without handling `dispatchTable` handlers ?

MVKozlov:
I hope now I foresaw all the cases with completions filtering...

MVKozlov:
@lzybkr , please tell what story lies behind `consistentQuoting` code in `Completion.cs` around line 150+
It does not work for the case when `&` inserted before CompletionText
``` powershell
PS C:\Users\User> $r = [System.Management.Automation.CommandCompletion]::CompleteInput('.\nt',1,$null)
>> $r.CompletionMatches
CompletionText                                                                                     ListItemText
--------------                                                                                     ------------
.\NTUSER.DAT                                                                                       NTUSER.DAT
& '.\NTUSER.DAT{d51862bb-b663-11e6-9335-dfe3e1845d6f}.TM.blf'                                      NTUSER.DAT{d51862...
.\ntuser.ini                                                                                       ntuser.ini
```
I would like to add similar feature inside IncrementalCompletion (Tab advance IncrementalCompletion to the longest unambiguous prefix) and reuse the code but not fully understanded why it was implemented in a such way

lzybkr:
IIRC, the scenario was `cd c:\Pro<TAB>`.

Without that code, the unambiguous prefix was the empty string because a directory `C:\ProgramData` which did not need quotes, and `C:\Program Files` which did.

It seems I missed the case where `&` is also added when the completion is a command name.

MVKozlov:
I see but why you calculate `consistentQuoting` itself  instead of just
`var replacementText = GetUnquotedText(firstResult.CompletionText, false);` ?

some speedup ?

what if I correct `IsQuoted()` for `& "` case as I did in my code block where
I filter matches `.StartsWith("& ")` ? 

btw, for example above `ResultType` is `ProviderItem`

MVKozlov:
@lzybkr , while I improving my implementation I discovered incorrect menu displaying with completions text longer than buffer width. For example long file names. 
Among bug in `ShowTooltips` code branch which I fixed there is weird menu display in both modes and screen pollution.

What your opinions on my code ?

Should I try to fix menu displaying thru shortening text of menu items(` ...` in the middle) or thru recalculating menu rows count ?

First option shorten file names and can make it unreadable and second affects on menu items count that can be displayed with active menu

lzybkr:
Sorry I haven't been too responsive lately, just busy on other things. I also have some significant work in progress on PSReadline to rewrite the rendering so it's fast on Linux and handles resized windows correctly, and I'm avoiding thinking about how your changes will impact that work.

I just built your changes and tried it for the first time, great first impression, I'm now very motivated to get this merged.

Regarding the consistent quotes code - I don't remember specifics - the important part is the user experience. I'll play with your changes and if I don't notice problems, we'll call it good.

I'm totally open to big changes to the `ShowToolTips` code path - what I have now was quick and easy to see if it was useful. I think it's probably not too useful because the tip gets truncated or worse too often.

Showing the tooltip for just the current selection either above or below the menu is probably worth trying - maybe with a different background. Then we would always show the tooltip or possibly after a short delay like what happens in editors like the ise.

MVKozlov:
First of all I'm interested in the opinion of processing long names, what to do with them. Tips for long names now do not cause a fall - and this is the main thing :)

lzybkr:
After playing with what you have right now, truncation is promising. The menu doesn't crash and is somewhat useless, but because you can see the expanded filename for the selected item, it actually works just fine.
So I'd go with truncation in the menu - it seems the simplest to implement, and should be a fine experience too.

PSReadLine/Completion.cs
```diff
@@ -483,30 +592,117 @@ private void PossibleCompletionsImpl(CommandCompletion completions, bool menuSel
...
+                        }
                         processingKeys = false;
+                        _visualSelectionCommandCount = 0;
+                        _singleton._mark = savedUserMark;
+                        Render();
                     }
-
-                    if (selectedItem != previousItem)
+                    else if (nextKey == Keys.Space || nextKey == Keys.Enter ||
+                        (doneCompletionKeys.ContainsKey(matches[selectedItem].ResultType) &&
```
lzybkr: I'm building with C# 7 now, so maybe introduce a local function so you can avoid the ContainsKey call and just use TryGetValue instead.

PSReadLine/Completion.cs
```diff
@@ -429,29 +458,87 @@ private void PossibleCompletionsImpl(CommandCompletion completions, bool menuSel
                 int selectedItem = 0;
                 bool undo = false;
 
-                DoReplacementForCompletion(matches[0], completions);
+                int savedUserMark = _singleton._mark;
```
lzybkr: This is a non-static method, so you should use `_mark` instead of `_singleton._mark`. There are 3 other places in this method that should be changed as well.

PSReadLine/Completion.cs
```diff
@@ -483,30 +592,117 @@ private void PossibleCompletionsImpl(CommandCompletion completions, bool menuSel
...
+                            int cursorAdjustment = 0;
+                            if (matches[selectedItem].ResultType == CompletionResultType.ProviderContainer)
+                                userCompletionText = GetReplacementTextForDirectory(matches[selectedItem].CompletionText, ref cursorAdjustment);
+                            _current -= cursorAdjustment;
+                            PrependQueuedKeys(nextKey);
+                        }
+                    }
+                    else if (nextKey == Keys.Backspace || nextKey.KeyChar != 0)
+                    {
+                        string replacementText = matches[selectedItem].CompletionText;
```
lzybkr: `replacementText` is unused.

PSReadLine/Completion.cs
```diff
@@ -467,9 +554,31 @@ private void PossibleCompletionsImpl(CommandCompletion completions, bool menuSel
...
+                        selectedItem = Math.Min(selectedItem + displayRows - (selectedItem % displayRows) - 1, matches.Count - 1);
+                    }
+                    else if (nextKey == Keys.PageUp)
+                    {
+                        selectedItem = Math.Max(selectedItem - (selectedItem % displayRows), 0);
+                    }
                     else if (nextKey == Keys.Tab)
                     {
-                        selectedItem = (selectedItem + 1) % matches.Count;
+                        string unAmbiguousText = GetUnambiguousPrefix(matches, true, out ambiguous);
```
lzybkr: There are couple more `true` and `false` parameters that should use named parameters like is done elsewhere.

Could you also add comments to this code? I'm not sure what's going on, and whatever the intent, I'm confused because I don't see similar code in the `Shift+Tab` case when if feels like the inverse probably needs similar treatment.

I am wondering if you're trying to fix a long-standing bug where I show two identical items in the menu but they have different replacement text. If that's the case, this doesn't feel like the right fix.

lzybkr (changes requested):
I think I've looked pretty closely at all your changes, it looks pretty good and mostly works as I'd expect.

I did hit a small usability issue - when I type something that doesn't match, I wonder if we should assume that's what they wanted. That's how Intellisense works - instead of selecting something from the menu, then we just insert the character and forget completion.

MVKozlov:
Thanks for your comments, I'll try to implement it on the next week, I think, right after shortening long names.
btw, How you think, in current implementation, may be I should turn off Tooltips when it will be shorter than 5 symbols ? they are still useless

I think it is need to change code branch from `if (showToolTips)`  to
`if (showTooltips || displayColumns  == 1) `. this make things much simpler

MVKozlov:
> I did hit a small usability issue - when I type something that doesn't match, I wonder if we should assume that's what they wanted.
> Before your changes, an unexpected key accepted the current selection and was applied as though the key was pressed after the selection completed

>> if (nextKey == Keys.Delete)

>I'm not sure this case is a good idea.

my way simplifies `_mark` restoration :)
I try to implement it other way, more closely to previous behavior. But I want to keep 'Delete' code, may be on the other chord, because sometimes you already typed loooong command and want to keep it with small change and automatically use <tab> completion... you can type space+backspace, but it happen so often...
- ...time passed...
I think, if I restore old `any unused key` trigger 'leave menu' behavior it will not be necessary anymore

MVKozlov:
>When completing methods, maybe we should accept '(' as an ending character.

This block widely open for additions.

MVKozlov:
>Maybe I'm not normal, but I use End a lot when I'm done with the menu

Yes, after implementing this I also realize than I use `End` more often to navigate to end of line, than to end of menu :)

MVKozlov:
>I'm building with C# 7 now, so maybe introduce a local function so you can avoid the ContainsKey call and just use TryGetValue instead.

I'm not sure I'm familiar with these changes. Can you show some sample anywhere

MVKozlov:
>>                //consider possible '& ' prefix 

>A space is not required, e.g. I could have typed &C:\Pro<TAB>. I doubt people do that much though.

No, `'& '` can appear only in CompletionMatches text. and it always have space. If user typed it yourself it not counted in completions because it is previous AST object
``` powershell
PS C:\> $r = [System.Management.Automation.CommandCompletion]::CompleteInput('C:\Pro',6,$null)
PS C:\> $r.CompletionMatches

CompletionText             ListItemText               ResultType ToolTip
--------------             ------------               ---------- -------
& 'C:\Program Files'       Program Files       ProviderContainer C:\Program Files
& 'C:\Program Files (x86)' Program Files (x86) ProviderContainer C:\Program Files (x86)
C:\ProgramData             ProgramData         ProviderContainer C:\ProgramData


PS C:\> $r = [System.Management.Automation.CommandCompletion]::CompleteInput('&C:\Pro',7,$null)
PS C:\> $r.CompletionMatches

CompletionText           ListItemText               ResultType ToolTip
--------------           ------------               ---------- -------
'C:\Program Files'       Program Files       ProviderContainer C:\Program Files
'C:\Program Files (x86)' Program Files (x86) ProviderContainer C:\Program Files (x86)
C:\ProgramData           ProgramData         ProviderContainer C:\ProgramData


PS C:\> $r = [System.Management.Automation.CommandCompletion]::CompleteInput('&"C:\Pro',8,$null)
PS C:\> $r.CompletionMatches

CompletionText           ListItemText               ResultType ToolTip
--------------           ------------               ---------- -------
"C:\Program Files"       Program Files       ProviderContainer C:\Program Files
"C:\Program Files (x86)" Program Files (x86) ProviderContainer C:\Program Files (x86)
"C:\ProgramData"         ProgramData         ProviderContainer C:\ProgramData
```

MVKozlov:
>>else if (nextKey == Keys.Tab) 

> Could you also add comments to this code? I'm not sure what's going on, and whatever the intent, I'm confused because I don't see similar code in the Shift+Tab case when if feels like the inverse probably needs similar treatment.

I try to mimic `<Tab>` behavior like in `CompleteImpl()` when first completes until first common prefix and next switch thru completions
So I can for example type `Get-ProcessOwner` like `Get-Pr<tab>oc<Tab>o<space>`
( the real use case is `Get-MailBox[more..and..more..postfixes]` )

I doesn't implement it in `<ShiftTab>` because it never used in a such way

>I am wondering if you're trying to fix a long-standing bug where I show two identical items in the menu but they have different replacement text. If that's the case, this doesn't feel like the right fix.

Not sure which bug you mean
Now It works when Completions constructed from different sources like commands from different modules:
``` powershell
PS C:\> $r = [System.Management.Automation.CommandCompletion]::CompleteInput('get-priv',1,$null)
PS C:\> $r.CompletionMatches

CompletionText     ListItemText  ResultType ToolTip
--------------     ------------  ---------- -------
Get-Privilege      Get-Privilege    Command ...
Pscx\Get-Privilege Get-Privilege    Command Get-Privilege...
```
or for cases when user type some text and completions have this text not from the beginning of a line
``` powershell

PS C:\> $r = [System.Management.Automation.CommandCompletion]::CompleteInput('pro',1,$null)
PS C:\> $r.CompletionMatches

CompletionText                 ListItemText               ResultType ToolTip
--------------                 ------------               ---------- -------
& '.\Program Files'            Program Files       ProviderContainer C:\Program Files
& '.\Program Files (x86)'      Program Files (x86) ProviderContainer C:\Program Files (x86)
.\ProgramData                  ProgramData         ProviderContainer C:\ProgramData
```
now it have small bug with cursor placement

MVKozlov:
@lzybkr, I think that's all I can do without your attention.

lzybkr:
Thanks for the update - I've been using your latest today and so far it seems pretty good. I'll review the code more closely when I get a chance.

lzybkr:
I might have hit a bug in your changes - see if you can repro:
```
Exception:
System.Management.Automation.PSArgumentNullException: Cannot process argument because the value of argument "completionText" is null. Change the value of argument "completionText" to a non-null value.
   at System.Management.Automation.CompletionResult..ctor(String completionText, String listItemText, CompletionResultType resultType, String toolTip)
   at Microsoft.PowerShell.PSConsoleReadLine.PossibleCompletionsImpl(CommandCompletion completions, Boolean menuSelect)
```

MVKozlov:
The single line where `new CompletionResult()` created - its line 745.
and I can repro this:  `Get-Process <Tab><Backspace><Backspace>`
It happen when you exit from Completion loop through  double `<backspace>` with empty completion.

I think it can be fixed but fix will be not easy as it should :)
In this case I call `DoReplacementForCompletion(replacement, completions)` with empty replacement text. And seems it can't be created by standard method. May be it is need to write overload for `DoReplacementForCompletion` with first parameter as string or make direct call to `Replace()` (it exiting loop anyway)

MVKozlov:
After I slept with this thought, I realized that for the user its look like undo, so I make it through undo.

PSReadLine/Completion.cs
```diff
@@ -725,6 +725,8 @@ private void PossibleCompletionsImpl(CommandCompletion completions, bool menuSel
                                     prependNextKey = true;
                                     // we exit loop with current completion up to cursor
                                     truncateCurrentCompletion = true;
+                                    if (userInitialCompletionLength == 0)
+                                        undo = true;
```
lzybkr: Please use braces around the body.

PSReadLine/Completion.cs
```diff
@@ -697,6 +697,8 @@ private void PossibleCompletionsImpl(CommandCompletion completions, bool menuSel
...
```
lzybkr: Please use braces around the body.

PSReadLine/Completion.cs
```diff
@@ -757,7 +761,7 @@ private void PossibleCompletionsImpl(CommandCompletion completions, bool menuSel
                                 _mark = savedUserMark;
                             }
                             // without render all key chords that just move cursor leave selection visible, but it can be wrong
-                            Render();
+                            if (!undo && !keepSelection) Render();
```
lzybkr: I'm not always consistent about single line if statements, but in this case, I'd prefer braces and the body not on the same line as the condition.

lzybkr:
Thanks for all the fixes, it's probably getting close to be ready to merge.

I hit another minor issue, it's a little weird. The cursor is moving outside the command line.

To reproduce:
```
PS> 1..100 # Doesn't reproduce at the top of the screen
PS> Set-PSReadlineOption -EditMode Emacs
PS> [pscred<TAB><CTRL+SPACE>
```

When you've done this, the screen will look like:

```
PS C:\Users\jason> [pscred
PSCredential         PSCredentialTypes      PSCredentialUIOptions
PS C:\Users\jason> [pscredential
PSCredential         PSCredentia**l**Types      PSCredentialUIOptions
```

And the cursor will be on the `l` that I added `**` around. (I'd paste a screen shot, but the cursor disappears when the window doesn't have focus, and any keybinding I try to get the screen shot also changes the focus to lose the cursor.

I don't see any real problems with this bug, but it's weird enough I'd like it fixed.

MVKozlov:
I see. The original version have similar bug with end of buffer.
I tried to fix it but seems broke other case :)

MVKozlov:
What you think about double call to `GetUnambiguousPrefix()` ?
If you accept it as solution for `PS C:\>prog<<CTRL+SPACE>` cases may be I should do it inside method itself and eliminate double call ?

It change nothing in my code but extend `Unambiguous` search style to `CompleteImpl()` method (Emacs <Tab> completion, line 202 )

For example if I have **program.exe** somewhere in search path, `CompleteImpl()` doesn't find _program_ prefix as ambiguous but `MenuComplete` find it and extend _prog_ to _program_

lzybkr:
That does sound like an improvement - I think I'm fine with that.

MVKozlov:
How it works now
``` powershell
PS C:\> Set-PSReadlineOption -EditMode Emacs
PS C:\> prog<tab> #Ding here and line changed to
PS C:\> Program
PS C:\> #Second <tab>
Program Files        - C:\Program Files
Program Files (x86)  - C:\Program Files (x86)
ProgramData          - C:\ProgramData
program.exe          - C:\Windows\system32\program.exe
```
the only thing that can be named 'problem' is that first letter case changes to upper

lzybkr (approved):

lzybkr:
I think this is ready to merge, but I'd like the history cleaned up a bit.

If you're fine with me squashing and editing the commit message as I see fit, then I'll go ahead and do that, but I thought I'd give you a chance to do that before I merge.

If you feel like adding some tests, that'd be nice too, but I'd merge without tests, this functionality will be heavily used so regressions would be quickly noticed.

MVKozlov:
I'm fine with commit messages editing.
Basically they are needed in the process of coding and not all have any long-term meaning, especially early.

About the tests, to be honest, I do not know how to write them correctly in this case.

MVKozlov:
about last fix:
``` powershell
PS C:\>Get-PR<tab>
Get-Privilege
...
Get-Process
...
PScx\Get-Privilege
```
with 'optimization', ListItemText comparing starts on last item, and return `Get-Pri` prefix

lzybkr:
Thanks so much for your contribution and your patience with me, this will be a popular feature.

I don't have many completion tests (partly because it's a bit painful), see https://github.com/lzybkr/PSReadLine/blob/master/UnitTestPSReadLine/CompletionTest.cs. There is a useful sort of dsl to make writing tests easier, but it's still tedious for the menu completion. I'll gladly accept a future PR with just tests if you're so inclined.

MVKozlov:
@lzybkr , I tried to write a small test, but I was stuck with exception `Shouldn't call ReadKey when there are no more keys`, seems, because of menu cycle. I should mock ReadKey inside it, but it is beyond my current abilities :)
I never wrote such tests before.
If you show me how can I move forward, I will add some tests, which I'm doing manually now.

PR #532, opened 2017-04-22
Honor HistorySearchCursorMovesToEnd in Vi mode

srdubya:
Multiple fixes to issue.

### DIFF
--- a/PSReadLine/History.cs
+++ b/PSReadLine/History.cs
@@ -315,7 +315,11 @@ namespace Microsoft.PowerShell
             _buffer.Append(line);
             if (moveCursor)
             {
-                _current = _options.EditMode == EditMode.Vi ? 0 : Math.Max(0, _buffer.Length + ViEndOfLineFactor);
+                _current = Math.Max(0, _buffer.Length + ViEndOfLineFactor);
+            }
+            else if (_options.EditMode == EditMode.Vi)
+            {
+                _current = 0;
             }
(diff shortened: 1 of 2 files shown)


lzybkr:
This PR looks odd - I think you need to rebase on master and the extra commits should go away.

srdubya:
I tried rebasing, but I'm not sure it helped.

lzybkr:
I see the problem - you have merge commits in your master - one on Oct 22 (5aee6c7) and another on Apr 22 (60ff7a5) - if you get rid of those, this will merge cleanly without any extra work on my part.

`git rebase -i` is useful in this situation, but if there is just this one commit you want in the PR, this might be easier:

```
# Starting from a clean slate
git clone https://github.com/srdubya/PSReadLine.git psreadline_tmp
cd psreadline_tmp
git remote add upstream https://github.com/lzybkr/PSReadLine.git
git fetch upstream
git checkout upstream/master
git checkout -b fix_history
git cherry-pick 9c56805
git branch -d master
git checkout -b master
```

Once you've verified everything is correct, you can run `git push -f origin` - this will rewrite your history and it will match upstream w/ your one commit.

srdubya:
OK, I followed your "clean slate" process, thank you.  How does it look now?

lzybkr:
Much better. In the future, try to avoid merges in `master` - an easy way to do that is to always create a branch for any fix/feature work.

PR #538, opened 2017-05-30
change behavior for swapping

andrewcromwell:
SwapCharacters now swaps two characters, as described [here](https://cnswww.cns.cwru.edu/php/chet/readline/readline.html#SEC16). In other words, it also works at the end of a line.

Useful for fixing typos. Saves a keystroke: I can CTRL-T, instead of CTRL-B, CTRL-T.


lzybkr:
It looks like @srdubya added SwapCharacters - so I'd like his review.

It also looks like we could add this binding to Emacs mode.

Does this also cover beginning of line? That's not working like it does in `bash`.

Last - I'd like to see tests added.

srdubya:
I see no obvious problems.  Looking at my Ctrl-T tests (BasicEditingTest.VI.cs, search for _.CtrlT), they don't cover the beginning of line nor end of line cases.  I'd be happy to help add them, as my implementation seems to mimic Bash in VI mode already.

andrewcromwell:
>It also looks like we could add this binding to Emacs mode.

Cool. I added it to the default key bindings.
>Does this also cover beginning of line? That's not working like it does in `bash`.

Sorry, don't understand. How is it supposed to work at the beginning of a line?
>Last - I'd like to see tests added.

I added some tests to cover beginning and end of line cases.

I noticed that in vi command mode, my original change had a problem: it pushed the cursor forward to the very end of the line, which shouldn't happen, and doesn't happen in `bash`. So I inserted a test to make it behave differently between emacs mode and vi mode.

But then I noticed that in vi insert mode, you can still have CTRL+T enabled to swap characters (and I think that's default in `bash`). If in insert mode, you should still be able to swap at the end of a line. So now the test tries to determine if it's in vi command mode by checking to see if `_viCmdKeyMap` is the current `_dispatchTable`.

If there's a better way to do this, or if I'm being unclear, please let me know.

srdubya:
`ViEndOfLineFactor` tests for command mode in the same way.

lzybkr:
> Sorry, don't understand. How is it supposed to work at the beginning of a line?

Sorry, my mistake, I thought I saw a difference. In hindsight, I think I noticed no beep vs. beep, but even now it seems they behave the same, so disregard.

> in vi insert mode, you can still have CTRL+T enabled to swap characters (and I think that's default in bash).

Seems like we should do the same.

Your commits introduce mixed CR/LF with LF line endings - can you fix this?

PSReadLine/ReadLine.vi.cs
```diff
@@ -665,24 +665,31 @@ public static void InvertCase(ConsoleKeyInfo? key = null, object arg = null)
         /// </summary>
         public static void SwapCharacters(ConsoleKeyInfo? key = null, object arg = null)
         {
-            if (_singleton._current <= 0 || _singleton._current >= _singleton._buffer.Length)
+            // if in vi command mode, the cursor can't go as far
+            int cursorRightLimit = (_singleton._dispatchTable == _viCmdKeyMap) ?
```
lzybkr: When using the ternary operator, I prefer either a single line, or this style:
```
int cursorRightLimit = _singleton._dispatchTable == _viCmdKeyMap
    ? _singleton._buffer.Length - 1
    : _singleton._buffer.Length;

lzybkr (changes requested):
Please fix the use of LF instead of CR/LF, and it would be nice to see Ctrl+T added by default to vi insert mode as well.

andrewcromwell (commented):
I've fixed the line endings, cleaned up the ternary expression, and added SwapCharacters to default vi mode.

lzybkr (approved):

Issue #241, opened 2015-05-28
Oops, something went wrong.


kevpar:
Oops, something went wrong.  Please report this bug with the details below.
Report on GitHub: https://github.com/lzybkr/PSReadLine/issues/new

Last 71 Keys:
 g c m Spacebar \* e v e n t \* Enter
 UpArrow Spacebar Spacebar Backspace | Spacebar ? Spacebar s o u r c e Spacebar - n e Spacebar N e t E v e n t P a c k e t C a p t u r e Enter
 g e t - e v e n t s u b Tab Escape Ctrl+u Ctrl+a Ctrl+u Ctrl+k

Exception:
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
   at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   at Microsoft.PowerShell.PSConsoleReadLine.Kill(Int32 start, Int32 length, Boolean prepend)
   at Microsoft.PowerShell.PSConsoleReadLine.ProcessOneKey(ConsoleKeyInfo key, Dictionary`2 dispatchTable, Boolean ignoreIfNoAction, Object arg)
   at Microsoft.PowerShell.PSConsoleReadLine.InputLoop()
   at Microsoft.PowerShell.PSConsoleReadLine.ReadLine(Runspace runspace, EngineIntrinsics engineIntrinsics, Object reserved)

MayaLuen:
getting the same almost - Last 58 keys etc

lzybkr:
Fixed via #540

PR #540, opened 2017-06-05
test for negative _killIndex

andrewcromwell:
Fixes issue #241 (Reproducible with "A CTRL+K CTRL+U", typed in a fresh Powershell window)
Just tests `_killIndex` for negativity. `_killIndex` starts at -1 and erroneously indexes `_killRing`.

### DIFF
--- a/PSReadLine/KillYank.cs
+++ b/PSReadLine/KillYank.cs
@@ -63,40 +63,47 @@ namespace Microsoft.PowerShell
 
         private void Kill(int start, int length, bool prepend)
         {
-            if (length > 0)
+            if (length <= 0)
             {
-                var killText = _buffer.ToString(start, length);
-                SaveEditItem(EditItemDelete.Create(killText, start));
-                _buffer.Remove(start, length);
-                _current = start;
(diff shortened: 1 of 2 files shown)


PSReadLine/KillYank.cs
```diff
@@ -70,7 +70,7 @@ private void Kill(int start, int length, bool prepend)
                 _buffer.Remove(start, length);
                 _current = start;
                 Render();
-                if (_killCommandCount > 0)
+                if (_killCommandCount > 0 && _killIndex >= 0)
```
lzybkr: Do you think it would be cleaner to not increment `_killCommandCount` when `length == 0`? In other words, move [this line](https://github.com/lzybkr/PSReadLine/blob/e1dcaaededb4e672491dd4b782a2b213b5e97c1d/PSReadLine/KillYank.cs#L102) up, or maybe just add an early return when `length == 0`, that would make the code less indented.

lzybkr:
I'd like to see a test added as well.

lzybkr (approved):

PR #564, opened 2017-10-23
Implement VT input on Windows

parkovski:
Change `ReadOneOrMoreKeys` to pass the keys through an `ICharMap` filter. The default behavior stays the same by using a no-op filter, but VT mode is activated by using a redirected input stream, using a console with `ENABLE_VIRTUAL_TERMINAL_INPUT` set, or setting the environment variable `PSREADLINE_VTINPUT=1`.

I don't remember why I took out the redirected input error - probably wishful thinking. It still crashes under tmux, but just with a cursor positioning function. I can change it back if that's not an easy fix.

### DIFF
--- a/PSReadLine/CharMap.cs
+++ b/PSReadLine/CharMap.cs
@@ -8,36 +8,36 @@ using System.Diagnostics;
 
 namespace Microsoft.PowerShell
 {
-    /// Character sequence translator for platforms that behavior is not
-    /// natively supported (currently just Windows ANSI input).
+    // Character sequence translator for platforms that behavior is not
+    // natively supported (currently just Windows ANSI input).
     internal interface ICharMap
     {
-        /// How long to wait after seeing an escape should we wait before
-        /// giving up on looking for a sequence?
+        // How long to wait after seeing an escape should we wait before
+        // giving up on looking for a sequence?
         long EscapeTimeout { get; set; }
-        /// A key may become available even if nothing else was read because
-        /// of the escape sequence timer.
+        // A key may become available even if nothing else was read because
+        // of the escape sequence timer.
         bool KeyAvailable { get; }
-        /// If this is true, we don't want to block on `Console.ReadKey` or
-        /// the escape won't get seen until the next key is pressed.
+        // If this is true, we don't want to block on `Console.ReadKey` or
+        // the escape won't get seen until the next key is pressed.
         bool InEscapeSequence { get; }
-        /// Read a key from the processing buffer. An unspecified value
-        /// is returned if `KeyAvailable` is false.
+        // Read a key from the processing buffer. An unspecified value
+        // is returned if `KeyAvailable` is false.
         ConsoleKeyInfo ReadKey();
-        /// Insert a new key into the processing buffer. It is important that
-        /// immediately after every call to this, you check `KeyAvailable` as
-        /// the implementations are not designed to hold keys that don't form
-        /// a recognizable sequence.
+        // Insert a new key into the processing buffer. It is important that
+        // immediately after every call to this, you check `KeyAvailable` as
+        // the implementations are not designed to hold keys that don't form
+        // a recognizable sequence.
         void ProcessKey(ConsoleKeyInfo key);
     }
 
-    /// No-op - relies on whatever processing the .NET Console class does,
-    /// which on Unix reads from terminfo, and on Windows is none to very little.
+    // No-op - relies on whatever processing the .NET Console class does,
+    // which on Unix reads from terminfo, and on Windows is none to very little.
     internal class DotNetCharMap : ICharMap
     {
         private ConsoleKeyInfo _key;
 
-        /// Unused
+        // Unused
         public long EscapeTimeout {
             get { return 0; }
             set {}
@@ -60,19 +60,19 @@ namespace Microsoft.PowerShell
         }
     }
 
-    /// Hard-coded translator for the only VT mode Windows supports.
+    // Hard-coded translator for the only VT mode Windows supports.
     internal class WindowsAnsiCharMap : ICharMap
     {
         private List<ConsoleKeyInfo> _pendingKeys;
-        /// The next index in `_pendingKeys` to write to.
+        // The next index in `_pendingKeys` to write to.
         private int _addKeyIndex;
-        /// The next index in `_pendingKeys` to read from. This index becomes
-        /// valid when:
-        /// - The first character is escape and the escape timeout elapses.
-        /// - A sequence is completed.
-        /// - A single readable character is inserted.
+        // The next index in `_pendingKeys` to read from. This index becomes
+        // valid when:
+        // - The first character is escape and the escape timeout elapses.
+        // - A sequence is completed.
+        // - A single readable character is inserted.
         private int _readKeyIndexFrom;
-        /// The upper bound of `_pendingKeys` to read from, exclusive.
+        // The upper bound of `_pendingKeys` to read from, exclusive.
         private int _readKeyIndexTo;
 
         private Stopwatch _escTimeoutStopwatch = new Stopwatch();
@@ -264,8 +264,8 @@ namespace Microsoft.PowerShell
             }
         }
 
-        /// '\0' is used as an invalid character - it is valid, but only
-        /// by itself, not as part of a sequence (^[^@ is Esc, Ctrl-@, not Alt-Ctrl-@).
+        // '\0' is used as an invalid character - it is valid, but only
+        // by itself, not as part of a sequence (^[^@ is Esc, Ctrl-@, not Alt-Ctrl-@).
         private char GetSeqChar(int i)
         {
             // Only return valid key indexes for this scan.
@@ -284,12 +284,12 @@ namespace Microsoft.PowerShell
             return ch;
         }
 
-        /// Called when _pendingKeys[0] == ESC but it's not a full sequence.
-        /// As far as I can tell, the only keys that can't be combined with
-        /// alt are Esc, ^@, and Backspace (which generates ^[^H), but that
-        /// gets translated in `ProcessControlKey` so we have to let it go here.
-        /// None of the keys this applies to have their KeyChar set to 0 -
-        /// the ones that do have a special alt sequence handled later.
+        // Called when _pendingKeys[0] == ESC but it's not a full sequence.
+        // As far as I can tell, the only keys that can't be combined with
+        // alt are Esc, ^@, and Backspace (which generates ^[^H), but that
+        // gets translated in `ProcessControlKey` so we have to let it go here.
+        // None of the keys this applies to have their KeyChar set to 0 -
+        // the ones that do have a special alt sequence handled later.
         private bool ProcessAltSequence()
         {
             var ch = GetSeqChar(1);
@@ -316,9 +316,9 @@ namespace Microsoft.PowerShell
             return true;
         }
 
-        /// Scan for input escape sequences.
-        /// We're only interested in the range 0 to _addKeyIndex when
-        /// _pendingKeys[0] == Escape.
+        // Scan for input escape sequences.
+        // We're only interested in the range 0 to _addKeyIndex when
+        // _pendingKeys[0] == Escape.
         private void ProcessMultipleKeys()
         {
             if (_pendingKeys[_addKeyIndex - 1].KeyChar == '\x1b')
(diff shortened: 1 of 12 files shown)


lzybkr:
Awesome, I'll take a closer look when I have more time, probably next week but hopefully sooner.

Can you add some tests?

parkovski:
Sure.

PSReadLine/ReadLine.cs
```diff
@@ -87,9 +92,33 @@ private void ReadOneOrMoreKeys()
...
+                    {
+                        if (_console.KeyAvailable)
+                        {
+                            _charMap.ProcessKey(_console.ReadKey());
+                        }
+                        else
+                        {
+                            // This should be ok since it will only run if there
+                            // are no keys waiting to be read.
+                            Thread.Sleep(5);
```
lzybkr: Should the sleep happen in `_charMap` which knows the maximum timeout after escape?

PSReadLine/ReadLine.vi.cs
```diff
@@ -515,7 +515,15 @@ private void ViIndicateCommandMode()
...
             {
-                _console.CursorSize = _normalCursorSize < 50 ? 100 : 25;
+                if (PlatformWindows.IsConsoleApiAvailable())
+                {
+                    _console.CursorSize = _normalCursorSize < 50 ? 100 : 25;
+                }
+                else
+                {
+                    // Set the cursor to a solid blinking block.
+                    Console.Write("\x1b[2 q");
```
lzybkr: This probably belongs in the console wrapper.

To help unit testing, I don't use any `Console` apis directly, if I do, point them out so I can fix.

By wrapping all uses of the `Console` api, I also stand a chance at getting this stuff working on Win7/Win8 where there is no support for escape sequences unless you have something like ConEmu.

PSReadLine/ReadLine.vi.cs
```diff
@@ -530,7 +538,15 @@ private void ViIndicateInsertMode()
...
             {
-                _console.CursorSize = _normalCursorSize;
+                if (PlatformWindows.IsConsoleApiAvailable())
+                {
+                    _console.CursorSize = _normalCursorSize;
+                }
+                else
+                {
+                    // Set the cursor to a blinking vertical line.
+                    Console.Write("\x1b[5 q");
```
lzybkr: Same here - avoid use of `Console`.

PSReadLine/CharMap.cs
```diff
@@ -0,0 +1,705 @@
...
+Copyright (c) Microsoft Corporation.  All rights reserved.
+--********************************************************************/
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Microsoft.PowerShell
+{
+    /// Character sequence translator for platforms that behavior is not
```
lzybkr: Why `///`? These aren't XML comments (and they don't need to be.)

lzybkr (commented):
The changes look pretty good, but it sounds like it's not quite ready to merge, so I'll wait for you to say it's ready.

Is there any chance you can get this working on Linux as well, maybe test under WSL? I'm fine with the support you have, maybe just add some checks for common TERM settings that seem close to what you've implemented and throwing if you don't think the TERM can be supported without full `terminfo` support (which I don't think we should bother with.)

PSReadLine/CharMap.cs
parkovski: I do that a lot by accident, I'll change them.

PSReadLine/ReadLine.vi.cs
parkovski: I'll move it.

parkovski:
This PR on its own basically works - I've been using it for a while and the only issue I've noticed that I'd like to fix before it can be merged is after a crash (I still get them in tab menu complete sometimes), control keys won't be recognized until the `CharMap` buffer gets cleared. After that it does recover though and works fine again.

On the Linux topic, I thought .NET handled terminfo. After some testing though, it looks like that implementation needs some work. PSReadLine works with the default implementation in Gnome terminal, Xterm, and the built in Mac terminal. It almost works in the Windows console with WSL, except every letter has spaces around it. And in iTerm lots of partial escapes show up in the input. So I'll see if I can get this working on those platforms so there's at least something reliable until .NET fixes their implementation.

As far as actually handling redirected input goes, I'm really not sure how that needs to work. Most of it should be pretty easy - just change the `IConsole` implementation to use escape sequences. I don't know how to actually read anything though - `Console.ReadKey` doesn't work for redirected input, and everything else seems to wait for a whole line before returning. Do you have any thoughts on this? I don't see any way to read a single character at a time without calling native APIs.

PSReadLine/ReadLine.cs
parkovski: The idea behind this is that as soon as it sees something that isn't part of an escape sequence, there's no need to wait out the whole timer, so this is more of a CPU yield during a poll loop than a sleep. I'll add a comment explaining that.

lzybkr:
Great, just let me know when it's ready to merge.

As for extra spaces under WSL, try rebasing on master - I think i fixed that.

And when Console.ReadKey doesn't work because of redirected input, you could use Console.In directly. I should say that I don't know how often this is an issue on Linux - maybe it's less of a problem, but I thought I'd seen it at least once, so I don't know.

As for .Net - it's not something they plan to work on soon, see https://github.com/dotnet/corefx/issues/12336.

parkovski:
Cool, the rebase pretty much fixed everything on Linux and Mac. Unfortunately the escape processing is built in to `Console.ReadKey` on those platforms and it's got some issues. I could make my changes run on redirected input, but ssh and tmux are working already so I'm not really sure where that would apply. I don't think there's a way to disable the built in processing with regular tty input.

The issues I've noticed so far with the Unix `Console.ReadKey` implementation:
- It matches the `Esc + key` pattern as an alt key, but the timeout is either really long or doesn't exist, so vi editing mode is really frustrating because you have to wait a couple seconds after you hit escape.
- Some escapes are not recognized properly under tmux, which shouldn't happen if it's reading terminfo - I have a terminfo entry for tmux and zsh handles it fine.
- Backspace works fine everywhere except ssh - it just prints `^H` to the console. This one should be fixable in PSReadLine. The others I'll just have to open bugs with .NET and hope they get around to it soon.

Otherwise the only other problems are console APIs that need to be replaced with ANSI escapes.

PSReadLine/PlatformWindows.cs
```diff
@@ -129,16 +165,59 @@ internal static void SetOurInputMode()
...
+            // the console doesn't spew escape sequences all over.
+            mode &= ~ENABLE_VIRTUAL_TERMINAL_INPUT;
+        }
+
         SetConsoleInputMode(mode);
     }
 
+    private static void EnableAnsiInput(ref ICharMap charMap)
+    {
+        if (uint.TryParse(Environment.GetEnvironmentVariable("PSREADLINE_ESCTIMEOUT"), out var escTimeout))
```
lzybkr: This should probably be an option that you can set with `Set-PSReadLineOption`.

parkovski:
I'm going to rewrite this last commit - it's really messy, but it does work with a couple small positioning glitches on Windows without using any console APIs. I'm hitting a problem though - when you start it from WSL (`ubuntu run pwsh.exe`), the escapes get written to the console and it's not able to read the cursor position response. I originally thought it was a synchronization issue so a lot of the mess here was from trying to work through that, but given that it works in a normal Windows console, I'm pretty sure WSL is giving me the default console mode when I start an exe from it. Since console APIs are not available, I can't really change it. Let me know if you get a chance to look at it and have any ideas. Otherwise I do have a last resort option that feels hacky, but if I can go up the process chain and find a conhost, maybe I can attach to its console and change the mode that way.

There's something strange I just noticed that might have to do with this. If you start a Windows PowerShell in a Linux GUI terminal (gnome-terminal, etc), for the most part the basic editing mode works. If you type clear though, it clears the Windows console you started WSL from, and if you don't provide cmdlet parameters it'll ask for them in that console, but it still reads input from the gnome-terminal. Write-Host and others still write to the gnome-terminal too. Not sure what's going on there.

parkovski:
Well, I'm not sure. I've been experimenting with this - https://github.com/parkovski/isvt - and it looks like the mode is ok but that doesn't seem to make sense.

lzybkr:
I haven't looked closely at `isvt`, but if it's not working the way you expect, maybe try this [gist](https://gist.github.com/lzybkr/f2059cb2ee8d0c13c65ab933b75e998c) which has PowerShell commands to get/set the console mode.

parkovski:
It’s a program to attach to another process’ console and report/change the mode, helpful here because console APIs aren’t available to the actual process when started under WSL. Not convinced it’s entirely the mode, it may still have to do with synchronized access to the input stream. I’ll experiment with it some more.

lzybkr:
Oh, I see.

When PowerShell starts, it saves the current mode, sets it's own mode, and restores the initial mode when running external commands, see [here](https://github.com/PowerShell/PowerShell/blob/c75357b192c5eb8e06abcda6db594de33b2ad6ea/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs#L997) - so that might affect what you think you're doing.

parkovski:
Pretty sure I've got this figured out. It wasn't the mode - it looks like the stream from WSL to the Windows exe is not buffered, and if input comes in too quickly and no one's reading immediately it sends it to the screen. I unfortunately can't use something like `BufferedStream` because there has to _always_ be someone waiting for input, but I should be able to, if input is redirected, make a thread that constantly blocks on `Console.In.Read` and puts it in a buffer for `ReadKey`. I'm guessing I'll also have to make sure anything the shell launches gets that stream for `stdin` so it doesn't compete with the thread. If that all works though, that should fully support redirected input.

lzybkr:
Thanks for the update.

There is one change that I'm planning to make that will conflict with your changes. Is what you have good to merge, or if not, do you have an idea on when you think it will be ready to merge?

My change is to infer a paste operation based on how quickly we're seeing keys and this belongs in the same loop you need to update for this PR.

parkovski:
I'll take the last messy commit off and fix the comments, and then this will be good to merge. This handles all the ANSI escapes for keyboard input, and I'll just open another PR for the ConsoleLib stuff once that's working.

parkovski:
If this all looks good you can go ahead and merge it. I've had `PSREADLINE_VTINPUT` set for a while and it's been working for me so far.

I should mention that a lot of the tests don't pass with that environment variable set (they still all pass without it). All the ones I've tried manually give the results you'd expect - the problem is the ANSI reader doesn't expect things with shift, control, etc. already set, and even if you did translate everything to ANSI you still need to encode escape processing timeouts or fake them the way I did in the tests I wrote, and you might need to exclude some anyway since some key combos have no ANSI representation. Let me know how you want to handle that.

lzybkr:
Thanks so much for your contribution.

Don't worry about the tests - code coverage looks fine with your changes.

parkovski:
Absolutely! Glad to play some part in modernizing the console.

Turns out I was wrong about the input pipe - it doesn't just keep pushing data through like an audio stream so a BufferedStream should work fine. No idea why I thought that, but I should be able to get a pure ANSI `IConsole` working soon.

What's the plan for PowerShell's copy of PSReadLine? If you can pull from this repo before release, give me a date and I'll do my best to make sure redirected input is fully working before that.

lzybkr:
I have a feeling PSReadLine 2.0 is too big a change for the PowerShell team to take in 6.0 since they have a release candidate and 2.0 hasn't seen a ton of use yet, and I dropped support for non-ansi terminals like Win7/8 (which can be fixed more or less, just not done.)

I don't know if 2.0 will make it's way back into Windows PowerShell - better CJK support should make it worthwhile, but they might wait on screen reader support which I need some help with.

PSReadLine/PlatformWindows.cs
```diff
@@ -218,4 +291,17 @@ private static bool IsHandleRedirected(bool stdin)
...
         return !GetConsoleMode(handle, out var unused);
     }
+
+    public static bool IsConsoleApiAvailable()
+    {
+        if (!RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+        {
+            return false;
+        }
+        if (IsHandleRedirected(stdin: true) && IsHandleRedirected(stdin: false))
```
lzybkr, line 301: @parkovski - is this line correct? It seems like redirected stdout is enough to be a problem, see #591.

parkovski: It should probably have an input/output argument like `IsHandleRedirected`. I think there are four cases for console APIs:
- Input is not redirected; input APIs are usable.
- Output is not redirected; output APIs are usable.
- At least one of input or output is not redirected; things like `SetConsoleTitle` are usable.
- Both streams are redirected; we can't use any console APIs.

It's only used for the first scenario right now, but in case any of the others are needed, my new PR fixes it to work like this.

Issue #37, opened 2013-09-20
Copy-paste of large body of code can't be interrupted


FallenGameR:
Repro 
- copy large body of a Powershell script (~100 lines).
- paste into the console and watch it paint itself
- try to cancel with Ctrl+C - nothing happens, the text would still slowly appear on the screen

And pasting takes significant amount of time. You can see chars appearing one by one as if there is 100ms delay between each character render.

lzybkr:
Closing in favor of #579 which covers all of the known issues with right-click paste.

Issue #245, opened 2015-06-15
CRLF is printing ^M


juneb:
![untitled2](https://cloud.githubusercontent.com/assets/2954883/8173535/248cec54-138c-11e5-80d3-d08826ba7a41.jpg)
Windows standard line endings are CRLF. In our PowerShell host, we make sure each \n is converted to a \r\n to conform to the standard, but PSReadLine expects only \n. When we pass \r\n, it prints ^M. 

PSReadLine should conform to the Windows standard. Is it possible to make this change?

lzybkr:
How do I reproduce this issue?

juneb:
Jason, this is the PowerShell Studio console, but it should repro on any host that interprets \n as \r\n, per the Windows standard. 
To repro in PowerShell Studio: 
1. Download and install the 45-day trial: https://www.sapien.com/software/powershell_studio
2. If not 5.0, import PSReadline.
3. In PowerShell Studio, Ctrl+N (new file).
4. In the new file, type (any command), e.g. Get-Process
5. Select the command and click Run / Run Selection in Console (Ctrl+Shift+F8).

lzybkr:
I think PSReadline is doing the correct thing here, and the PowerShell Studio host is incorrectly sending \r as a keystroke.

For comparison, conhost does not send that key when pasting via right click.  Similarly, if you paste via Ctrl+v, PSReadline does the right thing here.

That said, it looks like PowerShell Studio is redirecting standard input.  That causes lots of problems because there is no terminal emulation support, so many key binding don't work.  When PSReadline sees redirected input, I think I'll be forced to deal with emulating what conhost does in terms of ignoring \r.

If I'm wrong about how PowerShell Studio is implemented, then I think PowerShell Studio needs to be changed because it would be acting more like conhost in that case.

juneb:
Thanks for the quick reply, Jason. I'll pass on the info.

juneb:
I just checked and we don't redirect anything. We use standard Microsoft APIs to attach a console to a Windows host (AttachConsole) and write to the console (WriteConsole). WriteConsole requires CRLF, so we use it. It seems to me that PSReadline should support this.

lzybkr:
Closing in favor of #579 which covers all of the known issues with right-click paste.

Issue #77, opened 2014-01-13
File name inserted into paste input


frizop:
Pasting in the following three lines causes the third line to insert the first tab completed file name (for me, it's .berkshelf):

function Get-Foo {
    $defaultDisplayPropertySet = New-Object System.Management.Automation.PSPropertySet('DefaultDisplayPropertySet',[string[]]$defaultProperties)
    $PSStandardMembers = [System.Management.Automation.PSMemberInfo[]]@($defaultDisplayPropertySet)

Please note, that there are one tab characters preceding the second and third lines.

kayasax:
I have same problem here, but ctrl+V only paste the first line of my clipboard ...? (I'm using powershell within ConEmu)

lzybkr:
Closing in favor of #579 which covers all of the known issues with right-click paste.

PR #565, opened 2017-10-25
Workaround for reversed right click paste

### LINKED ISSUE #496: Right-click paste with \n comes out upside down?
When using PSReadLine 1.2, and pasting text that has newlines _without_ carriage returns, the lines come out in reverse order.

Tadas:
Fixes #496 

Not sure if this is the correct way to detect right click, but I found the same method in https://github.com/lzybkr/PSReadLine/blob/9d592b875ebea83d6385b9cc6b6899584b08277a/PSReadLine/BasicEditing.cs#L246-L249

### DIFF
--- a/PSReadLine/BasicEditing.cs
+++ b/PSReadLine/BasicEditing.cs
@@ -466,6 +466,20 @@ namespace Microsoft.PowerShell
         /// </summary>
         public static void InsertLineAbove(ConsoleKeyInfo? key = null, object arg = null)
         {
+            // If there are more keys, assume the user pasted with a right click and
+            // we shouldn't reverse the order of the lines
+            if (_singleton._queuedKeys.Count > 0)
+            {
+                _singleton.AcceptLineImpl(false);
+            }
+            else
+            {
+                _singleton.InsertLineAboveImpl();
+            }
+        }
+
+        private void InsertLineAboveImpl()
+        {
             // Move the current position to the beginning of the current line and only the current line.
             if (_singleton.LineIsMultiLine())
             {


lzybkr:
Thanks for the PR.

There is no reliable way I know of to detect paste - we must use a heuristic.

I'm not sure this is the best approach. I think the right fix to this issue will also fix #37 and #77, maybe #245 and possibly others.

What I'd planned was to implement a stronger heuristic that measures the time between keys and makes assumptions about how fast a human can type - and if that rate is greatly exceeded, collect all of the keys and do a single paste operation, hopefully the same as though the user used Ctrl+V.

I guess there are a few patterns where a key is obviously not something pasted - e.g. when Alt is specified, and probably some with Ctrl, though some keys with Ctrl set could be pasted, so it's not a clear cut.

Tadas:
Ah the pleasures of conhost :) Looking at how fast the keys are coming in might actually be a workable solution. Alternatively, have you thought about getting a hold of the mouse events? [It seems to be possible](https://docs.microsoft.com/en-us/windows/console/reading-input-buffer-events)

Let's close this in favour of a more robust solution.

lzybkr:
I hadn't considered reading mouse events for this issue - but reading those does require not using `Console.ReadKey`, which means duplicating all of that code. I know this because I investigated detecting resize events.

That said, it's not clear to me that the right click event will be available - and even if it is, that doesn't help if the user types Alt+Space,e,p to paste.

lzybkr:
I opened #579 to track all of the right click paste issues - and I think they are all addressable without this PR so I'm closing.

PR #570, opened 2017-11-02
Mention that PSReadLine has a different default than bash

mikemaccana:


README.md
```diff
@@ -93,7 +93,7 @@ To start using, just import the module:
 Import-Module PSReadLine
 ```
 
-To use Emacs key bindings, you can use:
+PSReadLine, unlike bash's readline, uses vi key bindings by default. To use Emacs key bindings, you can use:
```
lzybkr, line 96: That's not correct. On Windows, the default is Windows. On Mac/Linux, the default is Emacs.

mikemaccana:
Odd. I ran `get-PSReadlineOption` and the results were: 

    EditMode                               : Vi

I'm not modifying the default in my profile scripts. I wonder where it's coming from? 🤔

lzybkr:
BTW, the code is [here](https://github.com/lzybkr/PSReadLine/blob/9d592b875ebea83d6385b9cc6b6899584b08277a/PSReadLine/Cmdlets.cs#L86).

I can't explain what you're seeing - maybe in another profile (try something like `($profile | gm -MemberType NoteProperty).Name | % { $Profile.$_ } | Select-String Vi` - this would search all of the possible profiles.)

mikemaccana:
Weird, I've already looked but also tried `($profile | gm -MemberType NoteProperty).Name | % { $Profile.$_ } | Select-String Vi` and nothing comes back. Thanks for the correction though.

lzybkr:
Closing this as it's definitely not supposed to be that way - I'd never hear the end of it if `vi` mode was the default - nobody could ever exit, right :)

gwojan:
Lol! @lzybkr, everyone knows you're supposed to reboot your machine to exit vi...

Issue #571, opened 2017-11-04
Exception after Escape in vi mode


lzybkr:
In trying to reproduce #539, I hit the following exception.

Environment data
----------------
```
PS version: 5.1.16299.19
PSReadline version: 2.0
os: 10.0.16299.15 (WinBuild.160101.0800)
PS file version: 10.0.16299.15 (WinBuild.160101.0800)
```

Steps to reproduce or exception report
--------------------------------------
```
#28 PS> powershell -nop
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

PS C:\PSReadLine\PSReadLine> Set-PSReadlineOption -EditMode vi
PS C:\PSReadLine\PSReadLine> Set-PSReadlineKeyHandler -Chord 'j,k' -Function NextLine -ViMode Command
PS C:\PSReadLine\PSReadLine> Set-PSReadlineKeyHandler -Chord 'j,k' -Function NextLine -ViMode Command
Oops, something went wrong.  Please report this bug with the details below.
Report on GitHub: https://github.com/lzybkr/PSReadLine/issues/new
-----------------------------------------------------------------------
Last 105 Keys:
 s e t - p s r e Tab Tab o Backspace Space - e d Tab Space v i Enter
 s e t - p s r e Tab Space - c h Tab Space ' , j k Backspace Backspace Backspace j , k ' Space - f u Tab Space O Tab Tab Backspace Backspace Space N e x t L Tab Space - v i Tab Space C o Tab Enter
 a Space b Escape k Alt+r Escape Escape Alt+r 0 D Escape i a Space b Escape j k Alt+r Escape 0 C UpArrow Alt+f Alt+f Alt+f Alt+f Alt+f Escape

Exception:
System.ArgumentOutOfRangeException: Non-negative number required.
Parameter name: count
   at System.ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument argument, ExceptionResource resource)
   at System.Collections.Generic.List`1.GetRange(Int32 index, Int32 count)
   at Microsoft.PowerShell.PSConsoleReadLine.EndEditGroup(Action`2 instigator, Object instigatorArg)
   at Microsoft.PowerShell.PSConsoleReadLine.GroupUndoHelper.EndGroup()
   at Microsoft.PowerShell.PSConsoleReadLine.ViCommandMode(Nullable`1 key, Object arg)
[...]

PR #572, opened 2017-11-05
Fix for #571

srdubya:
When in Insert mode, UpArrow now results in:

1. Switch to Command mode
2. Move up in the history
3. Switch back to Insert mode

This addressed an ArgumentOutOfRangeException when using the UpArrow to move up in the history while in Insert mode immediately after making a command line edit.

Note, I choose to create a new file with this fix, History.vi.cs, but I'd be happy to add the change to History.cs if you prefer.

### DIFF
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -772,10 +772,17 @@ namespace Microsoft.PowerShell
             case BellStyle.None:
                 break;
             case BellStyle.Audible:
-                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
-                    Console.Beep(Options.DingTone, Options.DingDuration);
-                else
-                    Console.Beep();
+                if (Options.DingDuration > 0)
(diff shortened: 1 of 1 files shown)


lzybkr:
You have merge commits in your master branch, so PRs from you will never merge nicely.

To fix, you need to force push to your master, I think this should fix it:

```
# In a new directory somewhere
git clone https://github.com/lzybkr/PSReadLine
git remote rename origin upstream
git remote add origin https://github.com/srdubya/PSReadLine
git push -f
```

If you're lucky, this won't clobber your most recent fix, which you could then apply with:

```
git fetch
git cherry-pick  6e0a663
```

I'm sure there are other ways to fix this, but this is the simplest for me to understand - basically starting your master branch from scratch.

srdubya:
Sorry, I thought I'd already fixed that.  I've made another attempt.  Hopefully this one is cleaner.

PR #600, opened 2017-12-03
Switch from MSTest to xUnit, enable tests in AppVeyor

lzybkr:


lzybkr:
@parkovski - I'd like your thoughts on the redirected input changes I made.

I started messing around with redirected input locally and made these [changes](https://github.com/lzybkr/PSReadLine/pull/600/commits/29d0d944acab611d18fae3188094607d608e3058) so that most of the tests pass when running using Resharper's test runner - I think 24 failed but I didn't investigate after setting the escape timeout to 0.

Then I tried these changes in WSL under tmux and I still get an exception from `Console.BufferWidth`.

That said, https://github.com/lzybkr/PSReadLine/pull/600/commits/29d0d944acab611d18fae3188094607d608e3058 is sufficient to get the tests to pass in AppVeyor (because Console.BufferWidth isn't getting called, my mock of it is instead.)

parkovski:
Yeah, most of the console APIs in ConsoleLib will still throw, which is why I left the exception in there for now. I had a mostly working `IConsole` that replaced them with ANSI sequences - unfortunately what I thought were my bugs [may not have been](https://github.com/Microsoft/WSL/issues/2704) and I apparently didn't `git stash` it when I thought I did 😞. So I need to finish rewriting that. (Post-edit - just realized the problem I was having is the off by one round-trip thing he mentions in that issue).

Your changes look fine. It looks like the tests are failing because of combined escape sequences (`^[[X;Ym`) - replacing the escape section in `TestConsole.Write` with this fixes some of them:
```csharp
if (s[i] == (char) 0x1b)
{
    // Escape sequence - limited support here, and assumed to be well formed.
    var endSequence = s.IndexOf("m", i, StringComparison.Ordinal);
    // Chop off the "^[[" and "m" and add them back later to allow combined sequences.
    var escapeSequences = s.Substring(i + 2, endSequence - i - 2).Split(';');
    foreach (var escapeSequence in escapeSequences)
    {
        EscapeSequenceActions[$"\x1b[{escapeSequence}m"](this);
    }
    i = endSequence;
    continue;
}
```

The rest are failing because of a background color mismatch.

More fail if you set `PSREADLINE_VTINPUT=1`, and to make that work you'd have to have `Keys` map all the `ConsoleKeyInfo`s to escape sequences.

lzybkr:
Thanks - I think I'll put the exception back then, but disable the exception if an environment variable is set so tests can run this way.

lzybkr:
Oh, I also remember looking at one random failure where `Shift+Enter` wasn't working - and that made sense because with VT input, you never see the shift state, so I think some key bindings just can't work with VT input.

Issue #606, opened 2017-12-08
Invoke-Build doesn't  work when nuget.exe is absent


MVKozlov:
``` powershell
Invoke-WebRequest http://nuget.org/nuget.exe -OutFile $nugetExe
```
this code get nuget 2.8 but not latest.

Additionally
First call failed with "I want nuget 2.12+"
But next fail at complation stage because it want `System.Runtime.InteropServices.RuntimeInformation.4.3.0`

MVKozlov:
solution for part 1:
``` diff
diff --git a/PSReadLine.build.ps1 b/PSReadLine.build.ps1
index 8f5db1f..685b790 100644
--- a/PSReadLine.build.ps1
+++ b/PSReadLine.build.ps1
@@ -35,7 +35,7 @@ task CheckNugetInstalled `
         $script:nugetExe = "${env:TEMP}/nuget.exe"
         if (!(Test-Path $nugetExe))
         {
-            Invoke-WebRequest http://nuget.org/nuget.exe -OutFile $nugetExe
+            Invoke-WebRequest https://dist.nuget.org/win-x86-commandline/latest/nuget.exe -OutFile $nugetExe
         }
     }
 }
```

lzybkr:
Feel free to submit a PR of you have it fixed.

Issue #605, opened 2017-12-08
PSReadLine 2.0 not fully implement all needed ANSI Sequences for Windows Platform before Windows 10


MVKozlov:
PSReadLine 2.0 not compatible with Windows7 ?

I have PSv5.1 but can't use MenuComplete :(

MVKozlov:
quick compatibility hack
``` diff
diff --git a/PSReadLine/PlatformWindows.cs b/PSReadLine/PlatformWindows.cs
index 2cb3c6d..104bb20 100644
--- a/PSReadLine/PlatformWindows.cs
+++ b/PSReadLine/PlatformWindows.cs
@@ -373,6 +373,9 @@ static class PlatformWindows
             {"\x1b[30;47m", () => {
                 Console.ForegroundColor = ConsoleColor.Black;
                 Console.BackgroundColor = ConsoleColor.Gray; } },
+            {"\x1b[35;43m", () => {
+                Console.ForegroundColor = ConsoleColor.DarkMagenta;
+                Console.BackgroundColor = ConsoleColor.DarkYellow; } },
             {"\x1b[40m", () => Console.BackgroundColor = ConsoleColor.Black},
             {"\x1b[44m", () => Console.BackgroundColor = ConsoleColor.DarkBlue },
             {"\x1b[42m", () => Console.BackgroundColor = ConsoleColor.DarkGreen},
```

MVKozlov:
I think this is much better
``` diff
diff --git a/PSReadLine/PlatformWindows.cs b/PSReadLine/PlatformWindows.cs
index 2cb3c6d..45c5b5d 100644
--- a/PSReadLine/PlatformWindows.cs
+++ b/PSReadLine/PlatformWindows.cs
@@ -370,9 +370,6 @@ static class PlatformWindows
         private static ConsoleColor InitialBG = Console.BackgroundColor;

         private static readonly Dictionary<string, Action> EscapeSequenceActions = new Dictionary<string, Action> {
-            {"\x1b[30;47m", () => {
-                Console.ForegroundColor = ConsoleColor.Black;
-                Console.BackgroundColor = ConsoleColor.Gray; } },
             {"\x1b[40m", () => Console.BackgroundColor = ConsoleColor.Black},
             {"\x1b[44m", () => Console.BackgroundColor = ConsoleColor.DarkBlue },
             {"\x1b[42m", () => Console.BackgroundColor = ConsoleColor.DarkGreen},
[...]

PR #608, opened 2017-12-09
Fixes for #605, #606

MVKozlov:
Added processing of possible errors. Maybe it's too much?

### DIFF
--- a/PSReadLine.build.ps1
+++ b/PSReadLine.build.ps1
@@ -35,7 +35,7 @@ task CheckNugetInstalled `
         $script:nugetExe = "${env:TEMP}/nuget.exe"
         if (!(Test-Path $nugetExe))
         {
-            Invoke-WebRequest http://nuget.org/nuget.exe -OutFile $nugetExe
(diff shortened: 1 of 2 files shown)


lzybkr:
@MVKozlov - thanks for the PR, it made it obvious what the problem was. I didn't see any real problems with your change, but I do want to minimize allocations.

I went ahead and made my own fix if you'd like to give it a try - see #611.

MVKozlov:
I think it is really better deal with allocations but I do not understand reasons why you don't `Console.Write(s.Substring(from, i - from)); ` before `while (!done && j < s.Length) ` as I do :)

In any way we have a couple of `action()` without any `Write` inside them, so write can be moved out of the cycle

lzybkr:
I've already had complaints about performance of the old implementation that was similar, so even though the code is simpler allocating many substrings, I'll stick with the better performing code.

Spans would be helpful here, but I don't want to ship another dll just for this.

I was optimizing for a scenario that really shouldn't happen, so I switched to your way, it's slightly cleaner I guess, but I went and made it less clean to correctly handle partially invalid sequences.

lzybkr:
I took your change for #606 and my own (with your suggestions) for #605, so I'm closing this PR.

PR #626, opened 2018-01-25
Changes for Editor Hosts (VSCode/Atom)

SeeminglyScience:
I've been working on getting PSReadLine working in VSCode/Atom and I came across two things that
I haven't been able to workaround without changes.

## Making ReadLine cancellable

For the most part I was able to get away with taking over the pipeline via the `OnIdle` events
that PSReadLine generates. For some things however, I needed a way to reliably cancel the ReadLine
pipeline. For example, stepping in the debugger or running a script that could trigger a debugger
stop.  To accomplish this I added an overload to `ReadLine` that accepts a cancellation token and
added the wait handle to the `ReadKey` key request loop.

## Workaround for Unix stdin locking

The Unix implementation of `System.Console` relies mainly on escape sequences. A side effect of
that implementation is you cannot use any of the cursor position API's (`Console.CursorLeft`, etc)
while `Console.ReadKey` is running or it'll block the thread until a key is pressed.  The main reason
this is an issue is because I run most of the background commands like command completion by taking
over the pipeline when PSReadLine checks for event subscribers.  Right before it checks, it saves
the cursor position so it can reposition the prompt if any of the events wrote to the pipeline.

You can actually see this outside of VSCode as well:

```powershell
$timer = [System.Timers.Timer]::new(100)
$timer.Start()
Register-ObjectEvent -InputObject $timer -EventName Elapsed -Action { Write-Host Testing! }
```

The event will only fire *after* a key is pressed. If you attach to the process and pause the debugger
it'll be stopped at [this line](https://github.com/lzybkr/PSReadLine/blob/2c8bfd39ce2b47a55647e25f655b81684596741f/PSReadLine/ReadLine.cs#L223).

We ran into this issue in the current implementation of our `ReadLine` in the Editor Services host as well.
Every time the debugger stopped on Unix a key needed to be pressed before the pipeline could
continue because the host would check the cursor position to re-adjust the prompt.

The most obvious fix to this issue is to wait until `Console.KeyAvailable` is true before calling 
`Console.ReadKey`. Unfortunately in Unix platforms input is echoed to the screen by default. The only
managed way we could find to disable that was to run `Console.ReadKey(true)`. So I stripped out the
native code from corefx that they use to disable input echo and created a separate package specifically
to solve this issue.

cc @daviwil

### DIFF
--- a/PSReadLine.build.ps1
+++ b/PSReadLine.build.ps1
@@ -202,6 +202,9 @@ task LayoutModule BuildMainModule, BuildMamlHelp, {
 
     Copy-Item PSReadLine/bin/$Configuration/Microsoft.PowerShell.PSReadLine2.dll $targetDir
     Copy-Item PSReadLine/bin/$Configuration/System.Runtime.InteropServices.RuntimeInformation.dll $targetDir
+    Copy-Item PSReadLine/bin/$Configuration/UnixConsoleEcho.dll $targetDir
+    Copy-Item PSReadLine/bin/$Configuration/libdisablekeyecho.so $targetDir
+    Copy-Item PSReadLine/bin/$Configuration/libdisablekeyecho.dylib $targetDir
 
     # Copy module manifest, but fix the version to match what we've specified in the binary module.
     $version = (Get-ChildItem -Path $targetDir/Microsoft.PowerShell.PSReadLine2.dll).VersionInfo.FileVersion
--- a/PSReadLine.build.ps1
+++ b/PSReadLine.build.ps1
@@ -202,9 +202,6 @@ task LayoutModule BuildMainModule, BuildMamlHelp, {
 
     Copy-Item PSReadLine/bin/$Configuration/Microsoft.PowerShell.PSReadLine2.dll $targetDir
     Copy-Item PSReadLine/bin/$Configuration/System.Runtime.InteropServices.RuntimeInformation.dll $targetDir
-    Copy-Item PSReadLine/bin/$Configuration/UnixConsoleEcho.dll $targetDir
-    Copy-Item PSReadLine/bin/$Configuration/libdisablekeyecho.so $targetDir
-    Copy-Item PSReadLine/bin/$Configuration/libdisablekeyecho.dylib $targetDir
 
     # Copy module manifest, but fix the version to match what we've specified in the binary module.
     $version = (Get-ChildItem -Path $targetDir/Microsoft.PowerShell.PSReadLine2.dll).VersionInfo.FileVersion
--- a/PSReadLine/ConsoleLib.cs
+++ b/PSReadLine/ConsoleLib.cs
@@ -97,7 +97,32 @@ namespace Microsoft.PowerShell.Internal
             set { try { Console.OutputEncoding = value; } catch { } }
         }
 
-        public ConsoleKeyInfo ReadKey()                  => Console.ReadKey(true);
+        private void WaitForKeyAvailable()
+        {
+            // On Unix platforms input echo is on by default. If we wait for KeyAvailable without
+            // disabling it, all input will be echoed.
+            UnixConsoleEcho.InputEcho.Disable();
+            try
+            {
+                while (!Console.KeyAvailable)
+                {
+                    System.Threading.Thread.Sleep(50);
+                }
+            }
+            finally
+            {
+                UnixConsoleEcho.InputEcho.Enable();
+            }
+        }
+
+        public ConsoleKeyInfo ReadKey()
+        {
+            // Wait for a key to be pressed before calling ReadKey to avoid locking stdin on
+            // non-Windows platforms.
+            WaitForKeyAvailable();
+            return Console.ReadKey(true);
+        }
+
         public bool KeyAvailable                         => Console.KeyAvailable;
         public void SetWindowPosition(int left, int top) => Console.SetWindowPosition(left, top);
         public void SetCursorPosition(int left, int top) => Console.SetCursorPosition(left, top);
--- a/PSReadLine/ConsoleLib.cs
+++ b/PSReadLine/ConsoleLib.cs
@@ -9,6 +9,11 @@ namespace Microsoft.PowerShell.Internal
 {
     internal class VirtualTerminal : IConsole
     {
+        private static Func<bool, ConsoleKeyInfo> _readKeyOverride;
+
+        private static Lazy<Func<bool, ConsoleKeyInfo>> _readKeyMethod = new Lazy<Func<bool, ConsoleKeyInfo>>(
+            () => _readKeyOverride == null ? Console.ReadKey : _readKeyOverride);
+
         public int CursorLeft
         {
             get => Console.CursorLeft;
@@ -97,7 +102,7 @@ namespace Microsoft.PowerShell.Internal
             set { try { Console.OutputEncoding = value; } catch { } }
         }
 
-        public ConsoleKeyInfo ReadKey()                  => Console.ReadKey(true);
+        public ConsoleKeyInfo ReadKey()                  => _readKeyMethod.Value(true);
         public bool KeyAvailable                         => Console.KeyAvailable;
         public void SetWindowPosition(int left, int top) => Console.SetWindowPosition(left, top);
         public void SetCursorPosition(int left, int top) => Console.SetCursorPosition(left, top);
--- a/PSReadLine/ConsoleLib.cs
+++ b/PSReadLine/ConsoleLib.cs
@@ -97,32 +97,7 @@ namespace Microsoft.PowerShell.Internal
             set { try { Console.OutputEncoding = value; } catch { } }
         }
 
-        private void WaitForKeyAvailable()
-        {
-            // On Unix platforms input echo is on by default. If we wait for KeyAvailable without
-            // disabling it, all input will be echoed.
-            UnixConsoleEcho.InputEcho.Disable();
-            try
-            {
-                while (!Console.KeyAvailable)
-                {
-                    System.Threading.Thread.Sleep(50);
-                }
-            }
-            finally
-            {
-                UnixConsoleEcho.InputEcho.Enable();
-            }
-        }
-
-        public ConsoleKeyInfo ReadKey()
-        {
-            // Wait for a key to be pressed before calling ReadKey to avoid locking stdin on
-            // non-Windows platforms.
-            WaitForKeyAvailable();
-            return Console.ReadKey(true);
-        }
-
+        public ConsoleKeyInfo ReadKey()                  => Console.ReadKey(true);
         public bool KeyAvailable                         => Console.KeyAvailable;
         public void SetWindowPosition(int left, int top) => Console.SetWindowPosition(left, top);
         public void SetCursorPosition(int left, int top) => Console.SetCursorPosition(left, top);
--- a/PSReadLine/ConsoleLib.cs
+++ b/PSReadLine/ConsoleLib.cs
@@ -9,6 +9,9 @@ namespace Microsoft.PowerShell.Internal
 {
     internal class VirtualTerminal : IConsole
     {
+        // These two fields are used by PowerShellEditorServices to inject a
+        // custom ReadKey implementation. This is not a public API, but it is
+        // part of a private contract with that project.
         private static Func<bool, ConsoleKeyInfo> _readKeyOverride;
 
         private static Lazy<Func<bool, ConsoleKeyInfo>> _readKeyMethod = new Lazy<Func<bool, ConsoleKeyInfo>>(
--- a/PSReadLine/ReadLine.cs
+++ b/PSReadLine/ReadLine.cs
@@ -30,6 +30,8 @@ namespace Microsoft.PowerShell
     {
         private static readonly PSConsoleReadLine _singleton = new PSConsoleReadLine();
 
+        private static readonly CancellationToken _defaultCancellationToken = new CancellationTokenSource().Token;
+
         private bool _delayedOneTimeInitCompleted;
 
         private IPSConsoleReadLineMockableMethods _mockableMethods;
@@ -41,6 +43,7 @@ namespace Microsoft.PowerShell
         private Thread _readKeyThread;
         private AutoResetEvent _readKeyWaitHandle;
         private AutoResetEvent _keyReadWaitHandle;
+        private CancellationToken _cancelReadCancellationToken;
         internal ManualResetEvent _closingWaitHandle;
         private WaitHandle[] _threadProcWaitHandles;
         private WaitHandle[] _requestKeyWaitHandles;
@@ -139,7 +142,12 @@ namespace Microsoft.PowerShell
                 if (handleId == 1) // It was the _closingWaitHandle that was signaled.
                     break;
 
+                var localCancellationToken = _singleton._cancelReadCancellationToken;
                 ReadOneOrMoreKeys();
+                if (localCancellationToken.IsCancellationRequested)
+                {
+                    continue;
+                }
 
                 // One or more keys were read - let ReadKey know we're done.
                 _keyReadWaitHandle.Set();
@@ -249,6 +257,13 @@ namespace Microsoft.PowerShell
                 throw new OperationCanceledException();
             }
 
+            if (handleId == 2)
+            {
+                // ReadLine was cancelled by the host, throw an exception so we can return
+                // an empty string.
+                throw new OperationCanceledException();
+            }
+
             var key = _singleton._queuedKeys.Dequeue();
             return key;
         }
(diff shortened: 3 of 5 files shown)


daviwil:
I've been discussing this approach with Patrick for a few months now and I'm fully in support of it.  This will enable PSReadline 2.0 to be a first class citizen in Editor Services!

PSReadLine.build.ps1
```diff
@@ -202,6 +202,9 @@ task LayoutModule BuildMainModule, BuildMamlHelp, {
 
     Copy-Item PSReadLine/bin/$Configuration/Microsoft.PowerShell.PSReadLine2.dll $targetDir
     Copy-Item PSReadLine/bin/$Configuration/System.Runtime.InteropServices.RuntimeInformation.dll $targetDir
+    Copy-Item PSReadLine/bin/$Configuration/UnixConsoleEcho.dll $targetDir
+    Copy-Item PSReadLine/bin/$Configuration/libdisablekeyecho.so $targetDir
```
lzybkr: Presumably this is an x64 binary? What about x86 or arm32 or arm64?

I see that the code comes from the CLR. Is there any reasonable way to ask the CLR to expose an API instead?

PSReadLine/ConsoleLib.cs
```diff
@@ -97,7 +97,32 @@ public Encoding OutputEncoding
...
+        private void WaitForKeyAvailable()
+        {
+            // On Unix platforms input echo is on by default. If we wait for KeyAvailable without
+            // disabling it, all input will be echoed.
+            UnixConsoleEcho.InputEcho.Disable();
+            try
+            {
+                while (!Console.KeyAvailable)
+                {
+                    System.Threading.Thread.Sleep(50);
```
lzybkr: I don't know about Linux, but on Windows, a sleep for less than 200ms or so (I forget exact details) prevents the CPU from going into low power mode because the kernel assumes the process is busy.

So I'm a bit concerned this might cause increased drain of the battery.

Is there some other way where we can block in the kernel instead of a busy loop?

PSReadLine.build.ps1
SeeminglyScience: For x86, it looks [it isn't supported](https://github.com/dotnet/core/blob/master/release-notes/2.0/2.0-supported-os.md) on Linux and Mac.  I do need to look into what I need to do for arm* though.

> Is there any reasonable way to ask the CLR to expose an API instead?

That definitely would have been preferred.  [Here's the thread](https://github.com/dotnet/corefx/issues/16352) we asked about calling their library for disabling input echo.  That's where they suggested stripping out their implementation.  And [here's](https://github.com/dotnet/corefx/issues/25036) the request to create an async version of `ReadKey`.

PSReadLine/ConsoleLib.cs
SeeminglyScience: I'm not having luck finding a way to reliably block until a key is available.  But a possible solution would be to set the sleep to 200ms until the first key is pressed, then switch to `SpinWait.SpinUntil` with a timeout of a couple seconds.  Then switch back to the 200ms sleep if the timeout is hit.  Do you have any concerns with that approach?

PSReadLine.build.ps1
lzybkr: Can you just run `stty -echo` instead? If you do that once before the call to PSReadLine and once to restore after, it seems not too bad. See https://stackoverflow.com/questions/5633472/how-do-i-turn-off-echo-in-a-terminal.

PSReadLine/ConsoleLib.cs
lzybkr: That also sounds less than ideal.

If you want to unblock - maybe we could agree on an escape sequence that you generate and PSReadLine ignores, other than to break out of the Console.ReadKey call of course.

PSReadLine.build.ps1
SeeminglyScience: I'm definitely open to the idea and looking into it.  Just to clarify though, when you say the call to PSReadLine do you mean call it from inside the `ReadLine` method or only call it from editor services?  Disabling input echo is needed to implement `WaitForKeyAvailable`.  If it was only disabled by the editor services host, every key press would appear to be rendered twice when in the standard host.

**Edit** If the escape sequence to break out of `ReadKey` works then this can be ignored and the package will be removed.

PSReadLine/ConsoleLib.cs
SeeminglyScience: Hmm, that's an interesting thought.  Early on when I was first looking into the issue for editor services I was experimenting with emitting the terminal information query escape sequence to break out of `ReadKey`.  That worked but wasn't reliable because of different possible return sequences.  If it's always the same and always ignored, that may work perfectly.  I'll do some testing.

PSReadLine.build.ps1
lzybkr: Yeah, I was referring to the method `ReadLine`. On Windows, we save the console mode at the beginning and restore it at the end, that's similar to what I was proposing for `stty`.

PSReadLine/ConsoleLib.cs
lzybkr: I think we'd need an invalid escape sequence so there's 0 chance the tty will interpret it. Alternatively, maybe we should use an [invalid Unicode character](http://www.unicode.org/faq/private_use.html).

SeeminglyScience: I'm remembering now that I did try to go this route initially, but I couldn't find a way to actually write to stdin directly.  At least not in a way that would release `Console.ReadKey`.  The closest I've come is query escape sequences written to stdout that emit back to stdin.  None of those that I'm aware of allow you to specify what they return.

Still digging, I would definitely prefer this method as well.  If you have any suggestions or hints, they would be greatly appreciated.

lzybkr: I only have hints, no direct experience.

On Windows, one can call `CreateFile` with the filename `CONIN$` and you will have a file handle to the keyboard. I've never tried writing to it, so it's possible it might not work.

I know less about Linux.  Maybe one of the files in `/dev` will work in a similar manner, e.g. `/dev/stdin` or `/dev/tty` - but I'm totally guessing here. You might look at how the tool `expect` is implemented, though it's possible that tool just redirects `stdin` when it creates the process.

Another option might be to just start powershell/pwsh with redirected input. On Linux, .Net converts the VT escape sequences generated by modified keys (e.g. `Alt-a`) to what PSReadLine expects. On Windows 10, thanks to @parkovski - PSReadLine can also do that mapping, but you need to set the console mode flag [ENABLE_VIRTUAL_TERMINAL_INPUT](https://docs.microsoft.com/en-us/windows/console/setconsolemode).

With a redirected stdin - the language service would own sending every key to PSReadLine, so it's possible you wouldn't need any of these changes. It does mean not supporting <Win10, or alternatively generating VT escape sequences for <Win10, and that might require some PSReadLine changes.

Hopefully this helps somehow - I'd really like to avoid shipping multiple platform dependent native libraries (and thanks for pointing out that x86 isn't supported, I scanned the PowerShell releases too quickly.)

SeeminglyScience: At this point I'm reasonably confident it isn't possible.  I've tried PInvoking `write(2)` and writing directly to the file descriptor 0 (stdin).  That comes out as output, which from what I've read appears to be by design (or at least to be expected).  Just to see if I could get *something* to work I tried inserting into various private buffers stored by `ConsolePal` and `StdInReader`, but it's already waiting on native code by the time `ReadKey` is blocking.

I really like the idea of redirecting and controlling all input, but that'd be a hefty undertaking for a (hopefully) temporary work around.  Ideally, this issue would be solved within corefx.

That said, did you have any concerns about the cancellation token changes from the first commit?  That's the only change required for PSReadLine to work within Windows.  I already have the logic to fall back to the current readline implementation for PowerShell 3/4.  If the busy loop and native libraries are enough of a blocker, I can remove the entire second commit from the PR and only support PSReadLine on Windows for the time being.

PSReadLine.build.ps1
parkovski: Any reason why this needs to be supplied as an arch-dependent binary? Why not use P/Invokes like is done for Windows with Get/SetConsoleMode?

PSReadLine/ConsoleLib.cs
parkovski: I'd like to look into this. There's still an issue with cursor position escapes when pwsh.exe is run from WSL, and the only way I could get close to making it work is by manually buffering the console input. I think if I can get that working, I'd have a version of `ReadKey` that blocks on an event and not the actual console.

SeeminglyScience: That'd be amazing @parkovski.  If you can figure that out you'd be my hero.

Some more information, [here's the PR](https://github.com/PowerShell/PowerShellEditorServices/pull/592) for editor services where I go into a little bit more detail on the cause.  And here's [the issue](https://github.com/PowerShell/vscode-powershell/issues/987) where I go through some of the things I tried. Good luck :)

SeeminglyScience: @lzybkr I'm not sure how much better it is, but I do have another thought.  If a better implementation isn't possible/feasible, maybe PSReadLine could surface an API to customize `ReadKey`.  Editor Services will need to ship these native binaries and use the busy loop either way, if we could give PSReadLine a `ReadKey` delegate then we could drop the second commit and still support Linux/Mac.  That wouldn't fix the event logic outside of PSES though.

PSReadLine.build.ps1
SeeminglyScience: @parkovski ...well I wish I had a good explanation for why I originally decided it couldn't be done with pure P/Invokes.  But you're 100% correct, it definitely can.  I just wrote up something basic to test and it worked great.

Only needed to import `tcsetattr` and `tcgetattr` from libc. They should be available everywhere, yeah? Thank you, can't believe I over looked that.  I'll get started on switching to P/Invokes.

PSReadLine/ConsoleLib.cs
lzybkr: It's not public, but there is already an interface to mock `Console.ReadKey` for the tests. The tests just change the mock via reflection, so you could do the same as an experiment.

That said, `stdin` feels like it is that "delegate". Ensuring `stdin` works for nearly everything helps in scenarios beyond PSES. (I think a few bindings are impossible on Windows with VT escape sequences but are possible with `Console.ReadKey`.)

SeeminglyScience: I found a way to block until input is received.  I can p/invoke [poll(2)](http://man7.org/linux/man-pages/man2/poll.2.html), also from libc.  That has an option to block indefinitely until input is received.  It works, but it also triggers on `Console.CursorLeft` which spills some of the escape sequence into `Console.ReadKey`.  Granted this is the exact problem corefx was trying to solve by locking.

If I can figure out a good way to filter that out, coupled with replacing the native libs with p/invokes, that should hopefully solve all issues.

parkovski: The problem with filtering that out is whoever called `CursorLeft` will get an invalid result, which can make line editing glitchy, unless you can put it back in the input stream before anything else is read. It might work to make a new tty and pass the original input through a filter, but to make that reliable I think you'd have to predict the terminal's responses, which seems like not a good idea. I'm leaning toward the only way to solve this is by patching CoreCLR, unfortunately.

SeeminglyScience: @parkovski Yeah I'm operating under the assumption I can find a way to determine if poll was triggered because of `CursorLeft` without actually consuming the input.  If I can't, then I agree it's isn't feasible.

SeeminglyScience: Unfortunately I ran into more issues.  Changing the terminal settings via p/invoke isn't working as expected.  I think there is an additional setting or two that is getting changed because new lines aren't triggering input unless `CTRL` is held.

On top of that, it looks like the `termios` stuct can be very different between architectures.  Some have more members, some flags have different values, and the control characters array can vary in size.  For some of those, getting them wrong results in a pretty nasty crash. That said, my experience with p/invoke (and Linux in general) is pretty limited, so [here's the code](https://gist.github.com/SeeminglyScience/18a38347625c47c11017908d4dc5874d) I was testing with.  It's very possible I missed something, but I don't think p/invoke will be reliable in this instance.

For Editor Services I'm going to move forward with the native binaries (or `stty -echo`) and the long/short busy loop because the Unix debugging experience is completely broken without it.  But  PSReadLine in VSCode may need be Windows only until the issue is fixed in corefx.

@lzybkr How would you like to proceed?  Assuming you don't have any issues with the first commit, should I remove the second or open a new PR?

PSReadLine/ReadLine.cs
```diff
@@ -139,7 +142,12 @@ private void ReadKeyThreadProc()
                 if (handleId == 1) // It was the _closingWaitHandle that was signaled.
                     break;
 
+                var localCancellationToken = _singleton._cancelReadCancellationToken;
                 ReadOneOrMoreKeys();
+                if (localCancellationToken.IsCancellationRequested)
```
lzybkr, line 154: If cancellation is requested, what should happen with buffered keys (keys pressed after cancellation, but before the next call to PSReadLIne)?

And what should happen to a partially typed command line (before the cancellation, but no Enter) - should that be restored. If you look at how `AcceptAndGetNext` is implemented, you can see it's not hard to restore.

lzybkr (changes requested):
I have some comments and questions on the first commit.

I'm rejecting the second commit for now.

As for the approach, I mentioned this before, but I'm reluctant to add a new public api **if** it makes sense to switch to using standard input with a handshake character.

What about making it internal and "unsupported"? I think you might need reflection anyway to see if the api is available, so it shouldn't matter if it was internal. This way, it's a "private" contract between PSES and PSReadLine, subject to change, but not an officially supported api.

PSReadLine/ReadLine.cs
SeeminglyScience: If cancellation is requested it'll continue to the next loop and any buffered keys will remain in
`_queuedKeys` until the next key is requested.

Originally any partially typed command lines were just being discarded.  Now the
command line is saved and restored when `ReadLine` is called again.  Thanks for pointing out `AcceptAndGetNext`, I didn't know it would be that easy to implement and it feels *much* better now.

PSReadLine/ReadLine.cs
```diff
@@ -275,6 +290,18 @@ private void PrependQueuedKeys(ConsoleKeyInfo key)
...
         /// <returns>The complete command line.</returns>
         public static string ReadLine(Runspace runspace, EngineIntrinsics engineIntrinsics)
+        {
+            // Use a default cancellation token instead of CancellationToken.None because the
+            // WaitHandle is shared and could be triggered accidently.
+            return ReadLine(runspace, engineIntrinsics, _defaultCancellationToken);
+        }
+
+        /// <summary>
+        /// Entry point - called from the PowerShell function PSConsoleHostReadLine
```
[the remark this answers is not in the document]
SeeminglyScience: I updated the comment with reflect intended usage.  I kept the comment generic because it can be used by more than just the PSES host, but that's right we aren't using `ConsoleHost`.  The way `ReadLine` is being called is more or less the same as in `ConsoleHost`, just a custom script instead of the `PSConsoleHostReadLine` function.

PSReadLine/ReadLine.cs
```diff
@@ -304,6 +331,8 @@ public static string ReadLine(Runspace runspace, EngineIntrinsics engineIntrinsi
                         _singleton.Initialize(runspace, engineIntrinsics);
                     }
 
+                    _singleton._cancelReadCancellationToken = cancellationToken;
+                    _singleton._requestKeyWaitHandles[2] = _singleton._cancelReadCancellationToken.WaitHandle;
                     return _singleton.InputLoop();
                 }
                 catch (OperationCanceledException)
```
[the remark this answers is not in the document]
SeeminglyScience: The comment has been updated to reflect the new reason for the exception.

SeeminglyScience:
> What about making it internal and "unsupported"? I think you might need reflection anyway to see if the api is available, so it shouldn't matter if it was internal. This way, it's a "private" contract between PSES and PSReadLine, subject to change, but not an officially supported api.

I have absolutely no issues with that. @tylerl0706 @daviwil @rkeithhill feel free to chime in if you disagree, but I don't have any objections.

TylerLeonhardt:
Abstain. I don't know enough to feel strongly either way. 🙂 I trust your judgement.

rkeithhill:
No objections from me.

SeeminglyScience:
@lzybkr I know I've been absent too long and I appreciate your patience with me on this.

I've added the delegate we discussed.  I'm able to override it from PSES without issue.  There's still some bugs to work out with our `ReadKey` implementation but I've tested that the delegate we inject is the one that runs.  So the rest needs to be worked out on our side.

I've also added, in a separate commit, a method to force event handling to happen immediately instead of after the wait timeout in `PSConsoleReadLine.ReadKey`.  I added this because I'm using the checking for events to take over the pipeline without cancelling PSReadLine.  The most common thing that is invoked this way is intellisense, which felt very sluggish if the 300ms timeout had just started.

I added this as another private member that we would access via reflection, as I'm not sure it makes sense as a public API.

PSReadLine/ReadLine.cs
```diff
@@ -174,9 +183,10 @@ internal static ConsoleKeyInfo ReadKey()
                     //   - a key is pressed
                     //   - the console is exiting
                     //   - 300ms - to process events if we're idle
-
+                    //   - processing of events is requested externally
+                    //   - ReadLine cancellation is requested externally
                     handleId = WaitHandle.WaitAny(_singleton._requestKeyWaitHandles, 300);
-                    if (handleId != WaitHandle.WaitTimeout)
+                    if (handleId != WaitHandle.WaitTimeout && handleId != 3)
```
lzybkr: Can you add constants for each handleId and use those - it will make reading the code easier in the future.

lzybkr:
After replacing the integer constants with named constants, I think this is ready to merge.

PSReadLine/ReadLine.cs
SeeminglyScience: Great idea, that's a lot easier to read now.

I assumed you'd prefer I omit the constant for `_keyReadWaitHandle` as it's index isn't referenced. Let me know if that's incorrect though.

Issue #623, opened 2018-01-22
Fix vi-mode cf  behavior


reaching-for-libra:
When using <cf> to replace text in vi-mode, the cursor is often placed one character to the right of where it should.

`asdf asdf`
`<0>` to go to beginning of line
`<W>` to go to next word
`<cfs>` to remove 'as' and drop into insert mode.
the cursor is now after the 'd' instead of before it.

i think that this happens any time the cursor is not in the first character of the command string when starting the <cf> action.

`PS version: 6.0.0`
`PSReadline version: 2.0.0-beta1`
`os: 6.1.7601.17514 (win7sp1_rtm.101119-1850)`
`PS file version: 6.0.0`

srdubya:
Thanks for submitting this.  I'll take a look soon.

reaching-for-libra:
awesome - i just remembered a similar issue. if you are on the last character of the line in normal mode and you type `<s>`, it will put the cursor before what is now the last character (instead of placing it after)

srdubya:
Pushed a fix to both issues to the srdubya branch and submitted a pull request to lzybkr, but I've been struggling to keep the srdubya branch clean.  We'll see what happens.

PR #627, opened 2018-01-27
Fixes for #571 and #623

srdubya:
I hope this branch is in good enough shape for these fixes to be acceptable.


lzybkr:
@srdubya - I fixed your branch on GitHub, but I had to force push to get rid of the merge commits, and this means you can't do a simple `git pull` to your local clone without getting into this same state.

Assuming you have no other local work (no branches or commits in master that you didn't push) - then it's probably easiest to delete your local files and clone https://github.com/srdubya/PSReadLine.git again.

In the future, if you run `git pull` and it says it can't fast forward, you need to pull from upstream before committing anything.

Alternatively, you can create a branch, make your fix, and submit the branch for the PR - this should always merge cleanly.

srdubya:
Looked like the changes for 456 were missing, so I added them.

lzybkr:
I think I revisited 456 when fixing something else and replaced your fix. You added a test in #573 which is still there, so I don't think we need the last 4 commits.

lzybkr:
@srdubya - these changes were squashed, so you will need to force push to your master branch to avoid a merge commit.

It's probably easiest to create a branch for each PR and never commit directly to master. If you find yourself using `git merge`, you can use `--ff-only` to avoid merge commits in master - it will reject the merge if you can't fast forward.

PR #637, opened 2018-02-25
Fixed #628

### LINKED ISSUE #628: Ctrl+u should work in vi insert mode
Ctrl+u (BackwardDeleteLine) is a convenient way to clear the line - it doesn't quite work from Command mode because the cursor can't be 1 past the end, but it should work from insert mode, plus it works in bash.

srdubya:
OK, let's give this a try.


lzybkr (changes requested):
Looks good.

Can you also add one more test that clears the entire line?

lzybkr (approved):

PR #644, opened 2018-03-10
Use -ListAvailable for Get-Module PSReadline bug template

bergmeister:
My VM is a bit weird where some modules do not auto-load in some PowerShell instances (at least the first time before they are being used by a different host such as e.g. the ISE). Therefore the issue template did not work for me.


lzybkr:
This tells me every version of PSReadLine installed on the machine, but not the one currently being used. I suppose I could assume you're always using the latest, but I shouldn't assume such things.

bergmeister:
Ok, I see, I'll close then because although we could add logic to it, a repro script should not be too complicated

PR #648, opened 2018-03-16
Disable PSReadline on Unix when stdin or stdout is redirected

### LINKED ISSUE #625: ConsoleHost tests fail with interactive use and redirected stdin
ConsoleHost tests fail with redirected stdin with interactive use. The error seen is: `An error has occurred that was not properly handled. Additional information is shown below. The PowerShell process will exit`

This failure is not seen on PSReadLine v1.2.

These failures are only seen on non-windows platforms.

Environment data
----------------
PS version: 6.0.0-rc.2
os: Linux aditya-ubuntu 4.4.0-109-generic #132-Ubuntu SMP Tue Jan 9 19:52:39 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

Execute the tests at: https://github.com/PowerShell/PowerShell/blob/master/test/powershell/Host/ConsoleHost.Tests.ps1#L335

daxian-dbw:
Fix #625

`System.Console` doesn't handle redirected input on Unix Platforms. It matches the behavior on Windows by throwing an `InvalidOperationException`. Therefore, if either stdin or stdout is redirected, PSReadLine doesn't really work on Unix. So we throw `NotSupportedException` in such cases and let PowerShell call `Console.ReadLine` or do whatever else it decides to do.

### DIFF
--- a/PSReadLine/ReadLine.cs
+++ b/PSReadLine/ReadLine.cs
@@ -285,6 +285,15 @@ namespace Microsoft.PowerShell
             }
             else
             {
+                if (Console.IsInputRedirected || Console.IsOutputRedirected)
+                {
+                    // System.Console doesn't handle redirected input. It matches the behavior on Windows
+                    // by throwing an "InvalidOperationException".
+                    // Therefore, if either stdin or stdout is redirected, PSReadLine doesn't really work,
+                    // so throw and let PowerShell call Console.ReadLine or do whatever else it decides to do.
+                    throw new NotSupportedException();
+                }
+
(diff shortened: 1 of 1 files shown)


daxian-dbw:
I have verified that the console stdin redirection tests in PowerShell passed with this change.

lzybkr (approved):

daxian-dbw:
@lzybkr Please let me know once a new version of `PSRealine` is published to PSGallery, so I make changes to use it for PSCore by default. Thanks!

PR #653, opened 2018-03-21
Adding color for prompt text

powercode:
When the promt is drawn with a different
color than the default text, it renders ugly when
we set it back to another color than what was
originally set.
prompt

### DIFF
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -361,15 +361,22 @@ namespace Microsoft.PowerShell
             PlaceCursor(_initialX, _initialY);
 
             // Possibly need to flip the color on the prompt if the error state changed.
-            if (!string.IsNullOrEmpty(_options.PromptText))
+            var promptText = _options.PromptText;
+            if (!string.IsNullOrEmpty(promptText))
             {
                 renderData.errorPrompt = (_parseErrors != null && _parseErrors.Length > 0);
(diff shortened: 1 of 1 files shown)


lzybkr:
Did you try adding the color to PromptText, like this?

```powershell
Set-PSReadLineOption -PromptText "$e[96m>"
```

powercode:
Ah, clever :)

powercode:
Almost clever :)
That changes the length of the prompt text, so we change color 5 characters behind the prompt text.

lzybkr:
I'm not a fan of excessive options.

What do you think about fixing `PromptText` to ignore color escape sequences? PowerShell does this [here](https://github.com/PowerShell/PowerShell/blob/c1c5344a8897262433141ecbc13bb06ac2c4bbef/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs#L2752) - though I think that code doesn't handle CJK correctly - I think this [PR](https://github.com/PowerShell/PowerShell/pull/5739) fixes that.

powercode:
The fewer options the better for sure!
I'll take a look.

lzybkr (approved):
Thanks for the fix,

I was actually hitting the same issue in my personal admin prompt, but too busy to give it any thought so I appreciate you taking care of it.

PR #655, opened 2018-03-23
Fix the ExtraPromptLineCount option

### LINKED ISSUE #634: Set-PSReadlineOption -ExtraPromptLineCount does not work anymore in 2.0.0 Beta (master)
This is follow-up of #470

Environment data
----------------
PS version: 5.1.15063.786
os: 10.0.15063.0 (WinBuild.160101.0800)
PS file version: 10.0.15063.0 (WinBuild.160101.0800)

According to the documentation when using multi-line prompt, one needs to add `Set-PSReadlineOption -ExtraPromptLineCount <NumberOfExtraLines>`. While it worked in 1.2, it does not work anymore in 2.0.0 Beta (master), so Ctrl+L is removing first line of the prompt regardless of the setting.
Use following `profile.ps1` to replicate the bug:
~~~powershell
function Prompt {
    Write-Host $($ExecutionContext.SessionState.Path.CurrentLocation)
    return '>' * ($nestedPromptLevel + 1) + ' '
}
Set-PSReadlineOption -ExtraPromptLineCount 1
~~~

daxian-dbw:
Fix #634
Attempt to fix the `-ExtraPromptLineCount` for `ctrl+l`.

### DIFF
--- a/PSReadLine/Movement.cs
+++ b/PSReadLine/Movement.cs
@@ -394,16 +394,17 @@ namespace Microsoft.PowerShell
         public static void ClearScreen(ConsoleKeyInfo? key = null, object arg = null)
         {
             var console = _singleton._console;
-            if (_singleton._initialY + console.WindowHeight > console.BufferHeight)
+            int newY = _singleton._initialY - _singleton.Options.ExtraPromptLineCount;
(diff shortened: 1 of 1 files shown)


lzybkr (approved):
Thanks for the fix.

Issue #666, opened 2018-04-10
Readline don't display characters after keyboard layout change


powercode:
Characters that are written with modifiers other than shift, are not displayed if 
the user changes keyboard layout to a layout where the key has different modifiers
than in the layout that the process started with.

Seems to be a bug in recent Windows version, but PSReadline should never hide 
characters in 
```
`~!@#$%^&*()-_=+[{]}\\|;:'\",<.>/?
```

Of these, at lease the following are behind `AltGr` in Swedish
```
~ @ $ { } |
```
making it challenging to write powershell commands.

PR #667, opened 2018-04-10
Always display special language characters 

powercode:
Always display special characters from the PowerShell language.
Many of these are behind AltGr in many
European keyboard layouts, but should always
be displayed.


powercode:
#666

PR #668, opened 2018-04-11
Enabled Ctrl+LeftArrow and Ctrl+RightArrow on Linux and cloudshell

jianyunt:
Currently the (Get-PSReadLineOption).EditMode is set to "Emacs" on pwsh Linux. The lookup key_dispatchTable for Emacs does not have Ctrl+LeftArrow or RightArrow defined. Therefore we cannot use them to move to next word on Linux or Azure CloudShell.
The workaround is set the edit mode to Windows on Linux OS to make the key work. This is a bit odd.
This change to add Ctrl+LeftArrow and Ctrl+RightArrow  to the emacsBindings for Linux. I tested the changes on my Linux VM and worked fine.

### DIFF
--- a/PSReadLine/KeyBindings.cs
+++ b/PSReadLine/KeyBindings.cs
@@ -269,6 +269,8 @@ namespace Microsoft.PowerShell
                 { Keys.ShiftEnter,      MakeKeyHandler(AddLine,              "AddLine") },
                 { Keys.LeftArrow,       MakeKeyHandler(BackwardChar,         "BackwardChar") },
                 { Keys.RightArrow,      MakeKeyHandler(ForwardChar,          "ForwardChar") },
+                { Keys.CtrlLeftArrow,   MakeKeyHandler(BackwardWord,         "BackwardWord") },
+                { Keys.CtrlRightArrow,  MakeKeyHandler(NextWord,             "NextWord") },
                 { Keys.ShiftLeftArrow,  MakeKeyHandler(SelectBackwardChar,   "SelectBackwardChar") },
                 { Keys.ShiftRightArrow, MakeKeyHandler(SelectForwardChar,    "SelectForwardChar") },
                 { Keys.UpArrow,         MakeKeyHandler(PreviousHistory,      "PreviousHistory") },


lzybkr:
I don't think I should take this change.

Linux users do not expect <kbd>Ctrl+LeftArrow</kbd> or <kbd>Ctrl+RightArrow</kbd> in Emacs mode, and in fact those keys do not work at all, they are indistinguishable from <kbd>LeftArrow</kbd> or <kbd>RightArrow</kbd>. at least on the versions of Linux that I've tried.

jianyunt:
@lzybkr are you saying Linux users do not expect ctrl+leftarrow and rightarrow keys? If so, that's fine with me for these keys unavailable on cloudshell.

On the other hand, I tried the changes on my Linux VM and CloudShell bash +pwsh, they move forward word by word as expected. Also you may try `Set-PSReadLineOption -EditMode Windows` on Linux, the keys work as expected.

jianyunt:
BTW, on bash ctrl+rightarrow or ctrl+leftarrow moves the cursor forward or backward one word by one word. Hence Linux users may expect these keys to work. Currently on pwsh, the cursor does not move.

lzybkr:
In Emacs mode (the default for Linux to be more similar to bash) users expect <kbd>Alt+b</kbd> for `BackwardWord` and <kbd>Alt+f</kbd> for `ForwardWord`.

The `Ctrl` binding would work if the terminal supports it, but not all do. On the VM I tried, the `Ctrl` bindings moved a single character, not a word, just as though `Ctrl` was not pressed.

I don't have any logic to detect if a binding could work and I'd prefer not adding a binding that is unexpected and doesn't work everywhere, so I'm closing this without merging.

lol768:
Every single terminal I've ever used in every single Linux distribution I've ever used has supported Ctrl + Left arrow to move left a word and Ctrl + Right arrow to move right a word.

I've just confirmed with gnome-terminal, xfce4-terminal, alacritty and konsole using bash and zsh.

Truly a bizarre decision.

Jason - where did you try it before you confirmed it didn't work?

lzybkr:
I don't recall where I tested, likely Ubuntu 18, maybe CentOS or Arch?

I use `showkey -a` and from PowerShell `[Console]::ReadKey()` to check if a terminal sends distinguishable characters for a particular key press and if PowerShell (via .Net) can detect those differences.

As my testing was nearly three years ago, something may have changed. Maybe can @daxian-dbw revisit this.

rkeithhill:
This is what I see from `Ctrl+LeftArrow` and `Ctrl+RightArrow` on Ubuntu 18.04:
```
^[[1;5D          27 0033 0x1b
         91 0133 0x5b
         49 0061 0x31
         59 0073 0x3b
         53 0065 0x35
         68 0104 0x44
^[[1;5C          27 0033 0x1b
         91 0133 0x5b
         49 0061 0x31
         59 0073 0x3b
         53 0065 0x35
         67 0103 0x43
 ```

PR #675, opened 2018-04-19
Custom text for vi mode indicator

parkovski:
For changing the cursor via escape codes (see commit message, these codes are now supported on Windows skip ahead), or whatever else you want to print to the screen as an indicator.

Note - when using Get-PSReadLineOption, it looks like the codes are unset. They would need to be unescaped to show up. Does PowerShell have a canonical way of doing this?

Also add Shift-Tab to Windows ANSI processing which I had missed.

### DIFF
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -403,6 +403,14 @@ namespace Microsoft.PowerShell
                 }
 
                 lenLastPhysicalLine = columns % bufferWidth;
+                if (lenLastPhysicalLine == 0)
+                {
+                    // Handle the last column when the columns is equal to n * bufferWidth
+                    // where n >= 1 integers
+                    lenLastPhysicalLine = bufferWidth;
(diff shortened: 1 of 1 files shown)


parkovski:
Thought I had generated the help. Just found out I skipped the markdown file. I'll get it updated.

lzybkr (commented):
The changes look reasonable if we decide this is the right way to go.

I'm not excited about the new options - it seems like we should just default to what you document when possible, probably with something like `ViModeStyle.Default`.

Is it totally unreasonable to attempt detection of DECSCUSR support.

parkovski:
I don't know if it's detectable. I do know Konsole uses a different sequence though, and I have these sequences in my vimrc and zshrc, so I don't think they do any detection.

What about having a single mode change option that takes a scriptblock? That's what zsh does (zle-keymap-select), and it'd be the most flexible.

lzybkr:
A scriptblock sounds good - that would support adjusting the prompt as well.

PR #680, opened 2018-05-03
enable redirected stdin to work with PSCore6s

SteveL-MSFT:
Currently unable to move to PSReadLine 2.0.0 due to PSCore6 redirected stdin console tests failing.  This change checks in PSReadLine if input is redirected and uses different apis that work against stdin.

PSCore6 ConsoleHost.tests.ps1 pass on Linux, macOS, and Windows

### DIFF
--- a/PSReadLine/ConsoleLib.cs
+++ b/PSReadLine/ConsoleLib.cs
@@ -97,8 +97,32 @@ namespace Microsoft.PowerShell.Internal
             set { try { Console.OutputEncoding = value; } catch { } }
         }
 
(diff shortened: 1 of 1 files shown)


lzybkr:
Maybe this is good enough for tests, I have a feeling the user experience might be terrible, perhaps worse than without PSReadLine.

I'd more or less envisioned the code in WindowsAnsiCharMap being used instead. @parkovski wrote that code originally for the Windows console, but it should work for the most common non-Windows terminals with some minor changes.

SteveL-MSFT:
Didn't realize Dongbo already made a change to address this.

PR #688, opened 2018-05-22
Fixed PhysicalLineCount for proper rendering for a long line user input

### LINKED ISSUE #686: Extra characters are not cleared before rendering
Environment data
----------------

```powershell
PS version: 6.0.2
os: Linux cc-72f9-4d912258-4045611380-gqf6v 4.4.0-124-generic #148-Ubuntu SMP Wed May 2 13:00:18 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
PS file version:
```

1. Linux VM ( I am using Linux Ubuntu 16.,04 LTS)
2. install PowerShell core
3. install the latest psreadline 2.0.0-beta1
4. launch terminal and pwsh
5. ls, dir etc until your cursor reaches the bottom of your terminal
6. keep typing. When reach the third line,  you will see the second line gets replicated.

image

Alternatively you can use Azure cloudshell bash + pwsh, and  install PSReadline 2.0 to repro it.

Note: this does not happen on Windows, nor PSReadline 1.2.

In Cloudshell, user inputs can be long especially typing Azure PowerShell cmdlets. In addition, people can navigate to their Azure resources deep down, so prompt length can be long too. Thus this issue is manifested on the cloudshell easily. 

Initial Investigation
---------------------
On my Linux VM terminal, which has 80x24 size, the length of my prompt text is 44, after type 7 hellos (7*5 = 35), the total length is 44+35=79. This means if I type  one more `h`,  the `h` will appear in the current line but the cursor will go next line. As you can see from the debugger PhysicalLineCount() will return 1 and lenLastPhysicalLine = columns = 36  (7 hello + `h`). So far so good.

image

image

Now I continue typing hello until reach the end of the line: 

image

I got cnt = 2, columns = 116. After executing [the line]( https://github.com/lzybkr/PSReadLine/blob/master/PSReadLine/Render.cs#L396), columns will be 116-36=80.  After executing [the line]( https://github.com/lzybkr/PSReadLine/blob/master/PSReadLine/Render.cs#L405), lenLastPhysicalLine = 0, and  cnt + columns / bufferWidth = 2 + 80/80 = `3`.  The expected  physicalLine should be 2, right?  If I shift the x position by 1, i.e., changing [the line](https://github.com/lzybkr/PSReadLine/blob/master/PSReadLine/Render.cs#L396) to 

columns -= maxFirstLine **+ 1**;  then all works fine.

image

jianyunt:
Fix https://github.com/lzybkr/PSReadLine/issues/686

Without the fix, text does not get refreshed for a case where a user has a long line input. For 80 x 24 console, if you type 3 lines long text, you will see the second line is showed up on Line 3 before you even type anything.

### DIFF
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -393,7 +393,7 @@ namespace Microsoft.PowerShell
                     if (columns > maxFirstLine)
                     {
                         cnt += 1;
-                        columns -= maxFirstLine + 1;
+                        columns -= maxFirstLine;
                     }
                     else
                     {
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -393,7 +393,7 @@ namespace Microsoft.PowerShell
                     if (columns > maxFirstLine)
                     {
                         cnt += 1;
-                        columns -= maxFirstLine;
+                        columns -= maxFirstLine + 1;
                     }
                     else
                     {
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -403,6 +403,14 @@ namespace Microsoft.PowerShell
                 }
 
                 lenLastPhysicalLine = columns % bufferWidth;
+                if(lenLastPhysicalLine == 0)
+                {
+                    // Handle the last column when the columns is equal to n * bufferWidth
+                    // where n >= 1 integers
+                    lenLastPhysicalLine = bufferWidth;
+                    return cnt + (columns - 1) / bufferWidth;
+                }
+
                 return cnt + columns / bufferWidth;
             }
 
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -403,12 +403,12 @@ namespace Microsoft.PowerShell
                 }
 
                 lenLastPhysicalLine = columns % bufferWidth;
-                if(lenLastPhysicalLine == 0)
+                if (lenLastPhysicalLine == 0)
                 {
                     // Handle the last column when the columns is equal to n * bufferWidth
                     // where n >= 1 integers
                     lenLastPhysicalLine = bufferWidth;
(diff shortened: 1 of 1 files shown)


jianyunt:
It looks like the fix breaks UPArrow key for history as it lefts one extra char. Will take a look at tomorrow.

jianyunt:
Updated the PR. Now we can type a long command and hit UpArrow, DownArrow for history.  Tested the changes on Linux, Windows and CloudShell. Works fine.
@lzybkr, please review. thanks.

PSReadLine/Render.cs
```diff
@@ -403,6 +403,14 @@ int PhysicalLineCount(int columns, bool isFirstLogicalLine, out int lenLastPhysi
                 }
 
                 lenLastPhysicalLine = columns % bufferWidth;
+                if(lenLastPhysicalLine == 0)
+                {
+                    // Handle the last column when the columns is equal to n * bufferWidth
+                    // where n >= 1 integers
+                    lenLastPhysicalLine = bufferWidth;
+                    return cnt + (columns - 1) / bufferWidth;
```
lzybkr: I find `columns - 1` confusing here - I think `cnt - 1 + (columns / bufferWidth)` is equivalent, and if so, clearer.

PSReadLine/Render.cs
```diff
@@ -403,6 +403,14 @@ int PhysicalLineCount(int columns, bool isFirstLogicalLine, out int lenLastPhysi
                 }
 
                 lenLastPhysicalLine = columns % bufferWidth;
+                if(lenLastPhysicalLine == 0)
```
lzybkr: Nit - space after `if`.

PSReadLine/Render.cs
jianyunt: Updated. thx

PR #691, opened 2018-05-30
Fix crash and rendering where large content is pasted to console on non-Windows

### LINKED ISSUE #663: Exception on non-Windows w/ command line w/ more lines than visible console buffer
Environment data
----------------

```
PS version: 6.0.2
os: Darwin BrianShlettsMBP.attlocal.net 17.4.0 Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64 x86_64
PS file version:
```

This occurred while attempting to paste in the Get-IniContent function at https://gallery.technet.microsoft.com/scriptcenter/ea40c1ef-c856-434b-b8fb-ebd7a76e8d91
```
Last 200 Keys:
 Space Space } Enter
 Space Space Space Space W r i t e - V e r b o s e Space " $ ( $ M y I n v o c a t i o n . M y C o m m a n d . N a m e ) : : Space F i n i s h e d Space P r o c e s s i n g Space f i l e : Space $ F i l e P a t h " Enter
 Space Space Space Space R e t u r n Space $ i n i Enter
 Space Space } Enter
 Enter
 Space Space E n d Enter
 Space Space { Enter
 Space Space Space Space W r i t e - V e r b o s e Space " $ ( $ M y I n v o c a t i o n . M y C o m m a n d . N a m e ) : : Space F u n c t i o n Space e n d e d " Enter
 Space Space } Enter
 } Enter


Exception:
Parameter name: top
Actual value was -69.
   at Microsoft.PowerShell.PSConsoleReadLine.Insert(Char c)
   at Microsoft.PowerShell.PSConsoleReadLine.SelfInsert(Nullable`1 key, Object arg)
```

SteveL-MSFT:
Several issues:

1. When large content is pasted to the console window on Windows, the buffer scrolls and PSReadLine now has a negative value for the cursor.  Exception is raised when trying to set the cursor location to a negative y value.  Fix is to check for this and set y to 0 (top).
2. When the buffer scrolls on paste, the re-rendering gets confused as it didn't track the scrolling appropriately.  Fix is to skip drawing content that is off screen.
3. If the last line drawn to the screen is at the bottom of the screen buffer and the width of the output is exactly the window width, scroll doesn't happen until the next character is drawn; however, PSReadLine expected it to scroll so it's off by one.  Fix here is to catch this and reset position back one line.

Fix https://github.com/lzybkr/PSReadLine/issues/663

### DIFF
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -50,6 +50,9 @@ namespace Microsoft.PowerShell
             lines = new[] { new RenderedLineData{ columns = 0, line = ""}}
         };
         private int _initialX;
+
+        // This represents the initial Y position of the cursor.  If this position has scrolled off screen, this value will be negative indicating
+        // the number of lines off screen.
         private int _initialY;
         private ConsoleColor _initialForeground;
         private ConsoleColor _initialBackground;
@@ -436,8 +439,8 @@ namespace Microsoft.PowerShell
                 var lineData = renderLines[logicalLine];
                 var physicalLineCount = PhysicalLineCount(lineData.columns, logicalLine == 0, out var lenLastLine);
 
-                // If the initial position was scrolled off the screen, only output for lines still
-                // in the screen buffer
+                // If the initial position was scrolled off the screen, _initialY will be negative.
+                // Only output for lines still in the screen buffer.
                 if (_initialY + physicalLine >= 0)
                 {
                     // First time redrawing from the top
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -427,15 +427,42 @@ namespace Microsoft.PowerShell
             var logicalLine = 0;
             var physicalLine = 0;
             var lenPrevLastLine = 0;
+            var suppressTopLineIfScrolled = true;
 
             for (; logicalLine < renderLines.Length; logicalLine++)
             {
                 if (logicalLine != 0) _console.Write("\n");
 
                 var lineData = renderLines[logicalLine];
-                _console.Write(lineData.line);
+                var physicalLineCount = PhysicalLineCount(lineData.columns, logicalLine == 0, out var lenLastLine);
 
-                physicalLine += PhysicalLineCount(lineData.columns, logicalLine == 0, out var lenLastLine);
+                // If the initial position was scrolled off the screen, only output for lines still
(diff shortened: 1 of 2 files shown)


lzybkr:
I think the changes are fine, I'll try them out for a bit before merging.

Are you also going to fix `MenuComplete`? See this [comment](https://github.com/lzybkr/PSReadLine/issues/663#issuecomment-385506238).

SteveL-MSFT:
@lzybkr that's a separate issue (able to repro with a large directory listing with my changes).  I'll look at it in between other work as a separate PR.

SteveL-MSFT:
@lzybkr it comes up easily on non-Windows when pasting a large script

lzybkr:
It happens easily on Windows as well, but you need a small window.

For my education and to better understand an appropriate fix - when are small console windows normally hit on non-Windows? I know that with no gui the window is small, but where else?

SteveL-MSFT:
The default terminal size on Ubuntu is 80x24.

lzybkr:
Maybe it's reasonable to disallow a command line that exceeds the buffer height, at least until you can actually edit it properly.

SteveL-MSFT:
@lzybkr I think we should take this fix as even if you can't edit the buffer, it does show the syntax highlighting correctly making it easier to read.  If you think we should explicitly disallow editing, I think we should have that as a separate issue.  This PR does fix the current issue of an unhandled exception.

lzybkr:
I'm still on the fence about allowing `_initialY` to be negative. Before this PR, there was no code that expected that, this change basically takes the position that it is an expected condition, and we need to carefully review all other uses.

SteveL-MSFT:
There aren't too many references to `_initialY` in the code.  The `menu` code relies on `_initialY` and is probably the related crash for tab-complete for a large result set which I'll look into after we resolve this one.  The alternative requires using a different variable to track how far off the screen we are.

lzybkr:
If we decide a negative `_initialY` is reasonable, the field declaration **and** where we change it should have comments saying as much and why.

lzybkr:
It would be good to add a test as well. The current console proxy does use a fixed buffer, but it might need some improvements to handle this scenario.

PR #695, opened 2018-06-01
Add ScriptBlock vi mode indicator

parkovski:
I messed something up with git and it closed my old PR, so continuing from #675...

Changed the new option to take a script block and added an example to the docs of using it to change the cursor.

The official docs say you can use `$_` in the script block, but I could only get `$args[0]` to work. I also made the option an error unless the correct mode is set. The script block can't be set to null since that's used to determine whether the option was passed, but you can always set it to an empty script block or just change the mode.

### DIFF
--- a/PSReadLine/PSReadLine.format.ps1xml
+++ b/PSReadLine/PSReadLine.format.ps1xml
@@ -148,6 +148,11 @@ $d = [Microsoft.PowerShell.KeyHandler]::GetGroupingDescription($_.Group)
                 <PropertyName>ViModeIndicator</PropertyName>
               </ListItem>
               <ListItem>
+                  <Label>ViModeChangeHandler</Label>
+                  <ItemSelectionCondition><ScriptBlock>$null -ne $_.ViModeChangeHandler</ScriptBlock></ItemSelectionCondition>
+                  <PropertyName>ViModeChangeHandler</PropertyName>
+              </ListItem>
+              <ListItem>
                 <PropertyName>WordDelimiters</PropertyName>
               </ListItem>
               <ListItem>
--- a/docs/Set-PSReadLineOption.md
+++ b/docs/Set-PSReadLineOption.md
@@ -35,6 +35,7 @@ Set-PSReadLineOption
  [-WordDelimiters <string>]
  [-AnsiEscapeTimeout <int>]
  [-ViModeIndicator <ViModeStyle>]
+ [-ViModeChangeHandler <ScriptBlock>]
 ```
 
 ## DESCRIPTION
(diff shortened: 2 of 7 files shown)


parkovski:
Hmm :( AppVeyor doesn't like the way I embedded PowerShell in the markdown file - any ideas?

lzybkr:
You used a feature from a newer version of platyPS - I updated `appveyor.yml` in your PR and the build passed.

PSReadLine/PSReadLine.format.ps1xml
```diff
@@ -147,6 +147,10 @@ $d = [Microsoft.PowerShell.KeyHandler]::GetGroupingDescription($_.Group)
               <ListItem>
                 <PropertyName>ViModeIndicator</PropertyName>
               </ListItem>
+              <ListItem>
```
lzybkr, line 150: Since the value will almost always be `$null`, I think it should be excluded by default, which can be done with `ItemSelectionCondition`.

And I think the script block text will be truncated, so maybe this is more reasonable?

```
<ListItem>
    <Label>ViModeChangeHandler</Label>
    <ItemSelectionCondition><ScriptBlock>$null -ne $_.ViModeChangeHandler</ScriptBlock></ItemSelectionCondition>
    <PropertyName>ViModeChangeHandler</ScriptBlock>
</ListItem>
```

docs/Set-PSReadLineOption.md
```diff
@@ -540,6 +541,8 @@ Valid values are:
 
 -- Cursor - the cursor changes size
 
+-- Escape - user-specified text is printed
```
lzybkr: Script

lzybkr (changes requested):
Code changes look good.

Fix the `ViModeStyle` help text and try out my suggestion for the formatting.

There is no public api to set `$_` when invoking a script block, at least not a simple one. It's possible with `InvokeWithContext`, but I think it's wrong to use `$_` in this way because it is the pipeline variable.

docs/Set-PSReadLineOption.md
```diff
@@ -552,6 +555,35 @@ Accept pipeline input: false
...
+    if ($args[0] -eq 'Command')
+    {
+        Write-Host "`e[1 q"
+    }
+    else
+    {
+        Write-Host "`e[5 q"
+    }
+}
+```
```
vors: Hm, glad platyPS was able to handle that 😅 
Maybe it makes more sense to put that into Example section and reference from here?

lzybkr:
The changes look fine. I normally squash, but you really have 3 distinct changes here (Shift-Tab, appveyor update, and the vi mode change) - so I'd like history rewritten so I can just rebase instead of squash, this makes it easier when I update release notes.

lzybkr:
Closing and reopening to clear Azure Pipeline check.

lzybkr (approved):

PR #704, opened 2018-06-06
Fix System.ArgumentOutOfRangeException while get-tab

jianyunt:
In the https://github.com/lzybkr/PSReadLine/issues/663, @gshacklett has commented get- and then tab PSReadLine will throw. This case is very often on cloudshell because Azure cmdlets are quite long and people use tab a lot. 

The fix is to let contents display until complete for the menu not selected case. I tested on Windows, Linux and cloudshell, the fix seems to be working. Please code review. Thanks.

However I broke the PossibleCompletions() tests.  The fix changed the parameter in calling InvokePrompt https://github.com/lzybkr/PSReadLine/blob/master/PSReadLine/Completion.cs#L736 so code path is different. With the fix, we are executing the else block in https://github.com/lzybkr/PSReadLine/blob/master/PSReadLine/ReadLine.cs#L942.

### DIFF
--- a/PSReadLine/Completion.cs
+++ b/PSReadLine/Completion.cs
@@ -473,17 +473,25 @@ namespace Microsoft.PowerShell
                 }
             }
 
-            public void DrawMenu(Menu previousMenu)
+            public void DrawMenu(Menu previousMenu, bool menuSelect)
             {
                 IConsole console = Singleton._console;
 
-                // Move cursor to the start of the first line after our input.
-                this.Top = Singleton.ConvertOffsetToPoint(Singleton._buffer.Length).Y + 1;
-                EnsureMenuAndInputIsVisible(console, tooltipLineCount: 0);
+                if (menuSelect)
+                {
+                    // Move cursor to the start of the first line after our input.
+                    this.Top = Singleton.ConvertOffsetToPoint(Singleton._buffer.Length).Y + 1;
+                    EnsureMenuAndInputIsVisible(console, tooltipLineCount: 0);
 
-                console.CursorVisible = false;
-                console.SaveCursor();
-                console.SetCursorPosition(0, this.Top);
+                    console.CursorVisible = false;
+                    console.SaveCursor();
+                    console.SetCursorPosition(0, this.Top);
+                }
+                else
+                {
+                    // Start a new line to show the menu contents
+                    console.Write("\n");
+                }
 
                 var bufferWidth = console.BufferWidth;
                 var columnWidth = this.ColumnWidth;
@@ -522,8 +530,18 @@ namespace Microsoft.PowerShell
                     }
                 }
 
-                console.RestoreCursor();
-                console.CursorVisible = true;
+                if (menuSelect)
+                {
+                    console.RestoreCursor();
+                    console.CursorVisible = true;
+                }
+                else
+                {
+                    // Update the cursor position to the start of the first line after our input.
+                    _singleton._initialX = console.CursorLeft;
+                    _singleton._initialY = console.CursorTop;
+                    _singleton._previousRender = _initialPrevRender;
+                }
             }
 
             public void Clear()
@@ -732,8 +750,8 @@ namespace Microsoft.PowerShell
             }
             else
             {
-                menu.DrawMenu(null);
-                InvokePrompt(key: null, arg: menu.Top + menu.Rows);
+                menu.DrawMenu(null, menuSelect:false);
+                InvokePrompt(key: null, arg:null);
             }
         }
 
@@ -780,7 +798,7 @@ namespace Microsoft.PowerShell
             var userInitialCompletionLength = userCompletionText.Length;
 
             completions.CurrentMatchIndex = 0;
-            menu.DrawMenu(null);
+            menu.DrawMenu(null, menuSelect:true);
 
             bool processingKeys = true;
             int previousSelection = -1;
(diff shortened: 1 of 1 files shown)


lzybkr:
It looks like `MenuComplete` is still a problem and should be fixed at the same time.

It seems reasonable to fall back to this fix if the menu + the command line won't fit on the screen.

Also - when selecting items from the menu, the tooltip is displayed below the menu. That should be disabled if it won't fit on the screen.

jianyunt:
Thank you @lzybkr for your comments. Could you please give me more detailed info when you say "it looks like MenuComplete is still a problem "?

lzybkr:
If you invoke `MenuComplete`, e.g. bind <kbd>Tab</kbd> like this:

```powershell
Set-PSReadLineKeyBinding -Key Tab -Function MenuComplete
```

Then you will hit the call that tries to set the cursor position incorrectly. Your change avoids that call, but it is still reachable and many people do use `MenuComplete`.

jianyunt:
For the cursor position issue while hit the Tab, we also need to apply the fix from https://github.com/PowerShell/PowerShell/pull/7299/files (the next release of pscore will contain it).

Once applied the fix, invoking the MenuComplete should work fine. I tested it on Linux and CloudShell. I don't see the cursor issue.
Also if the tooltip does not fit the screen while a user selects a menu, PSReadline scroll up the screen first. It seems to work fine.

lzybkr:
I don't think that's correct. I have your to PSReadLine and shouldn't need the fix to PowerShell (no Write-Progress calls).

Try opening an instance of PowerShell, set the window height to 20, type <kbd>G</kbd><kbd>e</kbd><kbd>t</kbd><kbd>-</kbd><kbd>C</kbd><kbd>Ctrl+Space</kbd> then use the arrow keys until `Get-ChildItem` is selected.

After doing this, I see the same exception as the one you are trying to fix.

PSReadLine/Completion.cs
```diff
@@ -483,17 +483,25 @@ void EnsureMenuAndInputIsVisible(IConsole console, int tooltipLineCount)
...
-                console.SaveCursor();
-                console.SetCursorPosition(0, this.Top);
+                    console.CursorVisible = false;
+                    console.SaveCursor();
+                    console.SetCursorPosition(0, this.Top);
+                }
+                else
+                {
+                    // Start a new line to show the menu contents
+                    console.Write("\n");
```
lzybkr: This could write over some of the buffer if you press <kbd>Tab</kbd> anywhere that is not the last line, e.g.:
```
PS> & { Get-C<Tab><Tab>
>> }
```

After pressing <kbd>Tab</kbd> <kbd>Tab</kbd>, the screen might look like:

```
PS> & { Get-C<Tab><Tab>
Get-C1   Get-C2    Get-C3
PS> & { Get-C
>> }
```

When it should look like:

```
PS> & { Get-C<Tab><Tab>
>> }
Get-C1   Get-C2    Get-C3
PS> & { Get-C
>> }
```


```

Comment (jianyunt) [reply PSReadLine/Completion.cs]: Jason, I submitted the fix and tested it. It seems working fine now. Please take a look it again. Thanks!

File: PSReadLine/Completion.cs [hunk 2]
```diff
@@ -483,16 +483,24 @@ void EnsureMenuAndInputIsVisible(IConsole console, int tooltipLineCount)
...
+                if (menuSelect)
+                {
+                    EnsureMenuAndInputIsVisible(console, tooltipLineCount: 0, rows:this.Rows);
+
+                    console.CursorVisible = false;
+                    console.SaveCursor();
+                }
+                else
+                {
+                    EnsureMenuAndInputIsVisible(console, tooltipLineCount: 0, rows:0);
```
lzybkr: Is this call necessary? I suspected it wasn't and tested briefly with it removed, I didn't see any problems.

PSReadLine/Completion.cs
```diff
@@ -532,8 +540,18 @@ public void DrawMenu(Menu previousMenu)
...
-                console.RestoreCursor();
-                console.CursorVisible = true;
+                if (menuSelect)
+                {
+                    console.RestoreCursor();
+                    console.CursorVisible = true;
+                }
+                else
+                {
+                    // Update the cursor coordinates after showing the menu.
```
lzybkr: Is this block necessary? `InvokePrompt` is called immediately after the call site that passes `menuSelect: false` and `InvokePrompt` is correctly updating these fields.

I tested briefly with these commented out and didn't see any problems.

PSReadLine/Completion.cs
lzybkr: @jianyunt - after resolving this comment (either removing the call or confirming it is needed), I think this PR is ready to merge.

jianyunt (commented):
@lzybkr Yes we need to keep this Line https://github.com/lzybkr/PSReadLine/blob/2f57fe5091e0e0cd343b86851dc31c9bdcb2afd5/PSReadLine/Completion.cs#L501  to avoid the ArgumentOutOfRangeException exception.

PSReadLine/Completion.cs
jianyunt: @lzybkr Yes we need to keep this Line https://github.com/lzybkr/PSReadLine/blob/2f57fe5091e0e0cd343b86851dc31c9bdcb2afd5/PSReadLine/Completion.cs#L501  to avoid the ArgumentOutOfRangeException exception.

lzybkr: Are you relying on the `+ 1` to write out a single newline?

Based on the comment in `EnsureMenuAndInputIsVisible`, I could see that `+ 1` being removed at some point, and then the method would do nothing, right?

If my understanding is correct, it would be good to have some way to ensure this problem doesn't come back if someone removes that `+ 1` - a test, or maybe pass `rows:1`.

lzybkr (approved):

PR #708, opened 2018-06-08
Set the cursor to the place where a user hits tab key

jianyunt:
Fix issue  https://github.com/lzybkr/PSReadLine/issues/707

We need the fix from the PowerShell to completely fix the issue.

### DIFF
--- a/PSReadLine/Completion.cs
+++ b/PSReadLine/Completion.cs
@@ -226,6 +226,11 @@ namespace Microsoft.PowerShell
             }
 
             var completions = GetCompletions();
+            // GetCompletions can triager the PowerShell Write-Progress that may scroll the console screen.
+            // For example, cd and hit the tab under the CloudShell Azure drive, it will go out to fetch data and show the
+            // progress bar. We need to update the _initialY in case the current cursor postion has changed.
+             _singleton._initialY = Console.CursorTop;
+
             if (completions == null || completions.CompletionMatches.Count == 0)
                 return;
 
--- a/PSReadLine/Completion.cs
+++ b/PSReadLine/Completion.cs
@@ -229,7 +229,7 @@ namespace Microsoft.PowerShell
             // GetCompletions can triager the PowerShell Write-Progress that may scroll the console screen.
             // For example, cd and hit the tab under the CloudShell Azure drive, it will go out to fetch data and show the
             // progress bar. We need to update the _initialY in case the current cursor postion has changed.
-             _singleton._initialY = Console.CursorTop;
+             _singleton._initialY = _console.CursorTop;
 
             if (completions == null || completions.CompletionMatches.Count == 0)
                 return;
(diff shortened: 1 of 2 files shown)


jianyunt:
Test.ReadLine.ViDefect651 does not fail on my dev box. Any ideas?

lzybkr:
ViDefect651 passes for me as well with your change.

I can't think of anything obvious. You can debug in AppVeyor, see https://www.appveyor.com/docs/how-to/rdp-to-build-worker/

PSReadLine/Completion.cs
```diff
@@ -226,6 +226,11 @@ private void CompleteImpl(bool menuSelect)
             }
 
             var completions = GetCompletions();
+            // GetCompletions can triager the PowerShell Write-Progress that may scroll the console screen.
+            // For example, cd and hit the tab under the CloudShell Azure drive, it will go out to fetch data and show the
+            // progress bar. We need to update the _initialY in case the current cursor postion has changed.
+             _singleton._initialY = _console.CursorTop;
```
lzybkr: This scenario is very similar to the `InvokePrompt` scenario where somebody explicitly wants to call their prompt function to update some status or whatever - and after that, the cursor may have moved.

In the `InvokePrompt` scenario, we render the entire command line (all lines) because either the cursor moved or we erased the previous prompt.

I think this scenario is equivalent, and if I'm correct, I think the code from `InvokePrompt` should be extracted into a function so it can be shared.

The code in question is:

```c#
            _singleton._initialX = console.CursorLeft;
            _singleton._initialY = console.CursorTop;
            _singleton._previousRender = _initialPrevRender;

            _singleton.Render();
```

jianyunt: @lzybkr Yes this scenario is very similar to the `InvokePrompt`.  Both may change the cursor position. And for the tab Completion case, if a user do `get<tab>`, the cursor is expected to set back to right after `get`, right? This means _initialX has no change. If we update `_singleton._initialX = console.CursorLeft`, we get `getGet` back.

lzybkr: Agreed, `_initialX` shouldn't change in your scenario, so now I'm just wondering if you might need the other 2 lines to render properly.

Can you test all the ways `GetCompletions` is called - I can think of 4 - `Complete`, `TabCompleteNext`, `PossibleCompletions`, and `MenuComplete`. I would test all of those with the completion on the first line, and again with the completion on a second line (multi-line input), and also where the completion is near the end of the line.

If the display is fine in all of those scenarios without calling `Render` (in other words, exactly how the code is now), then we can call it good.

jianyunt: Thanks. The multi-line case does not work. It becomes something like this. Any advices?

image

jianyunt: Updated the code and tested on Windows and Linux with the following cases:
Complete, PossibleCompletions, MenuComplete, TabCompletNext, multipleline such as start-job{..}, close to end of the line. Works as expected now.

jianyunt:
@lzybkr,  I updated the code and tested. Could you please take look at it if we are good to go? Thanks!

lzybkr (approved):

PR #711, opened 2018-06-11
Adding space as an `essential` PS character.

powercode:
Space should be added even if a modifier is added.

Currently, Shift+Space is eaten by PSReadline, which is different from all other editors I've tried.
Notepad, Code, VS etc all input a space even if shift is pressed.

### DIFF
--- a/PSReadLine/Keys.cs
+++ b/PSReadLine/Keys.cs
@@ -553,7 +553,7 @@ namespace Microsoft.PowerShell
         {
             var keyChar = key.KeyChar;
             if (keyChar == '\0') return false;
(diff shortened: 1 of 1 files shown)


webash:
When will this change hit live bits? This has been an incredibly frustrating few months having this broken.

powercode:
@webash You can always build your own and drop the dll in as a stop-gap. That's what I did. The beauty of open source.

gwojan:
Why not just pull the latest build from Appveyor?

lzybkr:
@webash - this fix is available in [2.0.0-beta3](https://www.powershellgallery.com/packages/PSReadLine/2.0.0-beta3).

PR #757, opened 2018-09-06
Fix a default value for the HistorySavePath

alexandair:
Fix a default value for the HistorySavePath.

### DIFF
--- a/docs/Set-PSReadLineOption.md
+++ b/docs/Set-PSReadLineOption.md
@@ -500,7 +500,7 @@ Aliases:
 
 Required: False
 Position: Named
-Default value: A file named $($host.Name)_history.txt in $env:APPDATA\Microsoft\Windows\PowerShell\PSReadLine on Windows, $env:XDG_DATA_HOME/powershell/PSReadLine on macOS, and $env:HOME/.local/share/powershell/PSReadLine on Linux
+Default value: A file named $($host.Name)_history.txt in $env:APPDATA\Microsoft\Windows\PowerShell\PSReadLine on Windows and $env:XDG_DATA_HOME/powershell/PSReadLine or $env:HOME/.local/share/powershell/PSReadLine on non-Windows platforms
 Accept pipeline input: false
 Accept wildcard characters: False
 ```
--- a/docs/Set-PSReadLineOption.md
+++ b/docs/Set-PSReadLineOption.md
@@ -500,7 +500,10 @@ Aliases:
 
 Required: False
 Position: Named
-Default value: On Windows - $env:APPDATA\Microsoft\Windows\PowerShell\PSReadLine\$($host.Name)_history.txt otherwise $XDG_DATA_HOME/powershell/PSReadLine/$($host.Name)_history.txt or $HOME/.local/share/powershell/PSReadLine/$($host.Name)_history.txt
+Default value: A file named $($host.Name)_history.txt in: 
+$env:APPDATA\Microsoft\Windows\PowerShell\PSReadLine on Windows  
+$env:XDG_DATA_HOME/powershell/PSReadLine on macOS
+$env:HOME/.local/share/powershell/PSReadLine on Linux
 Accept pipeline input: false
 Accept wildcard characters: False
 ```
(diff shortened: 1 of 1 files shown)


docs/Set-PSReadLineOption.md
```diff
@@ -500,7 +500,7 @@ Aliases:
 
 Required: False
 Position: Named
-Default value: On Windows - ~\AppData\Roaming\Microsoft\Windows\PowerShell\PSReadLine\$($host.Name)_history.txt otherwise $XDG_DATA_HOME/powershell/PSReadLine/$($host.Name)_history.txt or $HOME/.local/share//powershell/PSReadLine/$($host.Name)_history.txt
+Default value: On Windows - $env:APPDATA\Microsoft\Windows\PowerShell\PSReadLine\$($host.Name)_history.txt otherwise $XDG_DATA_HOME/powershell/PSReadLine/$($host.Name)_history.txt or $HOME/.local/share/powershell/PSReadLine/$($host.Name)_history.txt
```
lzybkr: I suppose you should use PowerShell syntax for the non-Windows use of environment variables too.

alexandair: I wanted to, but I couldn't find an environment variable on Linux that points to $HOME/.local/share.

lzybkr: And I wonder if it would be more readable if `$($host.Name)_history.txt` was not repeated 3 times, e.g.

> Default value: A file named $($host.Name)_history.txt in $env:APPDATA\Microsoft\Windows\PowerShell\PSReadLine on Windows otherwise $env:XDG_DATA_HOME/powershell/PSReadLine or $env:HOME/.local/share/powershell/PSReadLine

lzybkr: Is $HOME not the environment variable?

lzybkr: It is - verified checking the code.

alexandair: It is and you use it already.

alexandair: ah, I got it.

alexandair: Linux has $HOME and $env:HOME. Is it the same on macOS for $env:XDG_DATA_HOME and $XDG_DATA_HOME?

alexandair:
I've used environment variables everywhere. And changed a formatting a little. I hope it doesn't break. Usually default value is just one line.

Update: It looks like a build is failing because of $env part of the value.

docs/Set-PSReadLineOption.md
```diff
@@ -500,7 +500,7 @@ Aliases:
 
 Required: False
 Position: Named
-Default value: On Windows - ~\AppData\Roaming\Microsoft\Windows\PowerShell\PSReadLine\$($host.Name)_history.txt otherwise $XDG_DATA_HOME/powershell/PSReadLine/$($host.Name)_history.txt or $HOME/.local/share//powershell/PSReadLine/$($host.Name)_history.txt
+Default value: A file named $($host.Name)_history.txt in $env:APPDATA\Microsoft\Windows\PowerShell\PSReadLine on Windows, $env:XDG_DATA_HOME/powershell/PSReadLine on macOS, and $env:HOME/.local/share/powershell/PSReadLine on Linux
```
lzybkr: The code literally checks XDG_DATA_HOME and HOME on non-Windows platforms, and I think the code is correct in that XDG_DATA_HOME is used on some Linux variants and HOME is set on macOS, so either variable should work on both macOS and Linux.

alexandair: Is this OK then:
A file named $($host.Name)_history.txt in $env:APPDATA\Microsoft\Windows\PowerShell\PSReadLine on Windows and $env:XDG_DATA_HOME/powershell/PSReadLine, or $env:HOME/.local/share/powershell/PSReadLine on non-Windows platforms

lzybkr: Yes.

lzybkr (approved):

PR #767, opened 2018-09-27
Fix build in VS

lzybkr:


lzybkr:
#692 addresses this as well, so I'll just wait for that PR to be ready.

PR #771, opened 2018-10-02
Fix dynamic changing of font in Windows console on CJK codepages

SteveL-MSFT:
On Windows, the outputencoding is tied to a codepage which is tied to a font.  When using a CJK locale, the default is using a raster font that can render the CJK glyphs.  Previously, PSReadLine 2.0.0 was manipulating the OutputEncoding to utf8 which resulted in a visible font change in the console and the resulting rendering in the screen buffer was not correct.

The fix is to detect that a raster font is being used and not change the output encoding.  Also in cases where the output encoding is already utf8, there's no need to change it.

### DIFF
--- a/PSReadLine/PlatformWindows.cs
+++ b/PSReadLine/PlatformWindows.cs
@@ -368,13 +368,18 @@ static class PlatformWindows
             && SetConsoleMode(handle, mode | ENABLE_VIRTUAL_TERMINAL_PROCESSING);
     }
 
+    internal static bool IsConsoleInput()
+    {
+        var handle = GetStdHandle((uint)StandardHandleId.Input);
+        return GetFileType(handle) == FILE_TYPE_CHAR;
+    }
+
     private static bool IsHandleRedirected(bool stdin)
     {
         var handle = GetStdHandle((uint)(stdin ? StandardHandleId.Input : StandardHandleId.Output));
 
         // If handle is not to a character device, we must be redirected:
-        int fileType = GetFileType(handle);
-        if ((fileType & FILE_TYPE_CHAR) != FILE_TYPE_CHAR)
+        if (GetFileType(handle) != FILE_TYPE_CHAR)
             return true;
 
         // Char device - if GetConsoleMode succeeds, we are NOT redirected.
--- a/PSReadLine/PlatformWindows.cs
+++ b/PSReadLine/PlatformWindows.cs
@@ -93,6 +93,41 @@ static class PlatformWindows
         return false;
     }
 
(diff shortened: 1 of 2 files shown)


lzybkr:
If PowerShell is not running in a console, this code probably won't work as expected - for example if we are running under ConPTY, so we should avoid checking the font in that case.

I *think* you can check this easily enough by calling [GetFileType](https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-getfiletype) on the file handle from STD_INPUT_HANDLE. If the return value is not FILE_TYPE_CHAR, then we can skip calling checking the font.

PSReadLine/PlatformWindows.cs
```diff
@@ -333,13 +368,18 @@ private static bool SetConsoleOutputVirtualTerminalProcessing()
...
+        return GetFileType(handle) == FILE_TYPE_CHAR;
+    }
+
     private static bool IsHandleRedirected(bool stdin)
     {
         var handle = GetStdHandle((uint)(stdin ? StandardHandleId.Input : StandardHandleId.Output));
 
         // If handle is not to a character device, we must be redirected:
-        int fileType = GetFileType(handle);
-        if ((fileType & FILE_TYPE_CHAR) != FILE_TYPE_CHAR)
```
SteveL-MSFT, line 342: Per the [documentation](https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-getfiletype), `GetFileType()` returns a single value and is not flags.

PSReadLine/ReadLine.cs
```diff
@@ -647,7 +651,19 @@ private void Initialize(Runspace runspace, EngineIntrinsics engineIntrinsics)
             _statusIsErrorMessage = false;
 
             _initialOutputEncoding = _console.OutputEncoding;
-            _console.OutputEncoding = Encoding.UTF8;
+
+            // Don't change the OutputEncoding if already UTF8 or using raster font on Windows
```
lzybkr: I think this is clearer if you invert the condition, so something like this:

```c#
_skipOutputEncodingChange = _initialOutputEncoding == Encoding.UTF8
    || (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
        && PlatformWindows.IsConsoleInput()
        && PlatformWindows.IsUsingRasterFont());

if (!_skipOutputEncodingChange) {
    _console.OutputEncoding = Encoding.UTF8;
}
```

lzybkr (approved):

PR #785, opened 2018-10-23
Save/restore initial fg/bg colors

### LINKED ISSUE #774: [console]::backgroundcolor gets reset by PSReadLine
Environment data
----------------

PS version: 5.1.18252.1002
os: 10.0.18252.1002 (WinBuild.160101.0800)
PS file version: 10.0.18252.1002 (WinBuild.160101.0800)

```powershell
[console]::backgroundcolor = "black"
cls
```

Expected:
background to be black

Actual:
background is still dark blue

Removing PSReadLine makes this work correctly

lzybkr:
Writing CSI 0 m resets the colors to the terminals initial colors, not
what the settings are before PSReadLine starts.

Fix #774

### DIFF
--- a/PSReadLine/ReadLine.cs
+++ b/PSReadLine/ReadLine.cs
@@ -433,6 +433,8 @@ namespace Microsoft.PowerShell
                         if (!_singleton._closingWaitHandle.WaitOne(0))
                         {
                             console.OutputEncoding = _singleton._initialOutputEncoding;
+                            console.ForegroundColor = _singleton._initialForeground;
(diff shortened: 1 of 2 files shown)


DHowett-MSFT:
How does this fix work on non-Windows, where the console colors cannot be read? Will we try to restore the state to something incomprehensible?

lzybkr:
Ah, right, I knew I was forgetting something.

lzybkr:
Non-Windows fixed via https://github.com/lzybkr/PSReadLine/commit/16ab1c6afe237dc6f38743c44d3047d498bf83bf

PR #802, opened 2018-11-08
Fix rendering issue when menu complete shifts due to length of completion

SteveL-MSFT:
When a menu item renders longer than the current line and takes up a second row, the menu shifts down.  When a shorter item is selected that fits within the line, the menu shifts back up resulting in the last row to be left on the screen.  Fix is to detect when the menu shifts and clear out the rest of the previous last line and blank out the last line.

Fix https://github.com/lzybkr/PSReadLine/issues/689

### DIFF
--- a/PSReadLine/Completion.cs
+++ b/PSReadLine/Completion.cs
@@ -519,7 +519,7 @@ namespace Microsoft.PowerShell
                 if (bufferEndPoint.Y < PreviousTop)
                 {
                     console.BlankRestOfLine();
-                    Singleton.WriteBlankLines(1);
+                    Singleton.WriteBlankLines(PreviousTop - bufferEndPoint.Y);
                 }
 
                 PreviousTop = bufferEndPoint.Y;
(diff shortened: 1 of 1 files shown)


PSReadLine/Completion.cs
```diff
@@ -513,6 +515,15 @@ public void DrawMenu(Menu previousMenu, bool menuSelect)
                     }
                 }
 
+                // if the menu has moved, we need to clear the lines under it
+                if (bufferEndPoint.Y < PreviousTop)
+                {
+                    console.BlankRestOfLine();
+                    Singleton.WriteBlankLines(1);
```
lzybkr: In an extreme case, we might need multiple blank lines, right? Like `PreviousTop - bufferEndPoint.Y`?

SteveL-MSFT: It's possible, although highly unlikely.  I can make the change.

lzybkr (approved):

Issue #795, opened 2018-11-06
Supporting '_' or '^' and '$' to move to the beginning and end of the logical line in Vi mode


springcomp:
Currently, the <kbd>^</kbd> key moves the cursor to the first non-blank character of the input.
Likewise, the <kbd>$</kbd> key moves the cursor to the very end of the input.

I would prefer those keys to be restricted to movements in the current logical line of a multi-line buffer.

![vi_movements](https://user-images.githubusercontent.com/8488398/48083917-b2a28580-e1f6-11e8-91e0-85e55383900e.gif)

I have prepared an implementation to this effect, but this touches the `GotoFirstNonBlankOfLine` and `MoveToEndOfLine`methods.

https://github.com/lzybkr/PSReadLine/blob/274381a89e937d70d9fb1b4f77c7cbf38909db8e/PSReadLine/Movement.vi.cs#L194-L207

https://github.com/lzybkr/PSReadLine/blob/274381a89e937d70d9fb1b4f77c7cbf38909db8e/PSReadLine/Movement.vi.cs#L150-L156

What I'm concerned about is that those methods are public and probably relied on in existing scripts. Furthermore, the documentation for the `MoveToEndOfLine` method specifically states that it moves the cursor at the end of the **input**.

What could be the best way to introduce such a change?

lzybkr:
I think it's very uncommon to use methods like these as an api - they are more often used for simple remapping of key bindings.

So from that perspective, I think it's relatively safe to change these methods, but it wouldn't hurt to wait a bit for some feedback before merging a change.

I'm also fine with you opening a PR and maybe someone will chime in on that - we can sit on the PR for some time as well.

springcomp:
All right, thanks. That seems reasonable.

PR #812, opened 2018-11-11
Supports '_' and '$' to move to the beginning and end of the logical line in Vi mode.

springcomp:
As discussed in #795 this is a PR that constraints the <kbd>^</kbd> and <kbd>$</kbd> keys to the current logical line, instead of to the entire buffer.

This PR also supports using '_' for moving at the beginning of the logical line as the <kbd>^</kbd> key does.

On writing this PR, I find myself struggling a lot with the current behavior of PSRL which allows the cursor to move past the last character of a logical line. I think that, in Vi mode, the cursor is never allowed to move past the end of a logical line, except when appending to a line with <kbd>a</kbd> or <kbd>A</kbd>.

For instance, when at the beginning of a line, the <kbd>h</kbd> key will move to the end of the previous line (debatable). But instead of moving the cursor on the last character of the previous line, it moves the cursor on the blank space past the end of that previous line.

There are many cases like this, such as moving up from a long line to the previous shorter line, etc.

I have work in progress to make it more Vi-like but I feel there should be some discussions and feedback on this. Would someone familiar with the Vi-mode design chime in. I would love to discuss those aspects.

### DIFF
--- a/PSReadLine/Movement.vi.cs
+++ b/PSReadLine/Movement.vi.cs
@@ -208,24 +208,10 @@ namespace Microsoft.PowerShell
         /// </summary>
         public static void GotoFirstNonBlankOfLine(ConsoleKeyInfo? key = null, object arg = null)
         {
-            if (_singleton.LineIsMultiLine())
+            var newCurrent = GetFirstNonBlankOfLogicalLinePos(_singleton._current);
+            if (newCurrent != _singleton._current)
             {
-                var newCurrent = GetFirstNonBlankOfLogicalLinePos(_singleton._current);
-                if (newCurrent != _singleton._current)
-                {
-                    _singleton.MoveCursor(newCurrent);
-                }
-            }
-            else
-            {
-                for (int i = 0; i < _singleton._buffer.Length; i++)
-                {
-                    if (!Char.IsWhiteSpace(_singleton._buffer[i]))
-                    {
-                        _singleton.MoveCursor(i);
-                        return;
-                    }
-                }
+                _singleton.MoveCursor(newCurrent);
             }
         }
 
--- a/PSReadLine/Movement.vi.cs
+++ b/PSReadLine/Movement.vi.cs
@@ -152,7 +152,18 @@ namespace Microsoft.PowerShell
         /// </summary>
         public static void MoveToEndOfLine(ConsoleKeyInfo? key = null, object arg = null)
         {
-            _singleton.MoveCursor(Math.Max(0, _singleton._buffer.Length + ViEndOfLineFactor));
+            if (_singleton.LineIsMultiLine())
+            {
+                var eol = GetEndOfLogicalLinePos(_singleton._current);
+                if (eol != _singleton._current)
+                {
+                    _singleton.MoveCursor(eol);
+                }
+            }
+            else
+            {
+                _singleton.MoveCursor(Math.Max(0, _singleton._buffer.Length + ViEndOfLineFactor));
+            }
         }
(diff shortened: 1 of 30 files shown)


lzybkr:
I believe @srdubya tried to faithfully emulate bash readline behavior, but I'm not sure how well bash supports multi-line.

PSReadLine/Movement.vi.cs
```diff
@@ -196,12 +208,23 @@ public static void GotoColumn(ConsoleKeyInfo? key = null, object arg = null)
...
-                if (!Char.IsWhiteSpace(_singleton._buffer[i]))
+                var newCurrent = GetFirstNonBlankOfLogicalLinePos(_singleton._current);
+                if (newCurrent != _singleton._current)
                 {
-                    _singleton.MoveCursor(i);
-                    return;
+                    _singleton.MoveCursor(newCurrent);
+                }
+            }
+            else
```
lzybkr: Is this branch still needed, or would your new code handle single lines correctly as well?

PSReadLine/Position.cs
```diff
@@ -35,5 +34,55 @@ private static int GetBeginningOfLinePos(int current)
...
+
+            return newCurrent;
+        }
+
+        private static bool IsBlank(int newCurrent)
+        {
+            var c = _singleton._buffer[newCurrent];
+
+            return Char.IsWhiteSpace(c)
+                && c != '\r'
```
lzybkr: It looks like you would more faithfully emulate `vim` if you used `char.IsControl(c)` instead.

PSReadLine/Movement.vi.cs
springcomp: You are right. The multiline case handles the more general single-line case as well.

PSReadLine/Position.cs
springcomp: I think my code is correct.

If I were to exclude all control characters, I would also exclude TAB which should be considered a "blank" character in that case.

Besides, `Char.IsWhiteSpace` handles many control characters but I just want to exclude new line characters. For completeness, this code excludes `\r` which cannot probably be input via PowerShell, though.

lzybkr: I wouldn't say your code is wrong, but I quickly tried a few random control characters in vim and `_` skipped over them. I did think it looked weird that vim skipped over characters that render as non-spaces, but it did.

It's not easy to insert control characters today, but it is possible. At some point somebody will ask for `Ctrl+q` support and then it will be easy to insert control characters.

lzybkr:
Please don't force push, it makes it harder to see the updates. I typically squash the PR anyway, so a messy history isn't a big deal.

And on that note, you have merge commits in your fork in master. I think that's why this PR shows 9 commits when it should show just one. You should force push your upstream/master to your origin/master and avoid merges into your master.

springcomp:
Thanks for the feedback.
I'm trying to learn how to deal with synchronizing forks.
I'll look into resources to improve my workflow.

I would be happy to recreate the PR properly if you like.

lzybkr:
I don't need a new PR, the diffs look fine and squashing cleans up the history.

PSReadLine/Position.cs
lzybkr: Maybe if the method was named `IsVisibleBlank` or `IsHorizontalWhitespace`, then the intention and code is clear.

lzybkr (approved):
This is a nice improvement. Sorry it took forever to review and merge,

Issue #798, opened 2018-11-07
Win7: Cyrillic input with shift partially does not work


MVKozlov:
Environment data
----------------
```powershell
PS version: 5.1.14409.1012
PSReadline version: 2.0.0-beta3
os: 6.1.7601.17514 (win7sp1_rtm.101119-1850)
PS file version: 10.0.14409.1005 (rs1_srvoob.161208-1155)
```
Steps to reproduce or exception report
--------------------------------------
Installed latest PSReadline under PSv5.1 / Win7

switch keyboard to russian input, type:
- without shift - `йцукенгшщзхъ`
- should be (without shift but with caps) -  `ЙЦУКЕНГШЩЗХЪ`
- with shift only - `ЦУШЩХЪ`
- with caps + shift - `цукшщхъ`

and many other symbols got eaten too

MVKozlov:
Something breaks between beta1 and beta2

lzybkr:
Are you changing the layout after starting PowerShell? If so, this might be hard to fix, the api I rely on the determine if <kbd>Shift</kbd> is necessary does not honor keyboard layout changes.

MVKozlov:
Yes, I start PS and press <kbd>shortcut</kbd> to change layout. The same version (Beta3) works ok on Linux but does not work on Windows (both 7 and 10)

Moreover, if I change default layout to Russian, (I start PSs with RUS layout) I can't get `any` russian upper case symbols with <kbd>shift</kbd> (at least qwerty line) and even when I switch to English I can't type `any` English symbols with <kbd>shift</kbd> as well!

Seems there should be another `PlatformWindows.cs` override

ForNeVeR:
The impact of this issue is very high, because multi-layout users switch their layouts literally all the time. I have some file names and content I want to enter into my PowerShell session in my native layout, and so I need to change my layout from time to time without restarting the whole PowerShell session, that's a disaster.

@lzybkr, how could we help you with the issue? Could you please share any thoughts on that or show where I could find the code of that problematic API and PSReadLine's usage of that API? I'll be glad to help with the fix.

ForNeVeR:
Just wanted to report that it became worse on the latest nightly build at that point: no matter which layout is active when I start PowerShell 6.1 with PSReadLine, I cannot enter any Cyrillic letters at all: some of the keys result in `?` being added to the input and some of them don't work at all.

Will check the build from #822 after it's finished on CI though!

lzybkr:
[...]

PR #838, opened 2018-12-24
WIP - Test with real ConsoleKeyInfo values

lzybkr:
The biggest weakness in testing was the reuse of the internal values in the class `Keys` - it worked well for testing en-US on Windows because that's how I've done most of the development, but there have been too many problems to continue with that approach.

The new approach is to collect the actual `ConsoleKeyInfo` values to test against, once for each keyboard layout that we care to test. The script `Get-KeyInfoData.ps1` can be used, it's a bit tedious, takes maybe 5-15 minutes depending on how careful you are and how familiar you are with the keyboard layout, but it's a one-time task to add a new layout.

I've collected the data for en-US ensuring all tests pass, then added fr-FR and some tests fail. Also note that testing both fr-FR with en-US also causes some en-US tests to fail, that's why this is WIP.

If anyone cares to help out collecting the keys for other layouts (e.g. `pt-BR`, `nb-NO`, `es-ES`, `sv-SE`, `it-IT`, any Mac/Linux) - I'd appreciate the help.

Note that the `Get-KeyInfoData.ps1` script isn't totally user friendly, e.g. if it asks you to enter a key that the system consumes, the script never sees any input, so you're forced to enter something random, and you'll have to do it twice to get past the typo detection that can't work perfectly, but works well enough. I suggest using <kbd>F12</kbd> but it doesn't matter too much, after adding the tests I'll figure out which keys don't make sense.


MVKozlov:
A couple of questions:
when it say
```
Enter <w>:
Enter <V>:
```
<kbd>CAPS, V</kbd> or <kbd>Shift+v</kbd> ?
```
Enter <Ctrl+C>:
```
<kbd>CAPS, Ctrl+C</kbd> or <kbd>Ctrl+c</kbd> or <kbd>Shift+Ctrl+c</kbd> ?

there is
```
Enter Ctrl+c
Enter Ctrl+C
```
should I use the same combination or press <kbd>Shift</kbd> for second ? and many others that exists only with <kbd>shift</kbd>

<kbd>Enter</kbd> is a gray `Enter` ? on primary it say "double check', and for `<Ctrl+Alt+?>` too (may be because `?` is a <kbd>Shift+/</kbd> )

for `/`, `*`, `-`, `+` should I use <kbd>Shift</kbd> combinations or gray keyboard ?

btw, `[CmdletBindingAttribute()]` is a typo or something latest PSv6 specific ?

lzybkr:
@MVKozlov - thanks for giving this a go.

It looks like for regular keys, <kbd>Caps Lock</kbd> and <kbd>Shift</kbd> should work identically, so either is fine and it will be good to test without <kbd>Shift</kbd>.

For control keys, it looks <kdb>Caps Lock,Ctrl+a</kbd> does not turn into <kbd>Ctrl+A</kbd>, so avoid <kbd>Caps Lock</kbd> for modified keys.

If you get the message `double check`, don't worry too much about that. The intent is to catch typos, but that logic is dependent on possibly buggy logic in `PSReadLine`. If you enter the same key twice, it accepts it as correct and moves on.

For symbols, I suggest using the most natural keys. If you're not too sure, then I suggest not using the numeric keypad as laptops often don't have those keys.

And regarding `CmdletBindingAttribute`, that's always been supported, I didn't realize I typed out the full type name. By convention, attributes use the suffix `Attribute` and languages like C# or PowerShell can resolve the type name when used as an attribute with or without the suffix. PowerShell goes a bit further since V3 an resolves the name when used as a type (e.g. `'CmdletBinding' -as [Type]`), whereas in C# you'd have to write `typeof(CmdletBindingAttribute)`.

MVKozlov:
You do not say a word about duplicates - <kbd>Ctrl+C</kbd> and <kbd>Ctrl+c</kbd> - should I use <kbd>Shift+Ctrl+c</kbd> for first. I think I should

<kbd>Caps Lock</kbd> and <kbd>Shift</kbd> can work differently at least at PSRealine :), see #798 
I try to use CAPS.

So, I make PR when i finished with it.

and about attribute:
``` powershell
PS D:\> D:\Work\PSReadLine\test\Get-KeyInfoData.ps1
Cannot find the type for custom attribute 'CmdletBindingAttribute'. Make sure that the assembly that contains this type is loaded
.
At D:\Work\PSReadLine\test\Get-KeyInfoData.ps1:1 char:1
+ [CmdletBindingAttribute()]
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: ([CmdletBindingAttribute()]:AttributeAst) [], RuntimeException
    + FullyQualifiedErrorId : CustomAttributeTypeNotFound

PS D:\> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      5.1.14409.1012
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.14409.1012
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```
but with `[CmdletBinding()]` its OK

MVKozlov:
PR #839

lzybkr:
Thanks @MVKozlov. `[CmdetBindingAttribute()]` worked for me because of something I have in my profile. I fixed the script so nobody else will hit the same thing.

Can I ask how you entered the English characters? Alt+numpad? Switch layouts? Or is there another way?

After I work out how to best run the tests with different layouts, we'll start adding language specific tests, so for example we can make sure Cyrillic characters work properly, and make sure you can enter other characters like ã or à or â correctly.

There is a bug in a Win32 api that I'm using that doesn't correctly handle keyboard layout changes. It doesn't affect too many key bindings, but it definitely affects running the tests, so I need to figure out a way to run the tests where the keyboard layout is changed before launching the test runner and the test runner only tests a single language.

MVKozlov:
I switched to US layout but if need I can collect Cyrillic key info on the same keys with Cyrillic layout (but seems I will need to enter every key twice :)

MVKozlov:
@lzybkr, so there is need to repeat collecting console info with cyrillic layout switched on ?

kborowinski:
PR #870

PR #872, opened 2019-03-21
ensure text files use crlf line endings

SteveL-MSFT:
The signing tool appends the signature with crlf, but the files in git have only lf endings.  When copying the text files to the release location, replace lf with crlf endings.

### DIFF
--- a/PSReadLine.build.ps1
+++ b/PSReadLine.build.ps1
@@ -32,6 +32,10 @@ Write-Verbose "Building for '$target'" -Verbose
 
 $CLI_VERSION = "2.1.300"
 
+function ConvertTo-CRLF([string] $text) {
+    $text.Replace("`r`n","`n").Replace("`n","`r`n")
+}
+
 <#
 Synopsis: Ensure dotnet is installed
 #>
@@ -327,7 +331,7 @@ task LayoutModule BuildMainModule, BuildMamlHelp, {
     {
         # ensure files have \r\n line endings as the signing tool only uses those endings to avoid mixed endings
         $content = Get-Content -Path $file -Raw
-        Set-Content -Path (Join-Path $targetDir (Split-Path $file -Leaf)) -Value $content.Replace("`n","`r`n") -Force
+        Set-Content -Path (Join-Path $targetDir (Split-Path $file -Leaf)) -Value (ConvertTo-CRLF $content) -Force
     }
 
     $binPath = "PSReadLine/bin/$Configuration/$target/publish"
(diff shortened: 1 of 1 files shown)


PSReadLine.build.ps1
```diff
@@ -325,7 +325,9 @@ task LayoutModule BuildMainModule, BuildMamlHelp, {
 
     foreach ($file in $extraFiles)
     {
-        Copy-Item $file $targetDir
+        # ensure files have \r\n line endings as the signing tool only uses those endings to avoid mixed endings
```
lzybkr, line 332: The line endings depend on your git configuration. It is possible to check out with with either style on both Windows and Unix.

Because of this, I don't think it's safe to assume LF endings - you don't want to replace CRLF with CRLFLF.

This code might also inadvertently change a binary file.

One possible fix is to take a dependency on the `unix2dos` command, installed on Ubuntu with `sudo apt install dos2unix`. On Windows, we could also expect it's installed, possibly in wsl, so e.g. running `wsl unix2dos $file`.

PSReadLine.build.ps1
```diff
@@ -342,7 +344,7 @@ task LayoutModule BuildMainModule, BuildMamlHelp, {
 
     # Copy module manifest, but fix the version to match what we've specified in the binary module.
     $version = (Get-ChildItem -Path $targetDir/Microsoft.PowerShell.PSReadLine2.dll).VersionInfo.FileVersion
-    $moduleManifestContent = Get-Content -Path 'PSReadLine/PSReadLine.psd1' -Raw
+    $moduleManifestContent = (Get-Content -Path 'PSReadLine/PSReadLine.psd1' -Raw).Replace("`n","`r`n")
```
lzybkr: Since the replace happens in 2 places, you should add a function.

lzybkr (changes requested):

PSReadLine.build.ps1
SteveL-MSFT: This list is only the text files already so there should be no concern about changing a binary.  I can add a check to only do the change if the file is only using LF.

PSReadLine.build.ps1
SteveL-MSFT: Will change to a function

SteveL-MSFT:
@lzybkr I believe I've addressed your feedback, can you take another look?  Appveyor failure looks unrelated to this change.

SteveL-MSFT:
@lzybkr I believe this is ready to be merged

lzybkr (approved):

PSReadLine.build.ps1
```diff
@@ -32,6 +32,10 @@ Write-Verbose "Building for '$target'" -Verbose
 
 $CLI_VERSION = "2.1.300"
 
+function ConvertTo-CRLF([string] $text) {
```
rkeithhill, line 35: @SteveL-MSFT Sure seems like such a basic, cross-platform support utility should be built-into PowerShell.  :-)  Along with ConvertTo-LF.   Or maybe just `Convert-LineEnding` with `-CRLF` and `-LF` exclusive switches.

SteveL-MSFT: @rkeithhill We do have https://github.com/PowerShell/PowerShell/issues/2872

PR #894, opened 2019-04-15
Fix - Incorrect links to documentation files.

springcomp:
The `PSReadLine.csproj` project file includes various text and markdown documentation files, but these are actually available from the root solution folder. This PR fixes the links to those documentation files.

### DIFF
--- a/PSReadLine/PSReadLine.csproj
+++ b/PSReadLine/PSReadLine.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <OutputType>Library</OutputType>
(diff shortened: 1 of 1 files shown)


daxian-dbw (approved):

lzybkr (commented):
Did you change the newlines from Windows to Unix? Every line seems to be changed.

springcomp:
> Did you change the newlines from Windows to Unix? Every line seems to be changed.

I may have done that unexpectedly indeed. Please, let me fix this.

PSReadLine/PSReadLine.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
daxian-dbw: Maybe restore the original encoding too?

springcomp: Done.

daxian-dbw (approved):

PR #898, opened 2019-04-17
Fix 'Shift+Backspace' to work the same as 'Backspace'

### LINKED ISSUE #896: Shift+Backspace should work the same as just Backspace
With the post-beta.3 PSReadLine, `Shift+Backspace` produces `^H`. This is a regression from beta.3, and `Shift+Backspace` should work just like `Backspace`.

daxian-dbw:
Fix #896
Make `Shift+Backspace` and `Shift+Escape` work the same as `Backsapce` and `Escape` on windows.

### DIFF
--- a/test/KeyBindingsTest.cs
+++ b/test/KeyBindingsTest.cs
@@ -47,5 +47,21 @@ namespace Test
             TestSetup(KeyMode.Vi);
             Test("", Keys(Enumerable.Repeat(_.Ctrl_Alt_Question, 3)));
         }
+
+        [SkippableFact]
+        public void ShiftBackspace()
+        {
+            TestSetup(KeyMode.Cmd);
+            Test("aa", Keys("aaa", _.Shift_Backspace));
+        }
+
+        [SkippableFact]
+        public void ShiftEscape()
+        {
+            TestSetup(KeyMode.Cmd);
+
+            System.Threading.Thread.Sleep(10000);
+            Test("", Keys("aaa", _.Shift_Escape));
+        }
     }
 }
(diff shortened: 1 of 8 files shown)


lzybkr (commented):
Can you add a test?

test/KeyBindingsTest.cs
```diff
@@ -47,5 +47,21 @@ public void ShowKeyBindings()
...
+            TestSetup(KeyMode.Cmd);
+            Test("aa", Keys("aaa", _.Shift_Backspace));
+        }
+
+        [SkippableFact]
+        public void ShiftEscape()
+        {
+            TestSetup(KeyMode.Cmd);
+
+            System.Threading.Thread.Sleep(10000);
```
lzybkr: Remove or comment why this is necessary.

test/KeyInfo-en-US-windows.json
```diff
@@ -1301,6 +1301,18 @@
         "ConsoleKey":  "UpArrow",
         "Modifiers":  "Shift"
     },
+    {
```
lzybkr, line 1304: The other keyboard layouts will need similar entries.

lzybkr (changes requested):

test/KeyBindingsTest.cs
daxian-dbw: It was for diagnosis :) Already removed.

test/KeyInfo-en-US-windows.json
daxian-dbw: Updated. I installed the keyboard layouts and collect the data using `[Console]::ReadKey()`.

lzybkr (approved):

PR #899, opened 2019-04-18
Make sure generate the 'OnIdle' event when there are other subscribers

### LINKED ISSUE #827: PowerShell.OnIdle seems not to fire in PSReadline 2.0 beta (repro)
Environment data
----------------

PS version: 5.1.17134.228
os: 10.0.17134.1 (WinBuild.160101.0800)
PS file version: 10.0.17134.1 (WinBuild.160101.0800)

Further to #773, I think I can repro this pretty reliably now. It seems that the `OnIdle` event isn't fired when another event subscriber already exists.

```sh
PS> powershell -noprofile
PS> Register-EngineEvent -SourceIdentifier PowerShell.OnIdle -Action {Write-Host '*idle*'} -MaxTriggerCount 1
PS> *idle*
PS> $timer = New-Object System.Timers.Timer -Property @{ Interval = 1000; AutoReset = $false }
PS> Register-ObjectEvent $timer Elapsed -Action { Write-Host '*timer*' } -MaxTriggerCount 1
PS> Register-EngineEvent -SourceIdentifier PowerShell.OnIdle -Action {Write-Host '*idle*'} -MaxTriggerCount 1
PS> # nothing...
PS> Get-EventSubscriber

SubscriptionId   : 3
SourceObject     : System.Timers.Timer
EventName        : Elapsed
SourceIdentifier : 6ac2a48b-3498-47af-b32b-f82652dda9f5
Action           : System.Management.Automation.PSEventJob
HandlerDelegate  :
SupportEvent     : False
ForwardEvent     : False

SubscriptionId   : 4
SourceObject     :
EventName        :
SourceIdentifier : PowerShell.OnIdle
Action           : System.Management.Automation.PSEventJob
HandlerDelegate  :
SupportEvent     : False
ForwardEvent     : False

PS> Remove-Module PSReadline
PS> *idle*
```

daxian-dbw:
Fix #827 

Make sure we generate the 'OnIdle' event when there are other subscribers.

After the fix, 'OnIdle' event registered after other event subscribers works:

image

### DIFF
--- a/PSReadLine/ReadLine.cs
+++ b/PSReadLine/ReadLine.cs
@@ -206,7 +206,7 @@ namespace Microsoft.PowerShell
                         bool runPipelineForEventProcessing = false;
                         foreach (var sub in eventSubscribers)
                         {
-                            if (sub.SourceIdentifier.Equals("PowerShell.OnIdle", StringComparison.OrdinalIgnoreCase))
+                            if (sub.SourceIdentifier.Equals(PSEngineEvent.OnIdle, StringComparison.OrdinalIgnoreCase))
                             {
                                 // There is an OnIdle event.  We're idle because we timed out.  Normally
                                 // PowerShell generates this event, but PowerShell assumes the engine is not
@@ -216,15 +216,13 @@ namespace Microsoft.PowerShell
                                 runPipelineForEventProcessing = true;
                                 break;
                             }
-
-                            // If there are any event subscribers that have an action (which might
-                            // write to the console) and have a source object (i.e. aren't engine
-                            // events), run a tiny useless bit of PowerShell so that the events
-                            // can be processed.
-                            if (sub.Action != null && sub.SourceObject != null)
+                            else if (sub.Action != null && sub.SourceObject != null)
                             {
+                                // If there are any event subscribers that have an action (which might
+                                // write to the console) and have a source object (i.e. aren't engine
+                                // events), run a tiny useless bit of PowerShell so that the events
+                                // can be processed.
                                 runPipelineForEventProcessing = true;
-                                break;
                             }
                         }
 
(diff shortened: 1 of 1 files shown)


PSReadLine/ReadLine.cs
```diff
@@ -206,7 +206,7 @@ internal static PSKeyInfo ReadKey()
                         bool runPipelineForEventProcessing = false;
                         foreach (var sub in eventSubscribers)
                         {
-                            if (sub.SourceIdentifier.Equals("PowerShell.OnIdle", StringComparison.OrdinalIgnoreCase))
+                            if (sub.SourceIdentifier.Equals(PSEngineEvent.OnIdle, StringComparison.OrdinalIgnoreCase))
```
lzybkr, line 210: The same change applies to line 215.

PSReadLine/ReadLine.cs
```diff
@@ -216,15 +216,13 @@ internal static PSKeyInfo ReadKey()
                                 runPipelineForEventProcessing = true;
                                 break;
                             }
-
-                            // If there are any event subscribers that have an action (which might
-                            // write to the console) and have a source object (i.e. aren't engine
-                            // events), run a tiny useless bit of PowerShell so that the events
-                            // can be processed.
-                            if (sub.Action != null && sub.SourceObject != null)
+                            else if (sub.Action != null && sub.SourceObject != null)
```
lzybkr: Else is not necessary, right?

lzybkr (approved):
If I'm reading `git blame` correctly, this scenario never really worked - I'm surprised it took so long for someone to notice.

It took me a minute to understand *why* the break was causing a problem, so maybe you could add a comment to make it clear.

Alternatively (I'm not sure it would be easy) - you could add a test.

PSReadLine/ReadLine.cs
daxian-dbw: Yeah, we can remove it.

PSReadLine/ReadLine.cs
daxian-dbw: Cannot believe I didn't see that ... Fixed.

Issue #930, opened 2019-05-28
Tests fail because of missing TestResults folder


Snaptags:
Build fails unless you add `.\PSReadLine\test\TestResults` manually:
_PathNotFound: .\PSReadLine\tools\helper.psm1:285 char:9_

After adding the folder the tests seem to run without error, but the TestResults-folder remains empty…

Environment data
----------------
PS version: 5.1.17763.503
PSReadline version: 2.0.0-beta4
os: 10.0.17763.526 (WinBuild.160101.0800)
PS file version: 10.0.17763.526 (WinBuild.160101.0800)

Steps to reproduce or exception report
--------------------------------------
- if fr-FR and/or en-US keyboard layouts are installed: uninstall them.
- `invoke-build` on a freshly cloned copy (master or ci branch)

Snaptags:
The tests can only be run if either the en-US or the fr-FR keyboard layout has been installed on the host. If you try running the build using any other layout the test runner simply skips **all** tests, resulting in a missing report and an empty TestResults folder…

daxian-dbw:
Fixed via #1071

PR #931, opened 2019-05-28
create TestResults folder if missing #930

Snaptags:
This addresses #930 by creating the folder if missing. Somebody should double check if the missing folder isn't a side-effect of something else going wrong.


Snaptags:
This does not fix the problem, it only hides the error

PR #933, opened 2019-06-03
Fix the conversion from ConsoleKeyInfo/PSKeyInfo to PSKeyInfo/ConsoleKeyInfo

### LINKED ISSUE #914: Beta 4: problems with dead keys and custom keyboard layout (Neo2)
I use a custom keyboard layout called Neo2 offering multiple layers using modifier keys, e.g. `<CAPS-LOCK>+<k>` results in `)`
Since installing beta4 of PSReadline, every time I press `<CAPS-LOCK>` a @ character is added to my input, So instead of
`git commit -am "message"`
 I end up "typing"
`git commit @-am @"message@"`

Environment data
----------------
PS version: 5.1.17763.316 / same for 6.2
os: 10.0.17763.470 (WinBuild.160101.0800)
PS file version: 10.0.17763.470 (WinBuild.160101.0800)
vi **or** standard edit mode.

Install the Neo2 driver: https://github.com/neo-layout/neo-layout/blob/master/windows/Neo2.0_setup.exe
And add the "German (Neo 2.0)" keyboard layout to your input settings, open powershell and hit \<caps-lock\>

daxian-dbw:
Fix #914

Fix the conversion from `ConsoleKeyInfo`/`PSKeyInfo` to `PSKeyInfo`/`ConsoleKeyInfo`, so that the dead keys of a particular keyboard layout can be properly ignored.

I have verified the fix with US International keyboard layout and French keyboard layout.

### DIFF
--- a/PSReadLine/Keys.cs
+++ b/PSReadLine/Keys.cs
@@ -268,8 +268,15 @@ namespace Microsoft.PowerShell
                     break;
 
                 case '\0':
-                    // This could be a special kind of a modifier key (dead key) on a particular keyboard layout.
-                    s = key.Key.ToString();
+                    // On Windows:
+                    //   This could be a special kind of a modifier key (dead key) for a particular keyboard layout.
+                    //   We use the text form of the virtual key in such case, so the resulted PSKeyInfo can be converted back to ConsoleKeyInfo correctly later on,
+                    //   and be properly ignore during rendering.
+                    // On non-Windows:
+                    //   The dead key is not an issue when there is a tty involved.
+                    //   But the virtual key is not captured as accurately as on Windows, e.g. Ctrl+2 results in `key.Key = 0`.
+                    //   So on non-Windows, we use `@` in case `key.KeyChar = '\0'`. This is ugly but familiar.
+                    s = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? key.Key.ToString() : "@";
                     break;
 
                 case char _ when (c >= 1 && c <= 26):
--- a/PSReadLine/Keys.cs
+++ b/PSReadLine/Keys.cs
@@ -268,8 +268,8 @@ namespace Microsoft.PowerShell
                     break;
 
                 case '\0':
-                    // This is ugly but familiar.
-                    s = "@";
+                    // This could be a special kind of a modifier key (dead key) on a particular keyboard layout.
+                    s = key.Key.ToString();
                     break;
 
                 case char _ when (c >= 1 && c <= 26):
--- a/PSReadLine/Keys.cs
+++ b/PSReadLine/Keys.cs
@@ -125,7 +125,7 @@ namespace Microsoft.PowerShell
 
         static readonly ThreadLocal<char[]> toUnicodeBuffer = new ThreadLocal<char[]>(() => new char[2]);
         static readonly ThreadLocal<byte[]> toUnicodeStateBuffer = new ThreadLocal<byte[]>(() => new byte[256]);
-        internal static void TryGetCharFromConsoleKey(ConsoleKeyInfo key, ref char result)
+        internal static void TryGetCharFromConsoleKey(ConsoleKeyInfo key, ref char result, ref bool isDeadKey)
         {
             var modifiers = key.Modifiers;
             var virtualKey = key.Key;
@@ -149,10 +149,19 @@ namespace Microsoft.PowerShell
             }
             int charCount = ToUnicode(virtualKey, scanCode, state, chars, chars.Length, flags);
 
-            // TODO: support diacriticals (charCount == 2)
             if (charCount == 1)
             {
                 result = chars[0];
+            }
+            else if (charCount == -1 || charCount >=2)
+            {
+                // Quoted from https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-tounicode#return-value:
+                //  "Return value  -1 --
+                //     The specified virtual key is a dead-key character (accent or diacritic).
+                //   Return value >=2 --
+                //     Two or more characters were written to the buffer specified by pwszBuff. The most common cause for this is that a dead-key character 
+                //     (accent or diacritic) stored in the keyboard layout could not be combined with the specified virtual key to form a single character."
+                isDeadKey = true;
             }
         }
 
@@ -217,6 +226,7 @@ namespace Microsoft.PowerShell
             }
 
             var c = key.KeyChar;
+            var isDeadKey = false;
             if (char.IsControl(c) )
             {
                 // We have the virtual key code and Windows has a handy api to map that to the non-control
@@ -226,7 +236,7 @@ namespace Microsoft.PowerShell
                 if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                 {
                     var keySansControl = new ConsoleKeyInfo(key.KeyChar, key.Key, isShift, isAlt, control: false);
-                    TryGetCharFromConsoleKey(keySansControl, ref c);
+                    TryGetCharFromConsoleKey(keySansControl, ref c, ref isDeadKey);
                 }
             }
             else if (isAlt && isCtrl)
@@ -268,15 +278,13 @@ namespace Microsoft.PowerShell
                     break;
 
                 case '\0':
-                    // On Windows:
-                    //   This could be a special kind of a modifier key (dead key) for a particular keyboard layout.
-                    //   We use the text form of the virtual key in such case, so the resulted PSKeyInfo can be converted back to ConsoleKeyInfo correctly later on,
-                    //   and be properly ignore during rendering.
-                    // On non-Windows:
-                    //   The dead key is not an issue when there is a tty involved.
-                    //   But the virtual key is not captured as accurately as on Windows, e.g. Ctrl+2 results in `key.Key = 0`.
-                    //   So on non-Windows, we use `@` in case `key.KeyChar = '\0'`. This is ugly but familiar.
-                    s = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? key.Key.ToString() : "@";
+                    // This could be a dead key for a particular keyboard layout in Windows console.
+                    // The dead key is not an issue when there is tty involved, so on non-Windows, `isDeadKey` is always false.
+                    //
+                    // When we believe it's a dead key, we use the text form of the virtual key so the resulted PSKeyInfo can be
+                    // converted back to ConsoleKeyInfo correctly later on, and be properly ignore during rendering.
+                    // Otherwise, we use `@` in case `key.KeyChar = '\0'`. This is ugly but familiar.
+                    s = isDeadKey ? key.Key.ToString() : "@";
                     break;
 
                 case char _ when (c >= 1 && c <= 26):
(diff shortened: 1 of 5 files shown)


PSReadLine/Keys.cs
```diff
@@ -268,8 +268,8 @@ void AppendPart(string str)
                     break;
 
                 case '\0':
-                    // This is ugly but familiar.
-                    s = "@";
+                    // This could be a special kind of a modifier key (dead key) on a particular keyboard layout.
+                    s = key.Key.ToString();
```
lzybkr: The point in displaying `@` was to be consistent with bash on Linux and to be consistent on Windows when TryGetCharFromConsoleKey failed.

I'm guessing if TryGetCharFromConsoleKey was passed `control: true` that the dead key would have been resolved correctly. If that's the case, maybe the call above on line 229 should retry?

If that doesn't work, can we be smarter here? At least with an English keyboard on Linux, this will typically result in `0`, and on Windows where the `Key` is typically correct, we might see weird things like `D2` or maybe `NumPad2`.

lzybkr:
This fix should also include a test.

PSReadLine/Keys.cs
daxian-dbw: > The point in displaying `@` was to be consistent with bash on Linux

Can you please give me an example of bash displaying `@` on Linux?

> If that's the case, maybe the call above on line 229 should retry?

I will check if `TryGetCharFromConsoleKey` can resolve the dead key as expected with `control: true`.

> At least with an English keyboard on Linux, this will typically result in `0`, and on Windows where the `Key` is typically correct, we might see weird things like `D2` or maybe `NumPad2`.

I'm a little confused here ... what do you mean by **this will typically result in 0**? Do you mean the `key.Key` could be `0` with English keyboard on Linux?
And by "**we might see weird things like `D2` or maybe `NumPad2`**", why is that wired? and could you please give me an example? Thanks!

daxian-dbw:
Will add a test, just don't know how to reliably test this yet :)

PSReadLine/Keys.cs
daxian-dbw: > I'm guessing if TryGetCharFromConsoleKey was passed `control: true` ....

You mean when calling `ToUnicode`, set `lpKeyState` as if the `CTRL` key is pressed? This feels weird to me ... and what about on non-Windows? On non-Windows, the code flow will still reach `case '\0'` and `@` will be used ...

lzybkr: <kbd>Ctrl+2</kbd> is `0`:

```
jason@jasonsh02:/mnt/c/WINDOWS/system32$ showkey -a

Press any keys - Ctrl-D will terminate this program

^@        0 0000 0x00
```

If you try the same in PowerShell via Console.ReadKey, you'll see that Key is 0.

On Windows, where the Key is captured correctly, <kbd>2</kbd> is `ConsoleKey.D2`.

lzybkr: I'm guessing when there is a tty involved, dead keys are not an issue, the CLR and hence PSReadLine won't see control or alt for simple characters.

On Windows, Console.ReadKey is in some ways not helpful reporting keyboard state, although it does allow for some key bindings that aren't supported on Linux.

daxian-dbw: I think your main concern is that with the change in this PR, the same key chord will result in different `PSKeyInfo` instances, which breaks the whole point of abstraction using `PSKeyInfo` (e.g. key bindings).
Let me think about it more.

~At the meantime, can you please comment on my above question regarding "passing `control:true` to TryGetCharFromConsoleKey"?~ Just saw your reply.
I will spin up a Ubuntu 18.04 VM with different keyboard layouts to check for how dead keys are handled.

daxian-dbw: Updates:
1. `TryGetCharFromConsoleKey` with `control: true` doesn't resolve dead key. For example, with "US international", `'` is a dead key, and `ToUnicode` returns `0` even with `control: true`, meaning the virtual key `Oem7` has no translation for the "US international" keyboard layout.
2. The dead key is not a problem on Linux -- `Console.ReadKey()` doesn't return when only a dead key is pressed. Like you said above, it's likely dead keys are not an issue when there is a tty involved. So this is a windows only problem

In my opinion, the fix should just be not rendering a dead key (a dead key always has `KeyChar` to be `'\0'`).
I'm having trouble understanding the point of using `'@'` here when `KeyChar` is `'\0'`. Although `showkey -a` prints `^@` for `CTRL+2`, bash doesn't print anything for this key chord, so why does PSReadLine has to print `@` for it? You also mentioned _consistency on Windows when `TryGetCharFromConsoleKey` failed_, again why do we have to print anything when `KeyChar` is `'\0'` and `TryGetCharFromConsoleKey` fails? and why does that consistency matter?

lzybkr: I might be misremembering, but I thought I've seen some Linux terminal displaying `^@` when typing <kbd>Ctrl+2</kbd>, and I emulated it even though it felt weird.

But you are correct - dead keys should not be rendered, so the trick is in distinguishing them, and I don't know without some investigation how you'd do that.

daxian-dbw: Before I go further to find out a way to determine a key chord is a dead key, I would like to understand the point of using `'@'` when `KeyChar` is `'\0'` here.

On Ubuntu 16.04 and 18.04 (maybe other Ubuntu versions too but didn't try), bash doesn't print out anything for <kbd>Ctrl+2</kbd>. Maybe some Linux terminal does display `^@`, but why do we want to emulate that instead of the behavior of Ubuntu bash?
You also mentioned _consistency on windows_, but again, why do we have to print out anything when `KeyChar` is `'\0'` and `TryGetCharFromConsoleKey` fails? Just not rendering the key sounds like a reasonable behavior to me.

So in my opinion, the current change to use `key.Key.ToString()` feels natural to me -- ignore `'\0'` KeyChar when rendering and can be converted back the the `ConsoleKeyInfo` relatively reliably.

daxian-dbw: Also want to add that, in beta.4, <kbd>Ctrl+2</kbd>, <kbd>Ctrl+4</kbd>, <kbd>Ctrl+5</kbd> and etc doesn't result in `^@`, `^\` and `^]` as shown in `showkey -a`, but result in `@`, `\`, `]` because the modifier state is `0`.

lzybkr: Consistency is probably impossible - `wsl` doesn't show anything for <kbd>Ctrl+4</kbd> or <kbd>Ctrl+5</kbd>, it's as though you didn't press a key.

Not showing anything is probably fine, but there is an bash key binding [`quoted-insert`](https://www.gnu.org/software/bash/manual/html_node/Commands-For-Text.html#Commands-For-Text) (unimplemented in PSReadLine) that should make it possible to insert control characters like <kbd>Ctrl+2</kbd>.

At any rate - I'm assuming that `key.Key.ToString()` returns `0` on Linux and that's probably not desirable.

daxian-dbw: The `quoted-insert` key binding is interesting. I see <kbd>Ctrl+3</kbd>, <kbd>Ctrl+4</kbd>, <kbd>Ctrl+5</kbd>, and <kbd>Ctrl+6</kbd> print `^[`, `^\`, `^]`, and `^^` as shown in `showkey -a`, but <kbd>Ctrl+2</kbd> still prints nothing. Quite stubborn, right 😄

> At any rate - I'm assuming that key.Key.ToString() returns 0 on Linux and that's probably not desirable.

Agreed. This results in printing `0` for <kbd>Ctrl+2</kbd> on Linux, which is less desired than `@`.
Since the dead key issue is windows-only, I guess we should keep using `@` in case of `KeyChar == '\0'` on non-windows. I will update the code change.

PSReadLine/Keys.cs
```diff
@@ -149,11 +149,20 @@ internal static void TryGetCharFromConsoleKey(ConsoleKeyInfo key, ref char resul
             }
             int charCount = ToUnicode(virtualKey, scanCode, state, chars, chars.Length, flags);
 
-            // TODO: support diacriticals (charCount == 2)
             if (charCount == 1)
             {
                 result = chars[0];
             }
+            else if (charCount == -1 || charCount >=2)
```
lzybkr, line 156: You'll want to test starting PowerShell (fresh new window so new conhost) with different layouts, e.g.

1. layout in en-us
2. start powershell
3. switch layout

The problem is some apis like `ToUnicode` (I think) don't honor the current keyboard layout - only the layout that was active when starting conhost.

daxian-dbw: Just tried this on my french locale machine (came with fr-FR keyboard by default, and installed English (US) keyboard layout). It works fine.
Will upload a .gif when I get a chance.

daxian-dbw: Here is the screen record. It looks `ToUnicode` does honor the current keyboard layout.
<kbd>AltGr+7</kbd> is the dead key for backtick. <kbd>AltGr+7</kbd> + <kbd>Space</kbd> will print the backtick character.
<kbd>AltGr+2</kbd> is the dead key for tilde. <kbd>AltGr+2</kbd> + <kbd>Space</kbd>will print the tilde character.
<kbd>[</kbd> is the dead key for caret. <kbd>[</kbd> + <kbd>Space</kbd> will print the caret character.

a

daxian-dbw: The tests passed on the French machine:

image

test/DeadKeyTest.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        [SkippableFact]
+        public void DeadKeyShouldBeIgnored()
+        {
+            Skip.If(this.Fixture.Lang != "fr-FR", "The dead key test requires Keyboard layout to be set to 'fr-FR'");
+            TestSetup(KeyMode.Cmd);
+
+            Test("aa", Keys("aa", _.DeadKey_Caret));
+            Test("aab", Keys("aa", _.DeadKey_Caret, 'b'));
+        }
```
lzybkr, line 17: So this is just testing that the deadkey is ignored?

Can you also test a key that requires the deadkey?

daxian-dbw: I would like to but not sure how to do it. With the French keyboard layout, when I press <kbd>[</kbd> then <kbd>q</kbd>, <kbd>[</kbd> will cause `Console.ReadKey` to return with a dead key, and then <kbd>q</kbd> will result in `<KeyChar = 'â', Key = A, Modifiers = 0>` returned by `Console.ReadKey`. In our tests, the mock console directly returns mocked `ConsoleKeyInfo` and doesn't cause any key actually be pressed, so there is no way to get the key `â` by mocking <kbd>[</kbd> + <kbd>q</kbd>.

I think we would need another mock console that can send keys to the `dotnet exec` process when `ReadKey` is called in order to test a key that requires the deadkey. Are you suggesting that or you have a different idea?

lzybkr: I guess I was just thinking of changing your test to use something more like `<KeyChar = 'â', Key = A, Modifiers = 0>` as the second key. In other words, capture what `Console.ReadKey` actually sends us for our desired key sequences and make sure that gets rendered properly - don't try to emulate the actual key presses.

daxian-dbw: OK, that sounds reasonable. I will make the change.

daxian-dbw: Done, and tests passed with the French keyboard layout.

image

lzybkr (approved):

PR #935, opened 2019-06-07
Make `Ctrl+[` bound-able on Windows

### LINKED ISSUE #906: "Ctrl+[" is not bondable as a key chord in PSReadLine
Environment data
----------------
**AppVeyor Build: 1.0.553**
```
PS version: 5.1.17763.316
PSReadline version: 2.0.0-beta4, **AppVeyor Build: 1.0.553**
os: 10.0.17763.1 (WinBuild.160101.0800)
PS file version: 10.0.17763.1 (WinBuild.160101.0800)
```
and
```
PS version: 6.2.0
PSReadline version: 2.0.0-beta4, **AppVeyor Build: 1.0.553**
os: 10.0.17763.1 (WinBuild.160101.0800)
PS file version: 6.2.0
```

Enable PSReadLine vi-mode
While in insert-mode press `Ctrl+[` and prior behavior was to change to command-mode
Currently, `^[` is rendered in the console and PSReadLine remains in insert-mode.

daxian-dbw:
Fix #906
When converting `PSKeyInfo` to `ConsoleKeyInfo`, use `VkKeyScan` to get the virtual key more accurately.

### DIFF
--- a/PSReadLine/ConsoleKeyChordConverter.cs
+++ b/PSReadLine/ConsoleKeyChordConverter.cs
@@ -221,6 +221,9 @@ namespace Microsoft.PowerShell
                 { "UpArrow",    new KeyPair { Key = ConsoleKey.UpArrow} },
             };
 
+        [DllImport("user32.dll", CharSet = CharSet.Unicode)]
+        static extern short VkKeyScan(char ch);
+
         static bool MapKeyChar(string input, ref ConsoleModifiers mods, out ConsoleKey key, out char keyChar)
         {
             var isCtrl = (mods & ConsoleModifiers.Control) != 0;
@@ -243,7 +246,20 @@ namespace Microsoft.PowerShell
                 else
                 {
                     key = 0;
+                    if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                    {
+                        short vkKeyScan = VkKeyScan(keyChar);
+                        if (vkKeyScan != -1)
+                        {
+                            int vkCode = vkKeyScan & 0xff;
+                            if (typeof(ConsoleKey).IsEnumDefined(vkCode))
+                            {
+                                key = (ConsoleKey)vkCode;
+                            }
+                        }
+                    }
                 }
+
                 if (isCtrl)
                 {
                     switch (keyChar)
--- a/test/KeyBindingsTest.cs
+++ b/test/KeyBindingsTest.cs
@@ -61,5 +61,12 @@ namespace Test
             TestSetup(KeyMode.Cmd);
             Test("", Keys("aaa", _.Shift_Escape));
         }
+
+        [SkippableFact]
+        public void ControlLBracket()
+        {
+            TestSetup(KeyMode.Cmd, new KeyHandler("Ctrl+[", PSConsoleReadLine.BackwardDeleteChar));
+            Test("aa", Keys("aaa", _.Ctrl_LBracket));
+        }
     }
 }
--- a/test/KeyBindingsTest.cs
+++ b/test/KeyBindingsTest.cs
@@ -65,6 +65,7 @@ namespace Test
         [SkippableFact]
         public void ControlLBracket()
         {
+            // 'Ctrl+[' should work for key binding.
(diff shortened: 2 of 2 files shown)


PSReadLine/ConsoleKeyChordConverter.cs
```diff
@@ -243,7 +246,20 @@ static bool MapKeyChar(string input, ref ConsoleModifiers mods, out ConsoleKey k
                 else
                 {
                     key = 0;
+                    if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
lzybkr, line 249: Is this a Windows only problem? Not using `VkKeyScan` at initialization time (e.g. registering a binding) was important because it doesn't handle keyboard layout changes at all.

daxian-dbw: Hmm, it's a windows only problem for this particular binding: <kbd>Ctrl+[</kbd> --> `CommandMode in VI`, but <kbd>Ctrl+[</kbd> cannot be bound in general for other actions on Linux.

This is because on Linux, <kbd>Ctrl+[</kbd> produces `CosnoleKeyInfo  { KeyChar = \u0016, Key = Escape, Modifiers = 0 }`, which maps to `PSKeyInfo { Escape }`. `PSKeyInfo { Escape }` is already bound with `CommandMode` in VI. However, when running `Set-PSReadLineKeyHandler -Chord 'Ctrl+[' -Function ...`, the chord string `'Ctrl+['` results in `CosnoleKeyInfo { KeyChar = \u0016, Key = 0, Modifiers = Control }`, which maps to `PSKeyInfo { Control+Escape }`, so it will never be hit when pressing <kbd>Ctrl+[</kbd>.

Similarly on Windows, the chord string `'Ctrl+['` generates `PSKeyInfo { Control+Escape }`, but <kbd>Ctrl+[</kbd> produces `ConsoleKeyInfo { KeyChar = \u0016, Key = oem4, Modifiers = Control }`, which results in `PSKeyInfo { Control+[ }`, so the binding will never be hit.

Ideally we should fix it on all platforms, but I'm not sure what the right behavior is on non-Windows. Today, since <kbd>Ctrl+[</kbd> is translated to `PSKeyInfo { Escape }` on Linux (maybe macOS as well), it always has the same behavior as <kbd>Esc</kbd>. If we change how it's mapped to `PSKeyInfo`, this current default behavior will be broken, which might be considered a breaking change ... so I chose to keep the current behavior on non-Windows.

daxian-dbw: > Not using VkKeyScan at initialization time (e.g. registering a binding) was important because it doesn't handle keyboard layout changes at all.

Does it really matter in that scenario? If the user choose to register a binding in one keyboard layout and then switch to another keyboard layout, then I won't expect the registered key binding to work, because after switching to another keyboard layout, the same key chord is likely to be mapped to a different `PSKeyInfo` anyways ...

lzybkr: If I bind <kbd>q</kbd> in `en-US`, should that binding not work when changing to `fr-FR`? The same applies to any key like <kbd>Ctrl+[</kbd> - we shouldn't be using the scan code at binding time.

Now the fact that `VkKeyScan` doesn't work on layout changes means that it doesn't really matter, but that's a bug that could get fixed someday.

daxian-dbw: I will play with the "registering binding in one layout and use in another layout" a bit more to better understand what might be the right thing to do.
But for <kbd>Ctrl+[</kbd> no being able to bind to custom actions in general on Linux, what's your thought about it? Shall we keep the current behavior (<kbd>Ctrl+[</kbd> basically equivalent to <kbd>Esc</kbd>)?

lzybkr: My general thoughts in this area are:
* Don't try to emulate things that don't make sense - that code doesn't need to be written.
* If something can't be supported easily, *consider* adding a warning/error.
* It's OK to silently do something a little weird for corner cases.

PR #939, opened 2019-06-12
Loose the check for dead keys

daxian-dbw:
In #933, I tried to use the win32 API `ToUnicode` to check for dead keys -- according to the documentation, if the return value is `-1` or `>=2`, the key is most likely a dead key.
However, it turns out `ToUnicode` is not reliable when the keyboard is not natively supported by Windows. For example, with the Neo keyboard, `ToUnicode` seems to return 0 for dead keys. So the fix could be to loose the check for dead keys -- treat a key as a dead key as long as `ToUnicode` failed to translate.

I haven't set up the Neo keyboard yet ... so I haven't tried this fix myself. Will update once I got the Neo keyboard correctly added.

/cc @Snaptags and @MovGP0, can you please try out this fix on your Neo keyboard?

### DIFF
--- a/PSReadLine/Keys.cs
+++ b/PSReadLine/Keys.cs
@@ -155,6 +155,8 @@ namespace Microsoft.PowerShell
             //   Return value >=2 --
             //     Two or more characters were written to the buffer specified by pwszBuff. The most common cause for this is that a dead-key character 
             //     (accent or diacritic) stored in the keyboard layout could not be combined with the specified virtual key to form a single character."
+            //   Return value   0 --
+            //     The specified virtual key has no translation for the current state of the keyboard. Nothing was written to the buffer specified by pwszBuff.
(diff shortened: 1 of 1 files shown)


Snaptags:
That does look better, but only as far as the modifier keys are concerned (Caps-Lock, right Alt), they don't enter the "@" any longer.

But the problem of duplicated "dead keys" remains, i.e.
```à` â^``` instead of just ```a` a^```

This is where the dead keys are located using the Neo2 layout:
dead_keys

They are used to modify the NEXT character inserted, that is I enter \<^\> and then \<a\> to get ```â```. Or \<a\>, \<^\> and then \<space\> to get ```a^``` PSReadline applies the dead key to already entered characters, too.

daxian-dbw:
Thanks @Snaptags for the feedback. I will close this PR in favor of #945

PR #945, opened 2019-06-15
Avoid rendering dead keys (2nd attempt)

### LINKED ISSUE #914: Beta 4: problems with dead keys and custom keyboard layout (Neo2)

daxian-dbw:
Use a heuristic check to determine if a key pressing is a dead key. Fix #914 

The Win32 API `ToUnicode` doesn't work well with non-natively-supported keyboard layouts, such as the Neo keyboard layout. `ToUnicode` maps Neo dead keys to unicode characters incorrectly and results in PSReadLine rendering the dead keys.


daxian-dbw:
@Snaptags this fix seems to be working fine with Neo. Can you please give it another try?

Snaptags:
Dead key handling is working perfectly! Thank you for this great fix.

daxian-dbw:
Thanks @Snaptags for trying it out.
@lzybkr can you please review this fix? Thanks!

PSReadLine/Keys.cs
```diff
@@ -221,22 +222,36 @@ void AppendPart(string str)
...
                 // If this fails, we rely on some hard coded control characters below.
                 if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                 {
-                    var keySansControl = new ConsoleKeyInfo(key.KeyChar, key.Key, isShift, isAlt, control: false);
-                    TryGetCharFromConsoleKey(keySansControl, ref c, ref isDeadKey);
+                    // A heuristic to check for dead keys --
+                    //   We got an 'OemXXX' ConsoleKey, '\0' key char, and no 'Ctrl' modifier. It's very likely generated by a dead key.
+                    //   We check for 'Ctrl' modifier because it's easy to generate '\0' KeyChar and 'OemXXX' by combinding 'Ctrl' with
+                    //   another special key, such as 'Ctrl+?' and 'Ctrl+;'.
+                    isDeadKey = (c == '\0') && (consoleKey >= ConsoleKey.Oem1 && consoleKey <= ConsoleKey.Oem102) && !isCtrl;
```
lzybkr, line 243: There is a large gap of invalid values in that range - maybe you should use a switch?
You also miss `OemClear` in that range.

lzybkr (approved):

PSReadLine/Keys.cs
daxian-dbw: I only see the OemXXX `ConsoleKey`s generated for the dead keys that I have inspected so far ... Yes, `OemClear` is not in that range. I didn't include it only because I didn't see it was generated by the dead keys I have inspected :) I will keep the check as is for now, and change it as needed when there are new dead-key issue reported.

daxian-dbw:
Test run passed with French keyboard layout:

image

Issue #932, opened 2019-06-03
Error received in PowerShell


milescw:
Environment data
----------------
PS version: 5.1.17763.503
PSReadline version: 2.0.0-beta2
os: 10.0.17763.1 (WinBuild.160101.0800)
PS file version: 10.0.17763.1 (WinBuild.160101.0800)

Steps to reproduce or exception report
--------------------------------------
All I did was start to type docker run -v and as soon as I pressed the letter "d" this error came up.
$d
Oops, something went wrong.  Please report this bug with the details below.
Report on GitHub: https://github.com/lzybkr/PSReadLine/issues/new
-----------------------------------------------------------------------
Last 152 Keys:
 c d Space C : \ U s e r s \ C h a r l e s \ D o c u m e n t s \ T r a i n i n g \ U d a c i t y \ N a n o d e g r e e \ B l o c k c h a i n - D e v e l o p e r \ T e r m 2 \ P r o j e c t s \ u d b c d - b l o c k c h a i n - c a p s t o n e Enter
 p w d Enter
 f u n c t i o n Space p r o m p t Space { " $ " } Enter
 p w d Enter
 d o

Exception:
System.ArgumentOutOfRangeException: The value must be greater than or equal to zero and less than the console's buffer size in that dimension.
Parameter name: left
Actual value was -1.
   at System.Console.SetCursorPosition(Int32 left, Int32 top)
   at Microsoft.PowerShell.Internal.VirtualTerminal.set_CursorLeft(Int32 value)
   at Microsoft.PowerShell.PSConsoleReadLine.ReallyRender(RenderData renderData, String defaultColor)
   at Microsoft.PowerShell.PSConsoleReadLine.ForceRender()
   at Microsoft.PowerShell.PSConsoleReadLine.Insert(Char c)
   at Microsoft.PowerShell.PSConsoleReadLine.SelfInsert(Nullable`1 key, Object arg)
   at Microsoft.PowerShell.PSConsoleReadLine.ProcessOneKey(ConsoleKeyInfo key, Dictionary`2 dispatchTable, Boolean ignoreIfNoAction, Object arg)
   at Microsoft.PowerShell.PSConsoleReadLine.InputLoop()
   at Microsoft.PowerShell.PSConsoleReadLine.ReadLine(Runspace runspace, EngineIntrinsics engineIntrinsics)
-----------------------------------------------------------------------
$cker run

daxian-dbw:
[...]

PR #967, opened 2019-07-16
Fix one variant of the `ArgumentOutOfRangeException` issues

### LINKED ISSUE #919: System.ArgumentOutOfRangeException throw when press quote key
There are some possible duplicates with same exception, but no one has the same reproduce steps with mine.

Environment data
----------------

PS version: 5.1.17763.316
os: 10.0.17763.1 (WinBuild.160101.0800)
PS file version: 10.0.17763.1 (WinBuild.160101.0800)

1. `cd` into a deep, long-named folder, e.g, `~\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Startup`
2. resize the console so that the prompt occupies exactly one whole line (the cursor is at the second line), as the screenshot shows.
    image
3. press <kbd>'</kbd>

Exception throw:

````
PS C:\Users\Simon> cd "C:\Users\Simon\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Startup"
PS C:\Users\Simon\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Startup> 
Oops, something went wrong.  Please report this bug with the details below.
Report on GitHub: https://github.com/lzybkr/PSReadLine/issues/new
Last 83 Keys:
 c d Space " C : \ U s e r s \ S i m o n \ A p p D a t a \ R o a m i n g \ M i c r o s o f t \ W i n d o w s \ S t a r t Space M e n u \ P r o g r a m s \ S t a r t u p " Enter
 '

Exception:
Parameter name: left
Actual value was -2.
   at Microsoft.PowerShell.Internal.VirtualTerminal.set_CursorLeft(Int32 value)
   at Microsoft.PowerShell.PSConsoleReadLine.Insert(Char c)
   at Microsoft.PowerShell.PSConsoleReadLine.SelfInsert(Nullable`1 key, Object arg)
````

### PR DESCRIPTION
**Pull request by daxian-dbw**
Fix #919

This is one variant of the many `ArgumentOutOfRangeException` issues.
For this particular issue, the exception happens in one of the 3 conditions:
1. Resize the windows, so the cursor appears in the next line (1 issue)
2. Prompt string gets longer, and cause the cursor to appear in the next line. (3+ issues)
3. Prompt function is changed and the new prompt string is too short. (3+ issues)

The fix is simple, just don't change the color of the prompt in those cases.

--------------

Issue #932 is an instance caused by changing the prompt function to produce a very short string: `$`.
Issue #849 is another instance caused by changing the prompt function to produce an extra new line at the end of the prompt message.
Issue #843 is another instance caused by changing the prompt function to produce extra new lines at the end of the prompt message.

I believe issue #863 happens because after `cd` into a nested folder, the prompt string is long enough to result in a wrapped line.
I believe issue #875 is also because `cd` results in a longer prompt string that gets wrapped into the next line.
Issue #843 also has a report about a long prompt causing the cursor to appear in its own line (line was wrapped).

Issue #786 is an instance where the cursor appears in its own line because of window resizing.

### DIFF
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -377,16 +377,38 @@ namespace Microsoft.PowerShell
                     // We need to update the prompt
 
                     // promptBufferCells is the number of visible characters in the prompt
-                    var promptBufferCells = LengthInBufferCells(promptText);
+                    int promptBufferCells = LengthInBufferCells(promptText);
+                    bool renderErrorPrompt = false;
 
-                    // The 'CursorLeft' could be less than error-prompt-cell-length when
-                    //   1. console buffer was resized, which causes the initial cursor to appear on the next line;
-                    //   2. prompt string gets longer (e.g. by 'cd' into nested folders), which causes the line to be wrapped to the next line;
-                    //   3. the prompt function was changed, which causes the new prompt string is shorter than the error prompt.
-                    // when this happens, we skip changing the color of the prompt.
                     if (_console.CursorLeft >= promptBufferCells)
                     {
+                        renderErrorPrompt = true;
                         _console.CursorLeft -= promptBufferCells;
+                    }
+                    else
+                    {
+                        // The 'CursorLeft' could be less than error-prompt-cell-length in one of the following 3 cases:
+                        //   1. console buffer was resized, which causes the initial cursor to appear on the next line;
+                        //   2. prompt string gets longer (e.g. by 'cd' into nested folders), which causes the line to be wrapped to the next line;
+                        //   3. the prompt function was changed, which causes the new prompt string is shorter than the error prompt.
+                        // Here, we always assume it's the case 1 or 2, and wrap back to the previous line to change the error prompt color.
+                        // In case of case 3, the rendering would be off, but it's more of a user error because the prompt is changed without
+                        // updating 'PromptText' with 'Set-PSReadLineOption'.
+
+                        int diffs = promptBufferCells - _console.CursorLeft;
+                        int newX = bufferWidth - diffs % bufferWidth;
+                        int newY = _initialY - diffs / bufferWidth - 1;
+
+                        // newY could be less than 0 if 'PromptText' is manually set to be a long string.
+                        if (newY >= 0)
+                        {
+                            renderErrorPrompt = true;
+                            _console.SetCursorPosition(newX, newY);
+                        }
+                    }
+
+                    if (renderErrorPrompt)
+                    {
                         var color = renderData.errorPrompt ? _options._errorColor : defaultColor;
                         if (renderData.errorPrompt && promptBufferCells != promptText.Length)
                             promptText = promptText.Substring(promptText.Length - promptBufferCells);
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -378,13 +378,22 @@ namespace Microsoft.PowerShell
 
                     // promptBufferCells is the number of visible characters in the prompt
                     var promptBufferCells = LengthInBufferCells(promptText);
-                    _console.CursorLeft -= promptBufferCells;
-                    var color = renderData.errorPrompt ? _options._errorColor : defaultColor;
-                    if (renderData.errorPrompt && promptBufferCells != promptText.Length)
-                        promptText = promptText.Substring(promptText.Length - promptBufferCells);
-                    UpdateColorsIfNecessary(color);
-                    _console.Write(promptText);
-                    _console.Write("\x1b[0m");
+
+                    // The 'CursorLeft' could be less than error-prompt-cell-length when
+                    //   1. console buffer was resized, which causes the initial cursor to appear on the next line;
+                    //   2. prompt string gets longer (e.g. by 'cd' into nested folders), which causes the line to be wrapped to the next line;
+                    //   3. the prompt function was changed, which causes the new prompt string is shorter than the error prompt.
+                    // when this happens, we skip changing the color of the prompt.
+                    if (_console.CursorLeft >= promptBufferCells)
+                    {
+                        _console.CursorLeft -= promptBufferCells;
+                        var color = renderData.errorPrompt ? _options._errorColor : defaultColor;
+                        if (renderData.errorPrompt && promptBufferCells != promptText.Length)
+                            promptText = promptText.Substring(promptText.Length - promptBufferCells);
+                        UpdateColorsIfNecessary(color);
+                        _console.Write(promptText);
+                        _console.Write("\x1b[0m");
+                    }
                 }
             }
 

### CODE REVIEW

File: PSReadLine/Render.cs [hunk 1]
```diff
@@ -378,13 +378,21 @@ void UpdateColorsIfNecessary(string newColor)
...
-                        promptText = promptText.Substring(promptText.Length - promptBufferCells);
-                    UpdateColorsIfNecessary(color);
-                    _console.Write(promptText);
-                    _console.Write("\x1b[0m");
+
+                    // The 'CursorLeft' could be less than error prompt cell length when
+                    //   1. console buffer was resized and the regular prompt takes a whole line of the new width buffer; OR
+                    //   2. the prompt function was changed and the new prompt is shorter than the error prompt.
+                    // when this happens, we skip changing the color of the prompt.
+                    if (_console.CursorLeft >= promptBufferCells)
```
lzybkr, line 383: It seems like this could be confusing - e.g. the prompt text would stay red even though it shouldn't be anymore.

I'm also surprised that line 368 works correctly after resize. In fact, that seems like a bigger problem and part of the necessary fix here to correctly change the prompt color back.

In the second case, that seems like possibly something to warn the user about? The `PromptText` option should normally be a very short string.

daxian-dbw: > I'm also surprised that line 368 works correctly after resize

The x and y are adjusted in `RecomputeInitialCoords()`, and that's why line 368 works correctly.

> part of the necessary fix here to correctly change the prompt color back

Hmm, I don't think this will be a problem when the above 2 conditions happen.
If the prompt color was changed to red previously, that means the editing is still in the same logical line, and thus `_initialX` and `_initialY` will not be changed, namely `_console.CursorLeft` will not change.

> In the second case, that seems like possibly something to warn the user about?

Maybe. There are actually a couple of tiny problems when the prompt is changed:
1. the error prompt is not changed, so when there is any parsing error, we will end up changing the prompt text. For example, prompt is changed to `PS>` while error prompt is still `> `, then the prompt displayed will be `P>>` when there is a parsing error.
2. we don't know the prompt is changed unless we monitor the prompt function (obj reference equality? definition text equality?), and we will have to get the error prompt again based on the new prompt string.
My take is that changing prompt function within a session should be very rare and it's not worth the effort to track the changes and make things look naturally good when it's changed. What's your thoughts?

lzybkr: Isn't this code hit only when the color needs to change, either to red or back to normal (line 375?)

Assuming that, not doing it will be misleading, but it's possible it was already misleading, in other words, a color change was missed already?

If the cursor is placed correctly after resize, you know the cursor is immediately after the prompt. Moving back is slightly more complicated when you have to span lines, but it's not hard.

As for knowing about the prompt change - the idea is to only color the end of the prompt - we infer the last non-whitespace character plus the trailing whitespace (so we can move the cursor backwards correctly).

Right now, if we can't infer what to color, it's the users responsibility to configure PSReadLine correctly. It would be nice to warn when it's not configured properly, but I don't think that's easy, and maybe not worth the complexity.

daxian-dbw: > Isn't this code hit only when the color needs to change, either to red or back to normal (line 375?)
Assuming that, not doing it will be misleading, but it's possible it was already misleading, in other words, a color change was missed already?

I thought changing color from red back to normal won't be affected even after resizing the buffer width, since once the color has been successfully changed to red, it can be changed back to normal only if the same logical lines are being edited, and in that case, `_initialX` and `_initialY` won't be changed even if a change in the buffer width happens. It turns out I was wrong at one edge case -- take `ps C:\Users\UserA\AppData\Local\Microsoft\ApplicationInsights> 'abcde` as an example, if the buffer width shrank to the point that the very first  single quote `'` gets shuffled to the next line, then `_initialX` and `_initialY` get changed (something seems wrong in this case), and a closing single quote will trigger the `ArgumentOutOfRangeException`. In this particular case, the red `>` from the above line won't be changed back to normal with this PR.

I will redo the fix to handle coloring in the "buffer resizing" situation.
But for the "prompt change" situation, I think it makes sense to just skip coloring, unless we want to go with the warning.

> It would be nice to warn when it's not configured properly, but I don't think that's easy, and maybe not worth the complexity.

Same thoughts here.

daxian-dbw: @lzybkr Is there a way to check if the current cursor is on a wrapped line (it's actually on the same line as the previous line and just got wrapped into a new line because the line is longer than the width buffer)?
I believe that knowledge is needed in order to decide if the `CursorLeft` should be changed to point to the end of the previous line.
After executing a command, `_initialX` could become less than `promptBufferCells` for a few reasons:

1. console buffer got resized, in which case we would know before executing the command.
2. executing the command causes the prompt to be longer just enough to cause the line to be wrapped.
3. the `prompt` function is changed to produce a very short prompt string, say 1 character.

Ideally, we should adjust `CursorLeft` to the end of previous line only for case 1 and 2, where the cursor is actually on the same line (wrapped). For case 3, we should just skip changing the color.

I could evaluate the prompt function in case `_console.CursorLeft >= promptBufferCells`, so as to decide if the cursor is on a wrapped line (length of the prompt string being equal or longer than buffer width), but I really don't want to do that if there is another way to find out this.

lzybkr: I forget the details, but the logic in `ReallyRender` needs the same sort of thing - that's the difference between a "physical line" (a line as rendered in the console) and a logical line (which, when long, might span more than 1 physical line.) It's mostly simple logic, the first line is special cased because of the prompt, otherwise it's simple arithmetic between the character (or in your case cursor position) and the buffer width.

Note that in versions <2.0, we never needed to reevaluate the prompt because we could scrape the console buffer. That's no longer possible in a portable way, but I've avoided reevaluating the prompt because people do sometimes display things like the time, and it could be misleading if the time changes while editing the command line. There are of course times when we must run the prompt again, but that happens either because the user asked for it (redraw over the current prompt), or after showing possible completions where the prompt needs to be drawn *below* those completions.

daxian-dbw: The `logical line` and `physical line` calculation in `ReallyRender` doesn't help here.
The existing `logical line` and `physical line` calculation is simple because we have all the text that are going to be rendered. But here, we need to decide if the current cursor is within the same line as the previous line, which has already been rendered on the console (by PowerShell itself writing out the prompt string). We don't have the text at hand to calculate if the current cursor is on a wrapped line or a new line.

The only way I can think of is to reevaluate the prompt function **without rendering it on the console**, then we have the text and can do the calculation to decide if the current cursor is on the wrapped line or the new line. Do you think this approach is acceptable?

lzybkr: You can't evaluate the prompt to capture it's output. Some prompts write some/much of their content directly, e.g. via `Write-Host` so they can get colors.

Are you just wanting `ConvertOffsetToPoint`?

daxian-dbw: About the prompt evaluation, you are right, no guarantee to capture its exact output. So that approach won't work.

`ConvertOffsetToPoint` is not what I want. It's useful for calculating cursor position but won't help to determine if the cursor is in the wrapped line of the previous line.

Here is an example where the cursor should be adjusted to the end of the previous physical line to change the color of `>`, because the cursor is actually in the wrapped line (same logical line).

image

Here is an example where the cursor shouldn't be adjusted because it's literally on a new line, not the wrapped line of the previous physical line.

image

The problem is I have trouble telling which cases we are in when `_console.CursorLeft >= promptBufferCells`. Monitoring the change of `prompt` function doesn't help too, because there is no way to tell whether it's changed to produce a very short string, like `:` in the example above, or it's now producing a longer prompt string that causes the line to be wrapped.

daxian-dbw:
Issue #932 is an instance caused by changing the prompt function to produce a very short string: `$`.

PSReadLine/Render.cs
SeeminglyScience: > You can't evaluate the prompt to capture it's output. Some prompts write some/much of their content directly, e.g. via `Write-Host` so they can get colors.

Since PSRL only supports 5.1+ you can handle that:

```csharp
using (var ps = PowerShell.Create(RunspaceMode.CurrentRunspace))
{
    var command = new Command("profile", isScript: false, useLocalScope: true);
    command.MergeMyResults(PipelineResultTypes.Information, PipelineResultTypes.Output);
    ps.Commands.AddCommand(command);
    var includesInfoRecords = ps.Invoke();
}
```

Accounting for things like escape sequences and `\b` would add a lot of complexity though.

daxian-dbw: IMO, fixing the color changing of `>` in this edge case doesn't worth the effort.
If we accept this PR as it is, then in the scenarios shown as above screenshots, what would happen is just the prompt's color won't change when there is any parsing errors. This is fine, as we don't change the color anyways when the prompt function [is not pure](https://github.com/PowerShell/PSReadLine/blob/master/PSReadLine/ReadLine.cs#L746-L765).

The worst case would be when the user started by typing on the same line:
image

And then, the user carefully resizes the window to make the typed text appear on the next line:
image

Then in this case, the existing red `>` won't be changed back to normal when the parsing error is gone. But it affects that particular prompt line only.
I think this is acceptable, as the user needs to carefully crafted it to make this happen.

lzybkr: It sounds like it might be hit somewhat frequently though, otherwise how did you end up finding this? Or did you find it from the short prompt and then realized this case as well?

Do note that `PromptText` is user configurable for non-trivial prompts.

It bothers me because somewhat sticky (you'd have to introduce another error then fix that before it is correct.)  I think people will <kbd>Ctrl+c</kbd> if they notice assuming the prompt is broken.

At any rate, isn't the correct fix to simply wrap to the previous line as though you were pressing left arrow?

daxian-dbw: > It sounds like it might be hit somewhat frequently though, otherwise how did you end up finding this? Or did you find it from the short prompt and then realized this case as well?

I'm not sure which case you are referring to here.
I went through all the `ArgumentOutOfRangeException` issues with the `Parameter name: left` being the problem, and they can be categorized in to 3 buckets. Please see the PR description for the details.
Basically, users run into this with one of 3 conditions:
1. Resize the windows, so the cursor appears in the next line (1 issue)
2. Prompt string gets longer, and cause the cursor to appear in the next line. (3+ issues)
3. Prompt function is changed and the new prompt string is too short. (3+ issues)

None of them result in the "worst case" as shown in https://github.com/PowerShell/PSReadLine/pull/967#discussion_r305447896. I found the "worst case" when I started to really play with some edge cases.

> At any rate, isn't the correct fix to simply wrap to the previous line as though you were pressing left arrow?

If we know we should do this, then it's simple to do. The problem is we don't know if we should do this.
See the screenshots in https://github.com/PowerShell/PSReadLine/pull/967#discussion_r305100779, when the prompt function is changed to produce a short prompt string, we shouldn't wrap to the previous line.
This is why I asked if there is a way to tell whether or not the cursor is on a wrapped line. So far, I don't find an easy way to determine it.
@SeeminglyScience had a suggestion at https://github.com/PowerShell/PSReadLine/pull/967#discussion_r305322159, but my personal opinion is the value doesn't worth the complexity.

lzybkr: Thanks, this makes more sense now.

So for 1 & 2, assuming we wrapped is correct.
For 3, assuming we wrapped is wrong.

If we always assume we wrapped and fix appropriately, we:
* avoid an exception
* get the color correct for cases 1 & 2
* get fewer bug reports
* in case 3, rendering might be off

We can also argue case 3 is user error - and that they need to update PromptText. I think we might see fewer bug reports too, especially if we publish a sample recipe for changing the prompt (like I have a `Push-Prompt` and `Pop-Prompt` pair of functions.

Is this reasonable?

daxian-dbw: This sounds reasonable. Great, it's a much easier problem if we don't need to accurately differentiate 1,2 from 3.

daxian-dbw:
@lzybkr The change has been update to wrap back to the previous line. Please take another look and let me know your concerns.

lzybkr (approved):

PR #984, opened 2019-07-31
Correct highlight of replaced variables/splats with braces

### LINKED ISSUE #983: ArgumentOutOfRangeException: Completion of Splat, but Not $
Using a custom build of PS 7 Preview, but also applies to 5.1 and 6/6.1/6.2 in the right conditions.

I have been working with the PowerShell code to try to get completion starting at a splat `@` to work.  With that working, the current completion code completes the `$` variable as `@{$}`, which is knowingly invalid, and it causes an ArgumentOutOfRange exception.

## Steps to reproduce or exception report

This reproduction is for the released builds of PowerShell, as they will not currently produce completions for just a `@` argument.

Complete the following:
```PowerShell
echo @?<CTRL SPACE> # crashes after list is displayed, but before the first result is inserted
echo $<CTRL SPACE> # notice selection varies with variable names that require a brace
```

Note, the window must be large enough as to be able to display the entire list without the input line scrolling off the top.  Also, I've only tested this in Windows.

## Code in Question

I traced down the code that is calling the exception:
https://github.com/PowerShell/PSReadLine/blob/b7b8f7bb1dd094cc86fc71ea74e7e8427b6415e0/PSReadLine/Completion.cs#L732-L737
I think the check for a `{` is incorrect here, and that there is no reason to ever substring the original text, removing the prefix, which causes the completion to look like its replacing the prefix as well.

msftrncs:
- Prevent adjustment of `userCompletionText` in `FindUserCompletionTextPosition` when the match contains `{` in CompletionText[1] if the `userCompletionText` is not longer than 1 char
- Preserve the splat sigil at start of completions of type `Variable`.

Adjusting `userCompletionText` blindly led to `ArgumentOutOfRangeException` on incorrectly braced splats on input `@?`.
Loss of the sigil of splats led to the splat sigil always being marked as replaced, and in situations where the PowerShell completion logic returned incorrect completions of splats with braces, led to a `ArgumentOutOfRangeException`.

Fix #983.

### DIFF
--- a/PSReadLine/Completion.cs
+++ b/PSReadLine/Completion.cs
@@ -106,7 +106,7 @@ namespace Microsoft.PowerShell
             {
                 if (IsQuotedVariable(s))
                 {
-                    return '$' + s.Substring(2, s.Length - 3);
+                    return s[0] + s.Substring(2, s.Length - 3);
                 }
                 return s;
             }
@@ -731,7 +731,7 @@ namespace Microsoft.PowerShell
 
         private int FindUserCompletionTextPosition(CompletionResult match, string userCompletionText)
         {
-            return match.ResultType == CompletionResultType.Variable && match.CompletionText[1] == '{'
+            return match.ResultType == CompletionResultType.Variable && userCompletionText.Length > 1 && match.CompletionText[1] == '{'
                 ? match.CompletionText.IndexOf(userCompletionText.Substring(1), StringComparison.OrdinalIgnoreCase) - 1
                 : match.CompletionText.IndexOf(userCompletionText, StringComparison.OrdinalIgnoreCase);
         }


msftrncs:
I think I figured out what the check for a `{` is doing …  I'll probably implement a different correction.

The detection for a `{` appears to be to handle the situation where `@en` is the input, and then when `@{env:blah(blah)}` is the completion, the selection becomes 'v:blah(blah)}`, not selecting the first brace.

msftrncs:
I've amended the commit for this PR.  ~Its very possible that either of the two changes would stop the ArgumentOutOfRangeException from occuring.  The correction for maintaining the sigil of the returned completions is probably the item that solves the problem, but the other change is a safety net.~  (See comments below)

daxian-dbw (approved):
LGTM

daxian-dbw:
> The correction for maintaining the sigil of the returned completions is probably the item that solves the problem, but the other change is a safety net.

The correction for maintaining the sigil of the returned `userCompletionText` is the item that solves the `ArgumentOutOfRangeException`, but the other change is needed to have the correct selection for `${$}` when typing `echo $`<kbd>Ctrl+Space</kbd>.

msftrncs:
@daxian-dbw, you are right, I was so focused on the `@` sigil, I overlooked the importance the other change (to FindUserCompletionTextPosition) had on the `$` selection when the `{` was present in the completion.

PR #985, opened 2019-07-31
Add missing curly quotes in check functions

msftrncs:
This adds the quotes \`u{201B} (single curly quote) and \`u{201E} (double curly quote), as they seem to have been left out, so this makes the check functions complete with regards to PowerShell syntax.

However, I am not sure that any active code currently reaches these quotes, as the completion code in PowerShell replaces curly quotes with their standard variation.


daxian-dbw (approved):
LGTM.

lzybkr (approved):

Issue #836, opened 2018-12-18
incorrect highlight of command with variable


msftrncs:
The syntax highlighting often gives the user the idea of what is going to happen, but in a few situations it seems to indicate the wrong things.

```PowerShell
$a='goodbye'
hello$a # 'hello' is yellow, '$a' is green like its going to be expanded, but it doesn't
# instead, "The term 'hello$a' is not recognized as the name of a …" is returned.
```

I think it is safe to say, everything highlighted yellow, does not allow expansion

```PowerShell
function hello () {echo $args}
$a= @{hello=3;goodbye=4}
$a[hello] # 'hello]' is yellow, like it could be a function call, but function calls seem not to be permitted in array index
$a[hello goodbye] # as you type, it looks like its going to be valid, but the prompt knows, its still red at the end.
$a[(hello goodbye)] # this is accepted
```
![image](https://user-images.githubusercontent.com/26179051/50131055-02d52480-0246-11e9-99c9-391a50365b97.png)

This seems to really confuse the syntax highlighting:
```PowerShell
$a[(hello$b goodbye)]
```
![image](https://user-images.githubusercontent.com/26179051/50131831-abd14e80-0249-11e9-8582-876fff2f0323.png)

Environment data
----------------
```
PS version: 6.1.1
PSReadline version: 2.0.0-beta3
os: 10.0.18298.1000 (WinBuild.160101.0800)
PS file version: 6.1.1
```

lzybkr:
The first case is easy to fix, PSReadLine shouldn't look at nested tokens when the TokenFlag `CommandName` is set on the token.

Your other examples are accurately representing how the PowerShell parser works. When the parser doesn't find an expression for an index, it assumes you're missing a closing `]`. It inserts an artificial `]` and keeps parsing, finding a valid command invocation.

msftrncs:
Quickly realized this occurs in a few other situations as well.  Its like the syntax highlighting is eager to find a command.

(assume 'hello' is the function defined in the earlier post)
```PowerShell
[int]hello 3 # prompt remains red, but 'hello' is yellow
[invalid hello 3 # eager to have a command, 'hello' is yellow
[$b]3 # all the text except $b is yellow, $b is green, but no expansion is allowed here, and the whole thing is actually tokenized as a command name
345+65.com # 65.com becomes yellow, but this is an expression, so commands are no longer possible.
$a = hello 3 # this is valid
[...]

PR #989, opened 2019-08-03
Prioritize highlight of command names, prevent bleed from aligned nested tokens

msftrncs:
Referencing #836, 
- Prioritize the selection of the command color when a token is flagged as 'CommandName'. (to handle a parameter token that is flagged as a command name, it is really a command name)  
- Prevent the bleed of a token with nested tokens in to the rest of the statement.

### DIFF
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -215,7 +215,7 @@ namespace Microsoft.PowerShell
                     // use the tokens color otherwise use the initial color.
                     var state = tokenStack.Peek();
                     var token = state.Tokens[state.Index];
-                    if (i == token.Extent.EndOffset)
+                    while (i == token.Extent.EndOffset)
                     {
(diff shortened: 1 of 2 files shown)


msftrncs:
I added another commit to complete what can be fixed in #836, where nested tokens would bleed the parent tokens color unto the rest of the statement, if the last nested token ended together with the parent token.  In the case of `& hello$ab -hello`, lets not prioritize command names so much the next item gets the same color.

daxian-dbw (approved):
Could you please add a test? I'm not sure if it's easy to do so though ...

msftrncs:
I have added a commit that implements two test scenarios, each of which attempts to test for the two items addressed above under slightly different scenarios.

msftrncs:
I force pushed as my previous commit was trying to change (remove) the BOM.

daxian-dbw (approved):
LGTM. Thanks!

PR #1003, opened 2019-08-21
Operator Highlight in ArgumentMode

### LINKED ISSUE #1002: Operators highlight in argument mode when not operators
Current behavior:
image

Operators highlight in argument mode, where they are not active as operators.  This is caused by the PowerShell tokenizer changing the TokenKind after the parser realizes they should instead be `Generic`, but it does not clear the TokenFlags that identify the token as belonging to a class of operators.

Expected behavior: (sorry its in a different theme)
image

msftrncs:
Operators shouldn't highlight in argument mode.

See https://github.com/PowerShell/PowerShell/issues/10348

Operators as TokenKind `Generic` are not actually operators.  This PR restricts the operator token detection to just those tokens that are not TokenKind `Generic`.

Fix #1002.

### DIFF
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -663,7 +663,7 @@ namespace Microsoft.PowerShell
                 return _options._keywordColor;
             }
 
-            if ((token.TokenFlags & (TokenFlags.BinaryOperator | TokenFlags.UnaryOperator | TokenFlags.AssignmentOperator)) != 0)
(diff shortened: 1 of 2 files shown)


lzybkr (dismissed):

lzybkr (approved):
The code changes look good, but it would be a good idea to add a quick test in test\RenderTest.cs.

msftrncs:
I made a simple addition for testing, testing against `master` first to confirm the tests would fail previously.

daxian-dbw:
Like being said in the referenced issue, it probably should be fixed in parser.
But fixing it in PSReadLine is nice to have since the tokens will be rendered more accurately on old versions of powershell.

PR #1028, opened 2019-09-09
Fix arrow-up when the cursor is at the end of a wrapped line in a multiple-line text

### LINKED ISSUE #1020: editing wrapped line in history, up-arrow doesn't function at end of wrapped line.
Using a build of PSReadLine from 'master' + @daxian-dbw's PR #979
Windows 10 Insiders
PowerShell 7 beta 3

scroll back through history, and edit one where at least one line is so long it wraps in the terminal.  On that line, arrow out to the end of the line, then attempt to go up by pressing up-arrow.  Nothing will happen.  Going down will work, but coming back up will again stop at this line.  Pressing left-arrow once, then up-arrow, and the cursor moves as expected.

image

daxian-dbw:
Fix #1020

When the cursor is at the end of a wrapped line, <kbd>Up</kbd> key sets the potential new cursor position to the previous physical line, and then calculates the end of the logical line rendered in that physical line, and eventually ends up in the same old cursor position.
That's why <kbd>Up</kbd> at the end of a warped line doesn't work.
The fix is to not depend on the physical line position when the original cursor is at the end of line, but directly search back (for <kbd>Up</kbd>) and search forward (for <kbd>Down</kbd>) for new line characters.

### DIFF
--- a/PSReadLine/Movement.cs
+++ b/PSReadLine/Movement.cs
@@ -82,7 +82,8 @@ namespace Microsoft.PowerShell
             //  - If the cursor is at the end of a logical line, then 'UpArrow' (or 'DownArrow') moves the cursor up (or down)
             //    'lineOffset' numbers of logical lines, and the cursor is always put at the end of the new logical line.
             //  - If the cursor is NOT at the end of a logical line, then 'UpArrow' (or 'DownArrow') moves the cursor up (or down)
-            //    'lineOffset' numbers of physical lines, and the cursor is always placed at the same column as is now.
+            //    'lineOffset' numbers of physical lines, and the cursor is always placed at the same column as is now, or at the
+            //    end of line if that physical line is shorter than the targeted column.
 
             const int endOfLine = int.MaxValue;
 
--- a/PSReadLine/Movement.cs
+++ b/PSReadLine/Movement.cs
@@ -80,6 +80,18 @@ namespace Microsoft.PowerShell
         {
             const int endOfLine = int.MaxValue;
 
+            // Actually not moving the line.
+            if (numericArg == 0)
+            {
+                return;
+            }
+
+            // Moving the line down when it's actually at the end of the last line.
+            if (numericArg > 0 && _current == _buffer.Length)
+            {
+                return;
+            }
+
             _moveToLineCommandCount += 1;
             var point = ConvertOffsetToPoint(_current);
             if (_moveToLineCommandCount == 1)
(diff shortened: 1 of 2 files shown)


daxian-dbw:
I will try adding a test.

daxian-dbw:
@msftrncs Would you mind review the changes? Thanks!

PSReadLine/Movement.cs
```diff
@@ -90,25 +102,53 @@ private void MoveToLine(int numericArg)
                         : point.X;
             }
 
-            var topLine = _initialY;
+            int newCurrent;
```
rjmholt, line 110: Given all the new logic added and the single integer output, this might be worth putting into a new method.

PSReadLine/Movement.cs
```diff
@@ -90,25 +102,53 @@ private void MoveToLine(int numericArg)
...
+            if (_moveToLineDesiredColumn == endOfLine)
+            {
+                newCurrent = _current;
+
+                if (numericArg > 0)
+                {
+                    // Moving to a subsequent line.
+                    for (int i = 0; i < numericArg; i++)
+                    {
+                        for (newCurrent++; newCurrent < _buffer.Length && _buffer[newCurrent] != '\n'; newCurrent++) ;
```
rjmholt, line 120: Does buffer hold physical lines, which themselves can be multiple logical lines if they contain `\n`?

If so, is there a danger arrow-up will move up a logical line at the end of a line, but a physical one everywhere else?

I'm suspicious because of how different the logic for moving up a line is now when the cursor is at the end of a line, compared to this being either a small edge-case catch or a broader change that deals with logical vs physical lines everywhere

PSReadLine/Movement.cs
```diff
@@ -80,6 +80,18 @@ private void MoveToLine(int numericArg)
         {
             const int endOfLine = int.MaxValue;
 
+            // Actually not moving the line.
+            if (numericArg == 0)
```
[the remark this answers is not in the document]
daxian-dbw: It is the offset, but not necessarily to the physical line of the cursor. When the cursor is at the end of a logical line, it means the offset to the logical line.

The `numericArg` is a parameter name pattern used almost for all key binding functions. I actually means `the number of such operations`, and is specified by user typing <kbd>Alt+Num</kbd> following the operation key. For example, <kbd>Alt+3</kbd><kbd>Backspace</kbd> means do <kbd>Backspace</kbd> 3 times.

PSReadLine/Movement.cs
daxian-dbw: I feel the logic is too simple/small to put in a separate method ... :)

PSReadLine/Movement.cs
daxian-dbw: `_buffer` holds the user input text, something like `fff\nfffffff\nffff` for the following input:
```
PS> fff
>> fffffff
>> ffff
```
I'm not sure I fully understand your comment, but the behavior I think it should have is:
1. if the cursor is **NOT** at the end of a logical line, then <kbd>Up</kbd> and <kbd>Down</kbd> move the cursor up/down 1 or N **physical lines** (depends on `numericArg`). This is the existing behavior.
2. if the cursor is at the end of a logical line, then <kbd>Up</kbd> and <kbd>Down</kbd> move the cursor up/down 1 or N **logical lines** (depends on `numericArg`), and cursor is always put at the end of the new logical line.

Let's sync tomorrow to understand your concerns.

msftrncs:
This does resolve the problem.  

When the cursor has been intentionally moved to the end of the wrapped line (or came from being intentionally placed at the end of a line below or a line above), pressing up arrow will take the cursor up to the end of the next full line up.   

On the other hand, if the cursor landed at the end of a wrapped line with a up-arrow not from the end of the line below, or down-arrow not from the end of the line above, the cursor resumes its normal place when going up or down another line (when it is able to).

I checked this behavior with multiple wrapped lines in a row, with both existing history from my session and with some jibberish here-strings quickly formulated.

I did find a new behavior that seemed not to exist before (I cannot repeat it with Windows PowerShell which I still have running PSReadLine 2.0.0 beta ~1~2):  if the cursor is logically not at the end of the line, and down arrow is pressed on last line but yet the cursor's physical position is at the end of that line, the cursor becomes logically assigned to the end of the line, so then moving up takes the cursor through the end of the logical lines again.  This does not require any lines that wrap.  Simple example:

```PowerShell
@'
hello
`@
```
With the cursor sitting on 'o' (not after it), pressing down and back up work as expected, but pressing down twice, then back up takes the cursor to past the 'o', as now it moves to the logical line ends.

lzybkr (approved):

Issue #911, opened 2019-05-07
Crash when deleting text from start-process command


sirandyb:
Environment data
----------------

PS version: 5.1.17763.316
PSReadline version: 2.0.0-beta2
os: 10.0.17763.1 (WinBuild.160101.0800)
PS file version: 10.0.17763.1 (WinBuild.160101.0800)

Steps to reproduce or exception report
--------------------------------------
Oops, something went wrong.  Please report this bug with the details below.
Report on GitHub: https://github.com/lzybkr/PSReadLine/issues/new
-----------------------------------------------------------------------
Last 55 Keys:
 s t a r t - p r o c e s s Space $ e n v : w i n d i r \ n o t e p a d . e x e Space | Space o u t - n u l l Home Delete Delete Delete Delete Delete Delete

Exception:
System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at Microsoft.PowerShell.PSConsoleReadLine.GenerateRender(String defaultColor)
   at Microsoft.PowerShell.PSConsoleReadLine.ForceRender()
   at Microsoft.PowerShell.PSConsoleReadLine.DeleteCharImpl(Int32 qty, Boolean orExit)
   at Microsoft.PowerShell.PSConsoleReadLine.DeleteChar(Nullable`1 key, Object arg)
   at Microsoft.PowerShell.PSConsoleReadLine.ProcessOneKey(ConsoleKeyInfo key, Dictionary`2 dispatchTable, Boolean ignoreIfNoAction, Object arg)
   at Microsoft.PowerShell.PSConsoleReadLine.InputLoop()
   at Microsoft.PowerShell.PSConsoleReadLine.ReadLine(Runspace runspace, EngineIntrinsics engineIntrinsics)
-----------------------------------------------------------------------

msftrncs:
@daxian-dbw, have you started on this yet?  I guess I should have seen this when I worked on PR #989 (token bleed), as it would seem the solution to this is to add a `continue` (now available because of the `while` that I added) after the `peek` when the `tokenStack` still has more items.  

Evidently, `process` is one of very few situations in the parser that the parser stops processing due to an error (or unexpected situation), as nothing past ~a pipe, when the pipe follows a `process` keyword~ the next token after `process` gets parsed (including no EOS token).

I'll post a PR shortly.

daxian-dbw:
I already have a fix ready. Will submit a PR soon.

Get Outlook for iOS<https://aka.ms/o0ukef>
[...]

PR #1049, opened 2019-09-15
Fix GenerateRender

daxian-dbw:
Fix #911

Fix two issues:
1. `process $a\b ` results in `IndexOutOfRangeException`. Fixed by the `continue` change.
2. `process out ` results in `NullReferenceException`. Fixed by the `break` change.

/cc @msftrncs

PR #1050, opened 2019-09-15
Alternate: Fix IndexOutOfRangeException in GenerateRender

msftrncs:
Offering slightly different, but really the same, fix for #911, along with a test, but with less explanation.

/cc: @daxian-dbw

### DIFF
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -226,7 +226,7 @@ namespace Microsoft.PowerShell
                     var token = state.Tokens[state.Index];
                     while (i == token.Extent.EndOffset)
                     {
-                        if (token == state.Tokens[state.Tokens.Length - 1])
+                        if (state.Index == state.Tokens.Length - 1)
                         {
                             tokenStack.Pop();
                             if (tokenStack.Count == 0)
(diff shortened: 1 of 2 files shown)


daxian-dbw:
I like your changes better. I will update mine to be the same as yours with additional comments and one more test to cover the `NullReferenceException`.

daxian-dbw:
@msftrncs Yes :)

msftrncs:
I almost submitted #989 with a the `break` in it, but thought the EOS would always be there.   I'm mentioning this for reference, as I should have added `break` when adding the `while`, because `token` is made null, and then without a break `token` is used in a reference.  My gut said the `break` was needed, I just couldn't see why immediately.  😃

daxian-dbw:
No worries at all. I didn't realize the `stop early` design for parsing `NamedBlock` until looking at this issue.
I looked into PowerShell code to see if `process xxxx xxx`, `begin xxxx ...` and `end xxx ...` should be fixed to generate tokens in a more expected/friendly way.
It doesn't look like a simple change to make it happen. The design is to stop parsing when `process`, `begin` and `end` don't end up with a valid `NamedBlock`, so tokenizing terminates in the middle as by design in those cases.
Fixing it in PSReadLine makes more sense.

daxian-dbw:
@msftrncs Do you mind taking a look at my updates to #1049?
I took your changes and added additional comments and changed the tests a bit.
I didn't update your branch directly only because it's easier to work in my branch :)

daxian-dbw:
Also, I invited you as a contributor to the PSReadLine repo, would you like to accept it? 😄

daxian-dbw:
Close this one as #1049 has been merged. Thanks @msftrncs!

PR #1058, opened 2019-09-18
Filter sensitive history items and avoid writing them to the history file

### LINKED ISSUE #934: Detect and scrub sensitive information in history
There are a number of tools that reads PSReadLine's history file to look for passwords provided in cleartext.  We should see if we can have an option (perhaps default) to detect and scrub any content that could be considered sensitive.

daxian-dbw:
Fix #934

Detect sensitive history items by using the regex `"password|asplaintext|token|key|secret"` [suggested by Lee](https://twitter.com/Lee_Holmes/status/1172640465767682048).

For sensitive history items, we still keep them in the history queue, so they are still accessible within the same session.
However, we don't write them to the history file, so they will be gone when the session ends, just like the built-in powershell history.

This is done by adding a default handler to `AddToHistoryHandler`. Quoted from the updated doc for `-AddToHistoryHandler`:
> Specifies a ScriptBlock that can be used to control which commands get added to PSReadLine history,
and whether they should be saved to the history file.
>
> The ScriptBlock is passed the command line, and it is expected to return either a Boolean value,
or an enum value of the type `[Microsoft.PowerShell.AddToHistoryOption]`.
The enum type `AddToHistoryOption` has 3 members: `SkipAdding`, `MemoryOnly`, and `MemoryAndFile`.
>
> If the ScriptBlock returns `$true`, it's equivalent to `AddToHistoryOption.MemoryAndFile`.
The command line is added to the in-memory history queue and saved to the history file.
If the ScriptBlock returns `$false`, it's equivalent to `AddToHistoryOption.SkipAdding`,
and the command line is not added to history at all.
>
> If the ScriptBlock returns `AddToHistoryOption.MemoryOnly`, then the command line is added to the in-memory history queue,
but will not be saved to the history file.
This usually indicates the command line has sensitive content that should not be written to disk.
>
> PSReadLine provides a default handler to this option:
>     `[Microsoft.PowerShell.PSConsoleReadLine]::GetDefaultAddToHistoryOption(string line)`
> The default handler attempts to detect sensitive information in a command line by matching with a simple regex pattern:
>     `"password|asplaintext|token|key|secret"`
> When successfully matched, the command line is considered to contain sensitive content, and `MemoryOnly` is returned.
Otherwise, `MemoryAndFile` is returned.
>
> To turn off the default handler, just set this option to `$null`.


PSReadLine/History.cs
```diff
@@ -76,6 +78,11 @@ public class HistoryItem
         private const string _failedForwardISearchPrompt = "failed-fwd-i-search: ";
         private const string _failedBackwardISearchPrompt = "failed-bck-i-search: ";
 
+        // Pattern used to check for sensitive inputs.
+        private static Regex s_sensitivePattern = new Regex(
+            "password|asplaintext|token|key|secret",
```
lzybkr, line 84: My history doesn't contain any secrets (I'm careful because I'm aware of the problem), and I use most of these terms, so this would unnecessarily impact my history.

I tested this regex against my history and had 147 hits - none contained secrets.

At a minimum, one should be able to remove or customize the regex.

But I also wonder if instead we should be using the regular expression(s) the CredScan uses.

SteveL-MSFT: I don't think credscan is OSS.  Intent here is to start with something and we can improve upon it.  This regex was an example from @LeeHolmes on what he uses.  I do agree we should have an Option to opt out of this.

SteveL-MSFT (changes requested):
Have PSReadLineOption to opt out

PSReadLine/History.cs
daxian-dbw: I agree. If we want to expose this option, I think it should be a handler just like the `AddToHistoryHandler`, maybe call it `DetectSensitiveInputHandler`.
By the way, it looks we don't handle the exception thrown from the custom `AddToHistoryHandler`.

lzybkr: In some ways, education might be more valuable. After all, `bash` has the same problem.

This change is silent. What about a different change that warns instead? Warning when reading the history file might be a good option too, with a way to disable it of course.

Also - I'm following up on the possibility of using something from CredScan.

daxian-dbw: Oh, that's nice. Curious if we can borrow something simple from CredScan.
Btw, warning when reading the history file feels weird to me ...

daxian-dbw: Two options are added:
```
Set-PSReadLineOption
    [-ScrubSensitiveHistory]
    [-DetectSensitiveInputHandler <Func[String, Boolean]>]
```
By default, `ScrubSensitiveHistory` is set to `true`. When it's set to `false`, all history commands will be saved to the history file.
`DetectSensitiveInputHandler` is used only when `ScrubSensitiveHistory` is enabled, to allow a user to define what is considered to be sensitive.
When it's null, the default logic will be used, which is the regex pattern borrowed from Lee.

PSReadLine/Cmdlets.cs
```diff
@@ -136,7 +136,9 @@ public PSConsoleReadLineOptions(string hostName)
             ContinuationPromptColor = Console.ForegroundColor;
             ExtraPromptLineCount = DefaultExtraPromptLineCount;
             AddToHistoryHandler = null;
+            DetectSensitiveInputHandler = null;
             HistoryNoDuplicates = DefaultHistoryNoDuplicates;
+            ScrubSensitiveHistory = DefaultScrubSensitiveHistory;
```
lzybkr: Are 2 options really necessary? I worry when a tool has too many configuration options and really tried to avoid unnecessary configuration options to PSReadLine.

PSReadLine/History.cs
lzybkr: I don't think we'll get anything from CredScan - they worry people will easily bypass the checks intentionally if the tests are shared.

daxian-dbw: That's a very reasonable concern :)

PSReadLine/Cmdlets.cs
daxian-dbw: Well, I guess I can just leave the `DetectSensitiveInputHandler` option and make our regex pattern logic as the default handler. If user set it to `null`, then meaning they want to disable `ScrubSensitiveHistory` functionality.
How does this sound?

lzybkr: Maybe we don't need a new option at all - instead creating a default `AddToHistoryHandler`.

I guess I'm not sure it's useful to keep sensitive command lines in memory but not on disk. After all, password managers clear the clipboard after a short period.

daxian-dbw:
I think it’s useful to keep the sensitive commands in memory, so it can be recalled in the same session. After all, they will be kept in PowerShell’s history buffer anyways.

Thanks,
Dongbo
________________________________

lzybkr:
It would be generally useful for the `AddToHistoryHandler` to be capable of specifying where a command should be saved, in memory, in a file, both, or neither.

This would require a signature change to `AddToHistoryHandler`, but I think that could be done without really breaking anything because the primary scenario is binding a script block and those aren't strongly typed. If the delegate expects `object`, we can have dynamic code testing the object type to support `bool` and this new `enum`.

daxian-dbw:
With a default handler for `AddToHistoryHandler`, once a user changes the handler, how can he/she set it back to the default? Casting a method to a delegate is supported in PS Core, but not in windows powershell. When the handler is null, it should mean "don't do add-to-history" filtering, not using the default handler.
Being unable to reset the handler feels broken to me.

lzybkr:
Does save and restore not work?

```PowerShell
$handler = (Get-PSReadLineOption).AddToHistoryHandler
...
Set-PSReadLineOption -AddToHistoryHandler $handler
```

daxian-dbw:
Well, that DOES work 😄

vexx32:
Is it worth adding a `Reset-PSReadLineOption` for things like this?

PSReadLine/Cmdlets.cs
daxian-dbw: PR updated to use `AddToHistoryHandler`. Changed its type to `Func<string, object>`.

daxian-dbw:
Added a public static property `PSConsoleReadLineOptions.DefaultAddToHistoryHandler` to hold the default handler delegate.

daxian-dbw:
@lzybkr Code has been updated, can you please take another look?

lzybkr (approved):
It'd be good to get some more feedback on this design, but I approve.

docs/Set-PSReadLineOption.md
```diff
@@ -192,13 +192,25 @@ Accept wildcard characters: False
 
 ### -AddToHistoryHandler
 
-Specifies a ScriptBlock that can be used to control which commands get added to PSReadLine history.
+Specifies a ScriptBlock that can be used to control which commands get added to PSReadLine history,
```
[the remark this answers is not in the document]
daxian-dbw: > how to add a customer handler that also chains to the default.

Support `MulticastDelegate` handler? `+` and `+=` operators in PowerShell don't support combining two delegates.
I feel this might be an overkill. Let's wait for some feedback to improve it.

PSReadLine/History.cs
```diff
@@ -76,29 +78,66 @@ public class HistoryItem
...
+                {
+                    if (Options.AddToHistoryHandler == PSConsoleReadLineOptions.DefaultAddToHistoryHandler)
+                    {
+                        // Avoid boxing if it's the default handler.
+                        return GetDefaultAddToHistoryOption(line);
+                    }
 
-                return true;
+                    object value = Options.AddToHistoryHandler(line);
+                    if (value is AddToHistoryOption enumValue)
```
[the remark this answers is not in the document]
daxian-dbw: Agreed. Will update.

daxian-dbw:
@SteveL-MSFT Can you please review again? Thanks!

docs/Set-PSReadLineOption.md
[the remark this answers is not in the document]
lzybkr: Not multi-cast, but at least expose a public static method that could be called.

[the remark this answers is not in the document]
daxian-dbw: Got it. Yes, I can make the `GetDefaultAddToHistoryOption` method public. Let me know if you think the method name should be changed.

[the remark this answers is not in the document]
daxian-dbw: Updated.

PSReadLine/Cmdlets.cs
```diff
@@ -237,10 +244,12 @@ public object ContinuationPromptColor
 
         /// <summary>
         /// This handler is called before adding a command line to history.
-        /// The return value indicates if the command line should be added
-        /// to history or not.
+        /// The return value indicates if the command line should be skipped,
+        /// or added to memory only, or added to both memory and history file.
```
rkeithhill, line 248: Are you sure the "memory only" option is that useful?  I think it makes the mental model here harder than it needs to be which, IMO should be simply this - you either save a command to history or you don't.  That mental model makes writing your own custom handler easier too - just return $true/$false - not some enum value I've got to look up.

daxian-dbw: I think it’s useful to provide this option, so commands can be recalled in the same session. This IMO is especially important for the default handler to be less surprising to the user.
In the meantime, it’s still allowed to return true/false from ‘AddToHistoryHandler’, and the behavior is exactly what you want: true == MemoryAndFile, false == SkipAdding.

rkeithhill: OK, cool $true/$false still works as expected. In that case, this seems reasonable.

daxian-dbw:
@SteveL-MSFT Would you like to take another look?

SteveL-MSFT (approved):

PR #1061, opened 2019-09-20
Make the history file mutext name unique across sessions

daxian-dbw:
`GetHistorySaveFileMutexName()` uses `GetHashCode()` to construct the mutex name.
`GetHashCode()` in .NET Core returns a value that is unique in the current process only, and it results in different mutex name being used for different pwsh sessions for the history file.

The `FNV-1a` hashing algorithm is used (32-bit) to provide a consistent hash code across different pwsh sessions.

### DIFF
--- a/PSReadLine/History.cs
+++ b/PSReadLine/History.cs
@@ -16,61 +16,27 @@ using Microsoft.PowerShell.PSReadLine;
 namespace Microsoft.PowerShell
 {
     /// <summary>
-    /// A simple implementation of CRC32.
-    /// See "CRC-32 algorithm" in https://en.wikipedia.org/wiki/Cyclic_redundancy_check.
+    /// FNV-1a hashing algorithm: http://www.isthe.com/chongo/tech/comp/fnv/#FNV-1a
     /// </summary>
-    internal class CRC32Hash
+    internal class FNV1a32Hash
     {
-        // CRC-32C polynomial representations
-        private const uint polynomial = 0x1EDC6F41;
-        private static uint[] table;
+        // FNV-1a algorithm parameters: http://www.isthe.com/chongo/tech/comp/fnv/#FNV-param
+        private const uint FNV32_PRIME = 16777619;
+        private const uint FNV32_OFFSETBASIS = 2166136261;
 
-        static CRC32Hash()
+        internal static uint ComputeHash(byte[] buffer)
         {
-            uint temp = 0;
-            table = new uint[256];
-
-            for (int i = 0; i < table.Length; i++)
+            uint hash = FNV32_OFFSETBASIS;
+            for (int i = 0; i < buffer.Length; i++)
             {
-                temp = (uint)i;
-                for (int j = 0; j < 8; j++)
-                {
-                    if ((temp & 1) == 1)
-                    {
-                        temp = (temp >> 1) ^ polynomial;
-                    }
-                    else
-                    {
-                        temp >>= 1;
-                    }
-                }
-
-                table[i] = temp;
+                hash = (hash ^ buffer[i]) * FNV32_PRIME;
             }
+            return hash;
         }
 
-        private static uint Compute(byte[] buffer)
+        internal static uint ComputeHash(string input)
         {
-            uint crc = 0xFFFFFFFF;
-            for (int i = 0; i < buffer.Length; ++i)
-            {
-                var index = (byte)(crc ^ buffer[i] & 0xff);
-                crc = (crc >> 8) ^ table[index];
-            }
-
-            return ~crc;
-        }
-
-        internal static byte[] ComputeHash(byte[] buffer)
-        {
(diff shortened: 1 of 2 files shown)


lzybkr:
Rather than hashing, we might as well just use the filename directly. There is a limit (260) in the mutex name, so if the name is too long, I'd drop the leading part of the filename, it's likely not unique. For example, `c:\Users\jason\AppData\` seems perfectly safe to drop and have reasonable uniqueness.

daxian-dbw:
Backslash `\` is a reserved character for mutex name, so using file name means we need extra work to normalize the path separator.
What is the disadvantage of using hashing? It's a simple hashing algorithm. There is possibility of collision, but should be rare.

lzybkr:
CRC32 is actually pretty bad, I've seen collisions in practice.

But I'm mostly objecting to needing to copy the code and hoping for fewer lines of code.

There is also a benefit to using the filename - it should be more obvious what the mutex is protecting if you were using low level tools like the module `NtObjectManager`:

```
#137 PS> Get-NtMutant PSReadLineHistoryFile_1095088493 | fl *

Owner                 : PID: 0 - TID: 0
CurrentCount          : 1
OwnedByCaller         : False
AbandonedState        : False
GrantedAccess         : QueryState, Delete, ReadControl, WriteDac, WriteOwner, Synchronize
FullPath              : \Sessions\1\BaseNamedObjects\PSReadLineHistoryFile_1095088493
GrantedAccessMask     : 2031617
SecurityDescriptor    : O:BAG:DUD:(A;;0x1f0001;;;SY)(A;;0x1f0001;;;BA)(A;;0x1f0001;;;S-1-5-5-0-16814681)
Sddl                  : O:BAG:DUD:(A;;0x1f0001;;;SY)(A;;0x1f0001;;;BA)(A;;0x1f0001;;;S-1-5-5-0-16814681)
Handle                : 0x108C
NtTypeName            : Mutant
NtType                : Name = Mutant - Index = 17
Name                  : PSReadLineHistoryFile_1095088493
CanSynchronize        : True
CreationTime          : 12/31/1600 4:00:00 PM
AttributesFlags       : None
HandleReferenceCount  : 2
PointerReferenceCount : 65491
Inherit               : False
ProtectFromClose      : False
Address               : 0
IsContainer           : False
```

daxian-dbw:
OK, got your concern. I will update the PR to not rely on the CRC32 hashing function.

lzybkr:
If you did want to stick with hashing, I'd use something like [fnv-1a](http://www.isthe.com/chongo/tech/comp/fnv/#FNV-1a) instead, it's very little code.

daxian-dbw:
Yes, personally I still want to use a hash algorithm for the unique name purpose. I will use `fnv-1a` instead.

rkeithhill:
I suppose using a generated GUID would be too long?

daxian-dbw:
The GUID needs to be generated based on the history file path, so that when the path is changed in a pwsh session, a different mutex name is used to guard that file.
Generating a GUID out of a file name is basically hashing again.

daxian-dbw:
@lzybkr The PR has been updated to use the FNV-1a hashing algorithm. PR description is also updated. Thanks for the suggestion!

SteveL-MSFT (approved):
LGTM

PSReadLine/History.cs
```diff
@@ -146,7 +171,7 @@ private string GetHistorySaveFileMutexName()
         {
             // Return a reasonably unique name - it's not too important as there will rarely
             // be any contention.
-            return "PSReadLineHistoryFile_" + _options.HistorySavePath.GetHashCode();
+            return "PSReadLineHistoryFile_" + FNV1a32Hash.ComputeHash(_options.HistorySavePath);
```
lzybkr: I was going to suggest avoiding the conversion to byte array (can we use a span somehow?) but then realized we probably should be normalizing the name anyway, e.g. ensuring consistent case on Windows.

daxian-dbw: > we probably should be normalizing the name anyway

Agreed. Update `Set-PSReadLineOption` to sort of resolve (not really) PSPath. Also call `ToLower` before sending the file path to hash method.

> avoiding the conversion to byte array (can we use a span somehow?)

This is a fun exercise!
We cannot use span because we need to compile it against .NET 4.6.1, but we don't really need to use a span here.
The update `ComputeHash(string input)` behaves exactly the same as the `ComputeHash(string input)` implementation in the following code
```
internal static uint ComputeHash(byte[] buffer)
{
    uint hash = FNV32_OFFSETBASIS;
    for (int i = 0; i < buffer.Length; i++)
    {
        hash = (hash ^ buffer[i]) * FNV32_PRIME;
    }
    return hash;
}

internal static uint ComputeHash(string input)
{
    return ComputeHash(Encoding.Unicode.GetBytes(input));
}
```

lzybkr (approved):

PR #1068, opened 2019-09-24
Handle 'PSObject' correctly so that script block handlers works for 'AddToHistoryHandler'

### LINKED ISSUE #1067: Prompt unresponsive with custom AddHistoryHandler since commit #1058
Environment data
----------------
PS version: 5.1.18362.145
PSReadline version: 2.0.0-beta5
PS file version: 10.0.18362.1 (WinBuild.160101.0800)
BufferWidth: 120
BufferHeight: 9001

**And:**

PS version: 7.0.0-preview.4
PSReadline version: 2.0.0-beta5
PS file version: 7.0.0.0
BufferWidth: 120
BufferHeight: 9001

@daxian-dbw:
1. Add custom PSReadLine history handler to PowerShell Profile:
```powershell
Get-Content $profile
#The only thing in profile is this:
$Parameters = @{
    AddToHistoryHandler = {
        param($line)
        $line -notmatch '^\s+|ToSecureString|AsPlainText'
    }
}
Set-PSReadLineOption @Parameters
```
2. Restart PowerShell session and start typing as soon as prompt is ready. The prompt is unresponsive during first few keystrokes and then outputs all typed keystrokes at once.
3. The workaround is to return new *AddToHistoryOption* enumerator type:
```powershell
Get-Content $profile
#The only thing in profile is this:
$Parameters = @{
    AddToHistoryHandler = {
        param($line)
        if ($line -notmatch '^\s+|ToSecureString|AsPlainText') {
            [Microsoft.PowerShell.AddToHistoryOption]::MemoryAndFile
        } else {
            [Microsoft.PowerShell.AddToHistoryOption]::SkipAdding
        }
    }
}
Set-PSReadLineOption @Parameters
```
It looks like there is issue with PSReadLine code handling returned *Boolean* (AKA old behavior)?

daxian-dbw:
Fix #1067

After changing the handler type to be `Func<string, object>`, PowerShell returns `PSObject` directly that wraps the true return value, which was not correctly handled.
The changes are:
- Unwrap the 'PSObject' returned from the handler call.
- Move `LanguagePrimitives.TryConvertTo` to the end, so we don't call it for Boolean values returned from the handler.
- Add a test to cover the script block handler.

### DIFF
--- a/PSReadLine/History.cs
+++ b/PSReadLine/History.cs
@@ -137,15 +137,27 @@ namespace Microsoft.PowerShell
                 }
 
                 object value = Options.AddToHistoryHandler(line);
+                if (value is PSObject psObj)
+                {
+                    value = psObj.BaseObject;
+                }
 
-                if (LanguagePrimitives.TryConvertTo(value, out AddToHistoryOption enumValue))
+                if (value is bool boolValue)
+                {
+                    return boolValue ? AddToHistoryOption.MemoryAndFile : AddToHistoryOption.SkipAdding;
+                }
+
+                if (value is AddToHistoryOption enumValue)
                 {
                     return enumValue;
                 }
 
-                if (value is bool boolValue && !boolValue)
+                // 'TryConvertTo' incurs exception handling when the value cannot be converted to the target type.
+                // It's expensive, especially when we need to process lots of history items from file during the
+                // initialization. So do the conversion as the last resort.
+                if (LanguagePrimitives.TryConvertTo(value, out enumValue))
                 {
-                    return AddToHistoryOption.SkipAdding;
+                    return enumValue;
                 }
             }
 
--- a/PSReadLine/History.cs
+++ b/PSReadLine/History.cs
@@ -152,6 +152,11 @@ namespace Microsoft.PowerShell
                     return enumValue;
                 }
 
+                if (value is string strValue && Enum.TryParse(strValue, out enumValue))
+                {
+                    return enumValue;
+                }
+
                 // 'TryConvertTo' incurs exception handling when the value cannot be converted to the target type.
                 // It's expensive, especially when we need to process lots of history items from file during the
                 // initialization. So do the conversion as the last resort.
--- a/test/HistoryTest.cs
+++ b/test/HistoryTest.cs
@@ -158,9 +158,9 @@ namespace Test
                 options.HistorySavePath = newHistoryFilePath;
                 options.HistorySaveStyle = newHistorySaveStyle;
 
-                /*
-                 * Set null to the handler means we don't do the check.
-                 */
+                //
+                // Set null to the handler means we don't do the check.
+                //
                 options.AddToHistoryHandler = null;
                 SetHistory(commandInputs);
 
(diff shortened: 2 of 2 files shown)


test/HistoryTest.cs
```diff
@@ -118,136 +119,289 @@ public void SensitiveHistoryOptionalBehavior()
 
             var options = PSConsoleReadLine.GetOptions();
             var oldHistoryFilePath = options.HistorySavePath;
-            var oldHistorySaveStyle = options.HistorySaveStyle;
-
+            var oldHistorySaveStyle = options.HistorySaveStyle;
```
lzybkr, line 122: Did the newlines change?

test/HistoryTest.cs
```diff
@@ -118,136 +119,289 @@ public void SensitiveHistoryOptionalBehavior()
...
+                ScriptBlock.Create(@"
+                    param([string]$line)
+                    if ($line.Contains('gal')) {
+                        [Microsoft.PowerShell.AddToHistoryOption]::MemoryOnly
+                    } elseif ($line.Contains('gmo')) {
+                        'SkipAdding'
+                    } else {
+                        [Microsoft.PowerShell.AddToHistoryOption]::MemoryAndFile
+                    }"));
+            Func<string, object> newAddToHistoryHandler_ReturnOther = LanguagePrimitives.ConvertTo<Func<string, object>>(
```
lzybkr, line 286: I don't see an explicit test for `PSObject` return value. Maybe the result is always wrapped in `PSObject`, but it still seems wise to add an explicit test for that case on the off chance that the implementation changes to bypass the `PSObject` wrapper.

test/HistoryTest.cs
```diff
@@ -118,136 +119,289 @@ public void SensitiveHistoryOptionalBehavior()
...
+                "gcm c*",
+                "ConvertTo-SecureString -AsPlainText -String abc -Force"
+            };
+
+            try
+            {
+                options.HistorySavePath = newHistoryFilePath;
+                options.HistorySaveStyle = newHistorySaveStyle;
+
+                /*
```
lzybkr: This project uses `//` for multi-line comments exclusively (modulo the copyright notices.)

test/HistoryTest.cs
daxian-dbw: Yes, changed from `CRLF` to `LF`.

test/HistoryTest.cs
daxian-dbw: Sure, will add an explicit test for `PSObject`.
I'm planning to change the implementation to not always wrap the result. It also seems to be a code reusing issue.

test/HistoryTest.cs
daxian-dbw: Will change to use `//`.

test/HistoryTest.cs
daxian-dbw: https://github.com/PowerShell/PSReadLine/pull/1068/files?w=1 could be helpful

daxian-dbw:
@lzybkr You comments are addressed. Can you please take another look?

PSReadLine/History.cs
```diff
@@ -137,15 +137,27 @@ private AddToHistoryOption GetAddToHistoryOption(string line)
...
+                }
 
-                if (LanguagePrimitives.TryConvertTo(value, out AddToHistoryOption enumValue))
+                if (value is AddToHistoryOption enumValue)
                 {
                     return enumValue;
                 }
 
-                if (value is bool boolValue && !boolValue)
+                // 'TryConvertTo' incurs exception handling when the value cannot be converted to the target type.
```
lzybkr, line 160: I'll bet some people will return strings that convert to `AddToHistoryOption`. You could make those conversions cheaper with a tiny bit of special case conversion code here.

daxian-dbw: Agreed. The special case for string conversion is added.

lzybkr (approved):

PR #1097, opened 2019-10-12
Fix for a nullref thrown when PSES calls ForcePSEventHandling

TylerLeonhardt:
In PowerShell Editor Services, we have a private contract with PSRL with a method called `ForcePSEventHandling` found in:

* [PSRL](https://github.com/PowerShell/PSReadLine/blob/master/PSReadLine/ReadLine.cs#L827-L836)
* [PSES](https://github.com/PowerShell/PowerShellEditorServices/blob/c744e79525566ef17c86d51c601e988188bbc9e4/src/PowerShellEditorServices/Services/PowerShellContext/Session/PSReadLinePromptContext.cs#L199-L202)

Due to the multi-threaded nature of PSES, we are calling `ForcePSEventHanding` before the `_forceEventWaitHandle` was initialized (which happens in `DelayedOneTimeInitialize(`) thus causing a null reference exception.

This PR moves the initialization to the field declaration which fixes the issue we were seeing (PSES crashing).

This should be low risk as this is only used by PSES and they are a private contract.

### DIFF
--- a/PSReadLine/ReadLine.cs
+++ b/PSReadLine/ReadLine.cs
@@ -52,7 +52,7 @@ namespace Microsoft.PowerShell
         private Thread _readKeyThread;
         private AutoResetEvent _readKeyWaitHandle;
         private AutoResetEvent _keyReadWaitHandle;
-        private AutoResetEvent _forceEventWaitHandle;
+        private AutoResetEvent _forceEventWaitHandle = new AutoResetEvent(false);
         private CancellationToken _cancelReadCancellationToken;
         internal ManualResetEvent _closingWaitHandle;
         private WaitHandle[] _threadProcWaitHandles;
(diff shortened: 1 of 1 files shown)


TylerLeonhardt:
There are some greater questions that can be asked on the overall design of this contract, but this at least mitigates the issue for now.

cc @SeeminglyScience

daxian-dbw:
Would changing `ForcePSEventHandling` to `_singleton?._forceEventWaitHandle?.Set();` solve the PSES problem too? (syntax needs to change to compile with 4.6.1 though)

TylerLeonhardt:
Thought about that but we want that `_forceEventWaitHandle` to be set when we call the method. It probably doesn't make a huge difference, but this PR keeps the idea consistent of "when this method is called, it sets this `AutoResetEvent`" rather than it being best effort.

daxian-dbw (approved):
I'm fine with this change.

TylerLeonhardt:
I consulted heavily with @SeeminglyScience on this - who originally did the ground work. I'd like to give him a chance to sign off so hold off on merging until he gives a thumbs up.

SeeminglyScience (approved):
LGTM!

TylerLeonhardt:
I don't see a problem with that, @daxian-dbw 👍 LGTM

PR #1120, opened 2019-10-24
In VI mode, moving left | right should stick cursor on logical line

### LINKED ISSUE #1119: In VI mode, the cursor should stick to the current logical line when moving right or left
The current behaviour of PSRL is to treat the entire buffer as a long line.

Therefore, when moving right past the end of a logical line, the cursor jumps to the beginning of the next line. Conversely, when moving left at the beginning of a logical line, the cursor jumps at the end of the previous logical line.

In VI mode, the cursor should stick to the current logical line.

springcomp:
Fixes #1119.

Using specific implementations of `ViBackwardChar` and `ViForwardChar`.

This also fixes an issue, in insert mode, pressing <kbd>Enter</kbd> creates a new line. Then pressing <kbd>Esc</kbd> moved the cursor at the end of the previous logical line. Now, pressing <kbd>Esc</kbd> leaves the cursor on the new (empty) line.


test/MovementTest.VI.Multiline.cs
```diff
@@ -5,6 +5,50 @@ namespace Test
...
+                "line2", _.Enter,
+                "line3", _.Enter,
+                _.DQuote,
+                _.Escape,
+                _.k, _.k, CheckThat(() => AssertCursorLeftIs(continuationPrefixLength + 0)),
+                // move right
+                _.l, CheckThat(() => AssertCursorLeftIs(continuationPrefixLength + 1)),
+                "10l", CheckThat(() => AssertCursorLeftIs(continuationPrefixLength + 4))
+            ));
+        }
```
msftrncs, line 51: I think there is to be an empty line here, before next test.

springcomp: Sure, I will add an extra line.

lzybkr (approved):

PR #1122, opened 2019-10-24
Ensure the desired column number is used while moving up or down in VI mode

### LINKED ISSUE #1121: In VI mode, the cursor does not always stick to the desired column number
The current behaviour of PSRL is as follows:

When in a continuous sequence of moving the cursor up or down, the cursor position in the current logical line is taken as the *desired column number* for the target logical line. At one point in the sequence of moves, if the cursor is located at the end of a logical line, all subsequent moves will place the cursor at the end of the logical lines. Additionnaly, the cursor is sometimes placed past the last non-blank character of a logical line, which is inconsistent with how it should behave in vi or vim.

In VI mode, the desired behaviour is as follows:

- The *desired column number* is either a physical column number, the logical 'end of line' position, or 'not specified'.

- When pressing <kbd>$</kbd>, the desired column number is explicitely set to the logical 'end of line' position. Therefore, subsequently moving the cursor up or down should place the cursor at the end of the target logical lines. 

- When performing a sequence of moves involving <kbd>j</kbd> or <kbd>k</kbd>:
  - If the desired column number is 'not specified', its value is taken from the position of the cursor on the logical line when performing the first move. Even if this position is the end of the logical line, the *desired column number* is still a known offset from the beginning of the logical line. 
  - When moving up  or down, the cursor is placed at the desired column number if possible, or at the end of the logical line. 

- When performing any other command, the *desired column number* is reset to 'not specified'.

springcomp:
Fixes #1121.

### DIFF
--- a/PSReadLine/Position.cs
+++ b/PSReadLine/Position.cs
@@ -36,6 +36,41 @@ namespace Microsoft.PowerShell
         }
 
         /// <summary>
+        /// Returns the position of the beginning of line
+        /// for the 0-based specified line number.
+        /// </summary>
+        private static int GetBeginningOfNthLinePos(int lineIndex)
+        {
+            System.Diagnostics.Debug.Assert(lineIndex >= 0 || lineIndex < _singleton.GetLogicalLineCount());
+
+            var nth = 0;
+            var index = 0;
+            var result = 0;
+
+            for (; index < _singleton._buffer.Length; index++)
+            {
+                if (nth == lineIndex)
+                {
+                    result = index;
+                    break;
+                }
+
+                if (_singleton._buffer[index] == '\n')
+                {
+                    nth++;
+                }
+            }
+
+            if (nth == lineIndex)
+            {
+                result = index;
+            }
+
+
+            return result;
+        }
+
+        /// <summary>
         /// Returns the position of the end of the logical line
         /// as specified by the "current" position.
         /// </summary>
--- a/PSReadLine/ReadLine.cs
+++ b/PSReadLine/ReadLine.cs
@@ -473,7 +473,8 @@ namespace Microsoft.PowerShell
                 var recallHistoryCommandCount = _recallHistoryCommandCount;
                 var yankLastArgCommandCount = _yankLastArgCommandCount;
                 var visualSelectionCommandCount = _visualSelectionCommandCount;
-                var movingAtEndOfLineCount = _moveToLineCommandCount;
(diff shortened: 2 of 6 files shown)


PSReadLine/ReadLine.cs
```diff
@@ -534,9 +535,21 @@ private string InputLoop()
...
                 }
-                if (movingAtEndOfLineCount == _moveToLineCommandCount)
+                if (moveToLineCommandCount == _moveToLineCommandCount)
                 {
-                    _moveToLineCommandCount = 0;
+                    _moveToLineCommandCount = 0;
+
+                    if (moveToEndOfLineCommandCount == _moveToEndOfLineCommandCount)
+                    {
+                        if (InViCommandMode())
```
daxian-dbw: Suggest to change to
```c#
if (InViCommandMode() && moveToEndOfLineCommandCount == _moveToEndOfLineCommandCount)
{
    ....
}
```

PSReadLine/Position.cs
```diff
@@ -35,6 +35,41 @@ private static int GetBeginningOfLinePos(int current)
...
+                if (nth == lineIndex)
+                {
+                    result = index;
+                    break;
+                }
+
+                if (_singleton._buffer[index] == '\n')
+                {
+                    nth++;
+                }
```
daxian-dbw, line 61: There might be an issue in a corner case: when the last character in the buffer is `\n`, execution falls in this if block and let's say `nth` becomes `10` after the increment, then `index++` in the `for` loop executes and `index` becomes `_buffer.Length`. If `lineIndex` happens to be `10`, then the result will be set to `_buffer.Length` and gets returned. Would that cause a problem?

daxian-dbw (commented):
Sorry that it takes forever to review this PR. It overall looks good to me. I left 2 comments, please take a look when you get a chance.

PSReadLine/Position.cs
springcomp: No, it does not cause any problem.

The offset that corresponds to the length of the string is a valid cursor position if the last character of the buffer is `\n`. That is, the cursor is on a new empty line, at a position that does not correspond to a physical caracter yet.

PSReadLine/ReadLine.cs
springcomp: Done, as suggested.

daxian-dbw (approved):

PR #1124, opened 2019-10-24
Return error if color property or value is invalid with Set-PSReadLineOption -Colors

### LINKED ISSUE #1118: Bad error message
Set-PSReadLineOption -Colors @{default='apple'}

Exception message is: o

SteveL-MSFT:
Add descriptive error messages if the color property or value is invalid.  Had to rerun resgen regenerating the resources cs file.

Fix https://github.com/PowerShell/PSReadLine/issues/1118

### DIFF
--- a/PSReadLine/Cmdlets.cs
+++ b/PSReadLine/Cmdlets.cs
@@ -483,6 +483,10 @@ namespace Microsoft.PowerShell
             {
                 setter(this, value);
             }
+            else
+            {
+                throw new ArgumentException(string.Format(CultureInfo.CurrentUICulture, PSReadLineResources.InvalidColorProperty, property));
+            }
         }
     }
 
@@ -964,7 +968,7 @@ namespace Microsoft.PowerShell
                     break;
             }
 
-            throw new ArgumentException("o");
+            throw new ArgumentException(string.Format(CultureInfo.CurrentUICulture, PSReadLineResources.InvalidColorValue, o.ToString()));
         }
 
         public static string AsEscapeSequence(ConsoleColor fg, ConsoleColor bg)
--- a/PSReadLine/PSReadLineResources.Designer.cs
+++ b/PSReadLine/PSReadLineResources.Designer.cs
@@ -19,7 +19,7 @@ namespace Microsoft.PowerShell.PSReadLine {
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "15.0.0.0")]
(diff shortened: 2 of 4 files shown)


lzybkr (changes requested):
`PSReadLineResources.Designer.cs` seems to have changed every line - probably a CRLF => LF change?

SteveL-MSFT:
@lzybkr ran resgen.exe on Windows.  VSCode probably changed it.  I can change it back to CRLF.

SteveL-MSFT:
Actually, resgen creates a file with CRLF.  Resaved as LF only.  VSCode cleaned up trailing whitespace.

SteveL-MSFT:
@lzybkr can you re-review?  Thanks.

lzybkr (dismissed):
Approved, assuming resgen removed the trailing whitespace.
If the trailing whitespace was removed by an editor, then it's probably better to keep it or somebody else will just end up adding it back in another PR.

lzybkr (dismissed):
Since resgen is apparently generating trailing whitespace, we should remove it as future changes will likely add it back.

SteveL-MSFT:
@lzybkr it seems that resgen has some hardcoded whitespace and also generates CRLF by default.  git saves files as LF.  In the future, we should probably have resgen as part of the build script if resources change.  I don't see a need value in putting in the "raw" resgen'd file unless we also want the CRLF line endings.

lzybkr:
Maybe it's best to run resgen as part of the build and remove the generated file. It should be much simpler now than when this project was started.

SteveL-MSFT:
@lzybkr if we depend on resgen.exe, then builds only work on Windows again

lzybkr:
I guess I'm not following what you're suggesting then. But I think my point still stands - we don't want unnecessary changes to this file as it is generated - those changes are too easily undone and make reviews harder than necessary.

`git` will do what you tell it to w.r.t. eol, so we could add a `.gitattributes` file with a line like:

```
PSReadLine/PSReadLineResources.Designer.cs eol=crlf
```

And I think that wouldn't change what's in the index (which is `lf` for this file, but `crlf` and even `mixed` for other files in this repo).

SteveL-MSFT:
@lzybkr the file in git is currently LF which is why I had to resave it from CRLF to just LF so that the entire file didn't show up as modified.   However, resgen (being a Windows tool) emits CRLF by default.  So at some point in the past that file was already resaved as just LF.  My suggestion is to defer fixing the resgen issue to later if we have to make more changes to the resx.  Perhaps publish the resgen we have in PS repo and use that.

daxian-dbw:
@SteveL-MSFT after reading the conversations here, it looks to me what @lzybkr asked was to keep the trailing spaces in the generated file.
If you don't mind, I can re-generate the strongly-typed resource class file, change `CRLF` to `LF` and keep the trailing spaces.

daxian-dbw (approved):

Issue #1083, opened 2019-10-03
bck-i-search (Ctrl-R) quits accepting input when it finds a very long command


Narvey:
When I press Ctrl-R, I keep having this gigantic line I once pasted in pop up in the history and clobber the bck-i-search. All the keyboard input then goes at the end of the giant command and so I can't finish typing into bck-i-search the rest of what I wanted to search for. Worse still, Ctrl-C does not get me out, and Ctrl-D just puts ^D on the end of the command. Up arrow does get me out, but causes the below exception. 

Environment data
----------------

PS version: 5.1.18362.145
PSReadline version: 2.0.0-beta2
os: 10.0.18362.1 (WinBuild.160101.0800)
PS file version: 10.0.18362.1 (WinBuild.160101.0800)
BufferWidth: 120
BufferHeight: 3000

Steps to reproduce or exception report
-----------------------------------------------------------------------
Last 4 Keys:
 Ctrl+r b a UpArrow

Exception:
System.ArgumentOutOfRangeException: The value must be greater than or equal to zero and less than the console's buffer size in that dimension.
Parameter name: top
Actual value was -4965.
   at System.Console.SetCursorPosition(Int32 left, Int32 top)
   at Microsoft.PowerShell.PSConsoleReadLine.ReallyRender(RenderData renderData, String defaultColor)
   at Microsoft.PowerShell.PSConsoleReadLine.ForceRender()
   at Microsoft.PowerShell.PSConsoleReadLine.HistoryRecall(Int32 direction)
   at Microsoft.PowerShell.PSConsoleReadLine.PreviousHistory(Nullable`1 key, Object arg)
   at Microsoft.PowerShell.PSConsoleReadLine.ProcessOneKey(ConsoleKeyInfo key, Dictionary`2 dispatchTable, Boolean ignoreIfNoAction, Object arg)
   at Microsoft.PowerShell.PSConsoleReadLine.InputLoop()
   at Microsoft.PowerShell.PSConsoleReadLine.ReadLine(Runspace runspace, EngineIntrinsics engineIntrinsics)

-----------------------------------------------------------------------

msftrncs:
@Narvey, you should give PSReadline 2.0.0-beta5 a try, from PowerShellGallery.  I cannot say that it will help or entirely eliminate the issue, but a better error report may be possible, knowing how the latest version handles.

daxian-dbw:
[...]

PR #1141, opened 2019-11-01
Don't SetCursorPosition when not Rendered.

### LINKED ISSUE #1081: Copying and pasting a comment raised an exception, until I ran Clear-Host
```
 A s s i g n Spacebar a l e r t s @ C L I E N T D O M A I N Spacebar t o Spacebar t h i s . Enter
 Ctrl+Enter ( G e t - A c c e p t e d D o m a i n Spacebar | Spacebar S e l e c t Spacebar - E x p a n d P r o p e r t y Spacebar D o m a i n N a m e ) Spacebar | Spacebar F o r E a c h - O b j e c t Spacebar { Enter
 Ctrl+Enter Spacebar Spacebar Spacebar Spacebar S e t - M a i l C o n t a c t Spacebar " a l e r t s " Spacebar - E m a i l A d d r e s s e s Spacebar @ { A d d = " a l e r t s @ $ _ " } Spacebar - E r r o r A c t i o n Spacebar S t o p Enter
 Ctrl+Enter } Enter
 Ctrl+Enter
```
>
> ### Exception
```
System.ArgumentOutOfRangeException: The value must be greater than or equal to zero and less than the console's buffer size in that dimension. (Parameter 'top')
Actual value was 63.
   at System.ConsolePal.SetCursorPosition(Int32 left, Int32 top)
   at Microsoft.PowerShell.PSConsoleReadLine.MoveCursor(Int32 newCursor)
   at Microsoft.PowerShell.PSConsoleReadLine.MoveToLine(Int32 numericArg)
   at Microsoft.PowerShell.PSConsoleReadLine.PreviousLine(Nullable`1 key, Object arg)
   at Microsoft.PowerShell.PSConsoleReadLine.InsertLineAbove(Nullable`1 key, Object arg)
```

Environment data
----------------
````powershell
PS version: 7.0.0-preview.3
PS file version: 7.0.0.0
BufferWidth: 254
BufferHeight: 63
````

Get a nice, full scrollback buffer.  Then, paste the text `### Assign alerts@CLIENTDOMAIN to this.` into PowerShell 7.0.0-preview3.  This one block of code always causes the same problem:
````powershell
### Assign alerts@CLIENTDOMAIN to this.
(Get-AcceptedDomain | Select -ExpandProperty DomainName) | ForEach-Object {
    Set-MailContact "alerts" -EmailAddresses @{Add="alerts@$_"} -ErrorAction Stop
}
````
There are some more lines of code before that that all ran fine.  However, that comment line caused PSReadLine to fail.  The code afterwards ran just fine.

Weirdly enough, after running `Clear-Host`, everything appeared to be fine.  I could paste that in all day.

This might (or might not) be a bug in Windows Terminal.  It appeared to work fine when I opened PowerShell 7 with `pwsh-preview`.

### PR DESCRIPTION
**Pull request by msftrncs**
Block the setting of the console cursor when the buffer hasn't been completely rendered due to an input queue waiting to be processed.

Context: https://github.com/PowerShell/PSReadLine/issues/1081#issuecomment-548679030

Fixes #1081.

### DIFF
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -61,6 +61,7 @@ namespace Microsoft.PowerShell
         };
         private int _initialX;
         private int _initialY;
+        private bool _waitingToRender;
 
         private ConsoleColor _initialForeground;
         private ConsoleColor _initialBackground;
@@ -112,6 +113,7 @@ namespace Microsoft.PowerShell
                 _tokens = null;
                 _ast = null;
                 _parseErrors = null;
+                _waitingToRender = true;
                 return;
             }
 
@@ -816,6 +818,7 @@ namespace Microsoft.PowerShell
             // TODO: set WindowTop if necessary
 
             _lastRenderTime.Restart();
+            _waitingToRender = false;
         }
 
         private static string Spaces(int cnt)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by lzybkr** [comment]
Nice job tracking down this issue.

It sounds like if we detect a simulated paste and convert that to a real paste (#579), we could avoid processing any keys that might move the cursor before rendering.

Then again, a fix for #579 might not be 100% reliable and this fix is still necessary, it's hard to say for sure.

**Comment by msftrncs** [comment]
Also, I think Windows Terminal plans to fix the paste behavior regarding the CTRL-ENTER being pasted (there is a PR for it) and with that, the issue this fixes becomes hard to reproduce again.

**Comment by msftrncs** [comment]
I am still getting another exception signature if I find the right material to paste, and I think that is related to #1083.   It involves when the cursor ends up going off the top of the buffer, in this case, during the paste due to the excessive CTRL-ENTER's filling up the bottom of the buffer.  To fix that, would really require being able to render a partial buffer other than just at the end.

File: PSReadLine/Render.cs [hunk 1]
```diff
@@ -980,19 +983,24 @@ private void RecomputeInitialCoords()
...
+                {
+                    Ding();
+                    return;
+                }
+
+                _console.SetCursorPosition(point.X, point.Y);
             }
 
-            _console.SetCursorPosition(point.X, point.Y);
             _current = newCursor;
```
daxian-dbw, line 1024: This should be moved in the `if (!_waitingToRender)` block too, right?

msftrncs: You mean the `_current = newCursor;` part?

No.  The insertion cursor needs to be updated, the physical cursor cannot be updated because the render on screen doesn't match what is going on in the buffer.  Not updating the insertion cursor would cause the paste operation to not match the exception report buffer if and when a exception was to occur.  The paste would also definitely result differently, though this would prevent the exceptions related to #1083 during the paste.

This whole issue however, would be much more difficult to reproduce if Terminal wasn't stuffing the `\n` during the paste, but I would think there is always the possibility of this occuring during laggy remote sessions.  I understand the next preview of Terminal will have this fixed.

msftrncs:
Noting a merge conflict will occur with #1146.

PSReadLine/Render.cs
daxian-dbw: In case `newCursor` results in `point.Y < 0`, we don't set `newCursor` to `_current`. But after this change, if that happens when `_waitingToRender == true`, `_current` will be set to `newCursor` unexpectedly.

msftrncs: while `_waitingToRender == false` the state of the render is known, and its possible to compute the coords and determine that the physical cursor will be outside the buffer, and then refuse the move request.

While `_waitingToRender == true`, the screen buffer is unknown, and the assumption is that we are receiving a burst of keyboard input and just trying to get it in to the command line buffer, we need to be able to process the current requests of the keys, hoping that when it is time to render, the cursor will be some place acceptable.   If when the render occurs, the cursor will be outside the screen buffer (most likely `point.Y < 0`), then an exception will still occur, but that is basically the same exception as #1083.

In the future, I am hoping that PSReadLine will be able to edit command lines that are larger than the buffer, in which case the test for `point.Y < 0` will execute a re-render to get the cursor back within the screen buffer.  This is the real fix for #1083.

daxian-dbw: I see your point. when rendering is pending, the `_initialX` and `_initialY` are not updated yet, and thus there is no guarantee that `ConvertOffsetToPoint` would produce the right cursor position. Hence, checking `point.Y < 0` is meaningless.
Can you please add some comments above the line `_current = newCursor;` to explain why we are not converting offset to point and check for if it's off the buffer? Also, please resolve the conflict. Thanks!

msftrncs: Merge conflict resolved and comment in place.

daxian-dbw (approved):

PR #1146, opened 2019-11-02
Handle cursor being moved to end of buffer.

### LINKED ISSUE #1144: Moving cursor to end of buffer causes exception
Environment data
----------------
```
PS version: 5.1.19013.1
PSReadline version: 2.0.0-beta5
os: 10.0.19013.1 (WinBuild.160101.0800)
PS file version: 10.0.19013.1 (WinBuild.160101.0800)
BufferWidth: 120
BufferHeight: 28
```

Similar to #884, and is easiest in a limited buffer size terminal, such as the current Windows Terminal or VS Code's terminal, enter a long line of text on the last line of the buffer, but before filling it, arrow back one character, and then extend the line to the end of the buffer.

image

Then press <kbd>END</kbd>, and an exception will occur.   I'm pretty sure it occurs because the desired cursor position is calculated to be the beginning of the next line, but that is beyond the end of the buffer, and no rendering event has occurred to be able to fix that.

### Exception
```
Parameter name: top
Actual value was 28.
   at Microsoft.PowerShell.PSConsoleReadLine.MoveCursor(Int32 newCursor)
```

`ReallyRender()` handles landing in this position by forcing a newline, but `MoveCursor()` doesn't have any handling for that.

I was able to replicate this issue in ConsoleHost, if I filled up the 9000 line buffer first.

Interestingly, it appears that most these terminals support emitting text in to the last character of the last line without the cursor bumping up the lines.  However, the SetCursorPosition, doesn't seem to allow moving the cursor in to this condition.  ~Evidently VS Code's terminal does not support this~ (my error while testing), even though its a well known ANSI terminal feature from more than 20 years ago.  It allows you to write the very last character cell on the screen without the screen scrolling.

msftrncs:
Handle when `MoveCursor()` attempts to move the cursor to the end of the terminal buffer, same as `Render()`, by issueing a line feed to force a scroll.

Fixes #1144.

### DIFF
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -992,7 +992,25 @@ namespace Microsoft.PowerShell
                 return;
             }
 
-            _console.SetCursorPosition(point.X, point.Y);
+            if (point.Y == _console.BufferHeight)
+            {
+                // The cursor top exceeds the buffer height, so adjust the initial cursor
+                // position and the to-be-set cursor position for scrolling up the buffer.
+                _initialY -= 1;
+                point.Y -= 1;
+
+                // Insure the cursor is on the last line of the buffer prior
+                // to issuing a newline to scroll the buffer.
+                _console.SetCursorPosition(point.X, point.Y);
+
+                // Scroll up the buffer by 1 line.
+                _console.Write("\n");
+            }
+            else
+            {
+                _console.SetCursorPosition(point.X, point.Y);
+            }
+
             _current = newCursor;
         }
 


PSReadLine/Render.cs
```diff
@@ -992,7 +992,23 @@ private void MoveCursor(int newCursor)
...
+                _console.SetCursorPosition(point.X, point.Y);
+
+                // The cursor top exceeds the buffer height, so we need to
+                // scroll up the buffer by 1 line.
+                _console.Write("\n");
+            }
+            else
+            {
             _console.SetCursorPosition(point.X, point.Y);
+            }
```
daxian-dbw, line 1012: How about changing to the following?

```c#
            if (point.Y == _console.BufferHeight)
            {
                // Adjust the initial cursor position and the to-be-set cursor position
                // after scrolling up the buffer.
                _initialY -= 1;
                point.Y -= 1;

                // The cursor top exceeds the buffer height, so we need to
                // scroll up the buffer by 1 line.
                _console.Write("\n");
            }

            _console.SetCursorPosition(point.X, point.Y);
```

msftrncs: The cursor position may not yet be on the last line, so `console.write("\n")` may not produce the desired result of scrolling the buffer, so it needs to insure the cursor is on the bottom line first.   This can be demonstrated similar as in the reported issue, using a single line that wraps two console lines, pressing <kbd>END</kbd> from the beginning.

However, I definitely missed a format issue there in the commit.

msftrncs:
Noting a merge conflict will occur with #1141.

PSReadLine/Render.cs
daxian-dbw: That make sense, thanks for the clarification! Can you please add a comment explain why `SetCursorPosition` should happen before `_console.Write("\n")`?

msftrncs: I've revised all the comments in this section, as they were no longer in order from where they were copied.  I think this will make things clearer.

daxian-dbw (approved):

PR #1160, opened 2019-11-07
Don't allow InsertLineAbove unless in multiline

SteveL-MSFT:
When at the prompt, if you use the keyhandler (Ctrl+j default in Windows mode), it will insert a line continuation below the prompt and put your cursor back at the prompt.  So it's not inserting above (it can't since you're at the prompt) and if you do this enough to scroll it off the screen, there is an unhandled exception.  Fix is to now allow InsertLineAbove unless you are in a multiline.

### DIFF
--- a/PSReadLine/BasicEditing.cs
+++ b/PSReadLine/BasicEditing.cs
@@ -472,10 +472,13 @@ namespace Microsoft.PowerShell
         /// </summary>
         public static void InsertLineAbove(ConsoleKeyInfo? key = null, object arg = null)
         {
-            // Move the current position to the beginning of the current line and only the current line.
-            _singleton._current = GetBeginningOfLinePos(_singleton._current);
-            Insert('\n');
-            PreviousLine();
+            if (_singleton.LineIsMultiLine())
+            {
+                // Move the current position to the beginning of the current line and only the current line.
+                _singleton._current = GetBeginningOfLinePos(_singleton._current);
+                Insert('\n');
+                PreviousLine();
+            }
         }
 
         /// <summary>


daxian-dbw:
This only happens in Windows Terminal, not the old conhost. With the latter, <kbd>Ctrl+j</kbd> always put the cursor at the new continuation line, and I think that's the right behavior.
~We need to understand why in Windows Terminal the cursor is kept at the prompt line.~ Got it, Windows Terminal and VSCode send <kbd>CTRL+ENTER</kbd> for <kbd>Ctrl+j</kbd>.

daxian-dbw:
I'm able to reproduce this issue using 2.0.0-beta5 shipped with preview.5.
However, I'm not able to reproduce this issue using the PSReadLine built from master. Can you maybe try building from master and see if you still see the exception?

SteveL-MSFT:
With master in WinTerminal and VSCode, the inserted line is below the prompt.  I don't see the exception anymore, but it doesn't seem right to insert below the prompt since this function is InsertLineAbove.  At a prompt, there is no above.

msftrncs:
My interpretation of `InsertLineAbove`

 - Position logical cursor to the beginning of current logical line.
   - Even if the buffer is empty, it has one line.
- Insert a `\n`.
   - At a prompt, this creates a blank line, and line continuation
- Move cursor in front of the new `\n`
   - The `\n` now represents the end of the new line above the original line.

The continuation that appears below the prompt is the line that was originally on the prompt.   It makes more sense after you type something in.

```
Prompt> "hello"_ <CTRL-ENTER>

# result

Prompt> _
>> "hello"
```

SteveL-MSFT:
@msftrncs I see.  If you type content into the prompt then the `insertlineabove` makes more sense.  Ok, closing this PR.

PR #1167, opened 2019-11-09
In Vi mode, 'y0' should only yank up to the start of the logical line.

springcomp:
Currently, <kbd>y0</kbd> yanks to the beginning of the buffer, even when using multiple lines.
This PR fixes this and only yanks text up to the beginning of the current logical line.

### DIFF
--- a/PSReadLine/YankPaste.vi.cs
+++ b/PSReadLine/YankPaste.vi.cs
@@ -252,8 +252,7 @@ namespace Microsoft.PowerShell
             if (length > 0)
             {
                 _clipboard.Record(_singleton._buffer, start, length);
-                _singleton._current = start;
-                _singleton.Render();
+                _singleton.MoveCursor(start);
             }
         }
 
--- a/PSReadLine/YankPaste.vi.cs
+++ b/PSReadLine/YankPaste.vi.cs
@@ -247,10 +247,13 @@ namespace Microsoft.PowerShell
         /// </summary>
         public static void ViYankBeginningOfLine(ConsoleKeyInfo? key = null, object arg = null)
         {
-            int length = _singleton._current;
+            var start = GetBeginningOfLinePos(_singleton._current);
+            var length = _singleton._current - start; 
             if (length > 0)
             {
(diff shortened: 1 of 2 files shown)


test/YankPasteTest.VI.cs
```diff
@@ -324,16 +324,17 @@ public void ViPasteAfterYankBeginningOfLine()
         {
             TestSetup(KeyMode.Vi);
 
-            Test("012", Keys(
```
lzybkr, line 327: Why would you remove the single line test?

springcomp: No particular reason, indeed.
I re-introduced the one-line test, taking into account the new cursor position change.

PSReadLine/YankPaste.vi.cs
```diff
@@ -247,10 +247,13 @@ public static void ViYankEndOfGlob(ConsoleKeyInfo? key = null, object arg = null
...
         {
-            int length = _singleton._current;
+            var start = GetBeginningOfLinePos(_singleton._current);
+            var length = _singleton._current - start; 
             if (length > 0)
             {
-                _singleton.SaveToClipboard(0, length);
+                _clipboard.Record(_singleton._buffer, start, length);
+                _singleton._current = start;
+                _singleton.Render();
```
daxian-dbw: Why do you choose to call `Render` instead of just moving the cursor?

springcomp: I'll have a better look at the `MoveCursor` method.

daxian-dbw: I'm about to publish `2.0.0-beta6` before Friday this week. I want to include this fix if time permits, but no rush, it can always catch the next release.

springcomp: Done.

daxian-dbw (approved):

PSReadLine/YankPaste.vi.cs
daxian-dbw: Thanks @springcomp!

Issue #986, opened 2019-07-31
Coloring is not restored when using PromptText and a custom prompt


JustinGrote:
Environment data
----------------
PS version: 6.2.2                                                                                                       
PSReadline version: 2.0.0-beta4                                                                                         
os: 10.0.18941.1001 (WinBuild.160101.0800)                                                                              PS file version: 6.2.2.0                                                                                                
BufferWidth: 120                                                                                                       
BufferHeight: 30     

Steps to reproduce or exception report
--------------------------------------

PSReadline uses the Error color with -PromptText to change color to show a syntax error, however when this reverts, it seems to reset to default colors rather than restore the prompt as it was previously.

![PSReadLineIssue](https://user-images.githubusercontent.com/15258962/62253096-a6c0aa00-b3a9-11e9-91dc-d57caa90ce96.gif)

Expected Behavior
--------------------------------------
**Ideal**: PSReadline remembers what ANSI codes were used for the characters were that it replaced and respects them. This would probably be a regex of the prompt string to find the last relevant ANSI code combinations.

**Less Ideal**: PSReadline doesn't seem to remember ANSI codes if they are placed in PromptText. If it did, then the coloring expected for the prompt could be defined there and "replayed" when PSReadline reverts from the error coloring.

**Least Ideal**: PSReadline reverts the colors to a defined option such as DefaultTokenColor

lzybkr:
It's not possible for PSReadLine to remember the colors used when originally writing the prompt. The prompt function can write to the console directly and PSReadLine can't scrape the screen to see what was written.

Maybe the right fix isn't to think of changing the prompt color, but instead specifying a error prompt text in the same way that you might specify the prompt text:

```PowerShell
Set-PSReadlineOption -PromptText "`e[44m>`e[0m " -ErrorPromptText "`e[40m>`e[0m "
```
Regardless, I think PSReadLine would need to recognize ANSI escape sequences in the prompt text so it can properly move the cursor back the right number of characters, in this example, 2.

JustinGrote:
@lzybkr I'd be OK with this, I didn't realize prompttext just goes back x amount of spaces and rewrites rather than looking for that specific prompttext.

JustinGrote:
https://github.com/PowerShell/PSReadLine/blob/3ed06e11dc6da4719ac69a756122bd482dc00fe0/PSReadLine/Render.cs#L380

[...]

PR #1180, opened 2019-11-14
Improve handling of color in prompts

lzybkr:
This PR is one possible approach to fixing #986.

It's nice in that it doesn't require adding a new configuration option - I'm thinking this would be used very infrequently.

Here is an example of how you might configure PSReadLine and an example of it working:

```PowerShell
$esc = [char]0x1b
$pc = [char]0xe0b0
$fg = "0"
$bg = "8;2;95;158;160"
$ebg = "1"
Set-PSReadLineOption -PromptText "$esc[4${bg};3${fg}mPS$esc[4${fg};3${bg}m$pc$esc[0m","$esc[4${ebg};3${fg}mPS$esc[4${fg};3${ebg}m$pc$esc[0m"

```

prompt

If we go with this approach, I'll also need to update the docs.

### DIFF
--- a/PSReadLine/Cmdlets.cs
+++ b/PSReadLine/Cmdlets.cs
@@ -321,7 +321,7 @@ namespace Microsoft.PowerShell
         /// If the prompt function is pure, this value can be inferred, e.g.
         /// the default prompt will use "> " for this value.
         /// </summary>
-        public string PromptText { get; set; }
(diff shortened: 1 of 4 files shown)


lzybkr:
Note that if you try my sample, you will want an powerline font installed and set the output encoding to utf8.

daxian-dbw (approved):

PR #1244, opened 2019-12-11
Fix docs to note -PromptText is now a String[]

### LINKED ISSUE #1243: Set-PSReadLineOption help not updated with PR#1180
PR #1180 updated `Set-PSReadLineOption -PromptText` to type `[string[]]` but the documentation was not updated to explain how to use this new feature.

heaths:
Fixes #1243 to update documentation after PR #1180 changed `Set-PSReadLineOption -PromptText` to be a `String[]` instead of `String`.

Also normalizes affected docs and uses escape sequence that is supported by PowerShell 5.0 (which, sadly, does not support \`e or \`u{}).

### DIFF
--- a/docs/Set-PSReadLineOption.md
+++ b/docs/Set-PSReadLineOption.md
@@ -136,7 +136,7 @@ For example, if my prompt function looked like:
 
 Then set:
 
-    Set-PSReadLineOption -PromptText "# ", "!"
+    Set-PSReadLineOption -PromptText "# ", "! "
 
 This would change the "#" in your prompt to "!" when a parse error is detected. This is especially useful with
 virtual terminal escape sequences to use colors in your prompt.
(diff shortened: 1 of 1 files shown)


docs/Set-PSReadLineOption.md
```diff
@@ -136,11 +136,14 @@ For example, if my prompt function looked like:
 
 Then set:
 
-    Set-PSReadLineOption -PromptText "# "
+    Set-PSReadLineOption -PromptText "# ", "!"
```
lzybkr: Maybe use `"! "` (add a space) so they are the same length.

I'm guessing if they aren't the same length, rendering will get messed up.

lzybkr (approved):
Thanks for taking care of this update.

docs/Set-PSReadLineOption.md
heaths: You're right. I meant to, but obviously didn't. :)

PR #1253, opened 2019-12-18
Fixed #998

### LINKED ISSUE #998: RepeatLastCharSearch(Backwards) are not behaving as expected when SearchChar initiates a character search
Environment data
----------------
PS version: 7.0.0-preview.2
PS file version: 7.0.0.0
BufferWidth: 120
BufferHeight: 9001

1. Configure the below key handlers
```powershell
Set-PSReadLineKeyHandler -Chord F3 -Function SearchChar
Set-PSReadLineKeyHandler -Chord Shift+F3 -Function SearchCharBackward
Set-PSReadLineKeyHandler -Chord F4 -Function RepeatLastCharSearch
Set-PSReadLineKeyHandler -Chord Shift+F4 -Function RepeatLastCharSearchBackwards
```
2. Enter the below text at the command line
```powershell
a b c a b c a b c
```
3. Set the cursor to the beginning of the line, then hit F3 -> b
4. Notice that the cursor moves to the first occurrence of b as expected
5. Press F4 twice, and notice each time that the cursor moves to the next occurrence of b.
6. With the cursor now on the last occurrence of b, press Shift+F4 to trigger RepeatLastCharSearchBackwards, notice that the cursor moves back to the middle occurrence of b as expected.
7. Press Shift+F4 again

Expected behavior:
The cursor continues moving backwards to the first occurrence of b

Actual behavior:
The cursor moves forward again to the last occurrence of b. If you press F4, the cursor will search backwards (not expected given the initial search forward), Shift+F4 will search forward. The expected behavior of RepeatLastCharSearch and RepeatLastCharSearchBackwards gets flipped after the first call to RepeatLastCharSearchBackwards.

Note:
This doesn't repro the other way around. If you start a SearchCharBackward via Shift+F3, RepeatLastCharSearch and RepeatLastCharSearchBackwards behave as expected with any combination searching forward and backward.

Bug:
I believe the bug to be on [this line](https://github.com/PowerShell/PSReadLine/blob/a7af25aa376ea21b356e5d35f4153b376fc424a9/PSReadLine/ReadLine.vi.cs#L79). SearchBackward is resetting isBackward to true in this scenario even if it is triggered by [RepeatLastCharSearchBackwards](https://github.com/PowerShell/PSReadLine/blob/a7af25aa376ea21b356e5d35f4153b376fc424a9/PSReadLine/ReadLine.vi.cs#L157). This overwrites the original intent of the user's search (forwards) and causes the behavior of RepeatLastCharSearch and RepeatLastCharSearchBackwards to flip even though the user's intent didn't change.

Recommendation:
I think [this line](https://github.com/PowerShell/PSReadLine/blob/a7af25aa376ea21b356e5d35f4153b376fc424a9/PSReadLine/ReadLine.vi.cs#L79) should just be removed. It is redundant with the [Set call within SearchCharBackward](https://github.com/PowerShell/PSReadLine/blob/a7af25aa376ea21b356e5d35f4153b376fc424a9/PSReadLine/ReadLine.vi.cs#L179) and inconsistent with how [Search is defined](https://github.com/PowerShell/PSReadLine/blob/a7af25aa376ea21b356e5d35f4153b376fc424a9/PSReadLine/ReadLine.vi.cs#L38) which doesn't call Set.

I built PSReadLine with this fix applied, and the expected behavior of the Repeat* commands in the repro was achieved.

mikebattista:
For the ViDeleteToChar functions, subsequent RepeatLastCharSearch(Backwards) calls behave the same as for SearchChar(Backward), so the fix here addresses those scenarios as well.

Fix #998

### DIFF
--- a/PSReadLine/ReadLine.vi.cs
+++ b/PSReadLine/ReadLine.vi.cs
@@ -76,7 +76,6 @@ namespace Microsoft.PowerShell
 
             public static void SearchBackward(char keyChar, object arg, bool backoff)
             {
-                Set(keyChar, isBackward: true, isBackoff: backoff);
                 int qty = arg as int? ?? 1;
 
                 for (int i = _singleton._current - 1; i >= 0; i--)
--- a/test/MovementTest.VI.cs
+++ b/test/MovementTest.VI.cs
@@ -508,6 +508,35 @@ namespace Test
                 "dd"
                 ));
 
+            Test("", Keys(
+                "0101010",
+                CheckThat(() => AssertCursorLeftIs(7)),
+                _.Escape,
+                _.Uphat,
+                "f1",
+                CheckThat(() => AssertCursorLeftIs(1)),
+                ";",
+                CheckThat(() => AssertCursorLeftIs(3)),
+                ";",
+                CheckThat(() => AssertCursorLeftIs(5)),
+                ",",
+                CheckThat(() => AssertCursorLeftIs(3)),
+                ",",
+                CheckThat(() => AssertCursorLeftIs(1)),
+                _.Dollar,
+                "F1",
+                CheckThat(() => AssertCursorLeftIs(5)),
+                ";",
+                CheckThat(() => AssertCursorLeftIs(3)),
+                ";",
+                CheckThat(() => AssertCursorLeftIs(1)),
+                ",",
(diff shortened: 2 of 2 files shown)


daxian-dbw (approved):
Thanks @mikebattista for your contribution!

Issue #1202, opened 2019-11-22
Pushing twice F8 leads to crash in console


Schwanowski:
Hi there

i am using Visual Studio Code 1.41 Insider edition with Powerhell V7 latest preview and have a very strange issue.

Everything looks stable but when I try to press twice F8 to debug my scripts my console crashs entirely.

Its not related to code because when i reverse to ps ise and v5 all runs perfectly fine.

Can you please advice?

Environment data
----------------

Steps to reproduce or exception report
--------------------------------------

![grafik](https://user-images.githubusercontent.com/34173158/69422425-ebd76600-0d23-11ea-80d3-b0ba5b16d8f1.png)

[Crash.txt]

[Crash.txt](https://github.com/PowerShell/PSReadLine/files/3879307/Crash.txt)

msftrncs:
I am pretty sure this is the same as #1083, and there is no perfect solution for it yet.   It happens because the history item that you are reaching with <kbd>F8</kbd> is too long for the buffer in VS Code, and PSReadLine, when it tries to display it to you, it expects to put the cursor on the first line at the end of the matching text, the same as <kbd>CTRL</kbd>+<kbd>r</kbd> (or + <kbd>s</kbd>), but that position is not accessible from VS Code's console.

Schwanowski:
Is there a chance to clear the history?

I think I should be able to reproduce the error at least (and i will be
able to work again).

Carl Morris <notifications@github.com> schrieb am Sa., 23. Nov. 2019, 07:35:

Schwanowski:
I could fix it on my machine be uninstalling the powershell plugins and re-install them again.

Unforunately, I was not able to reproduce the issue. It seems be to related to

- a long code in the history ran by f8
- maybe a braket } error in the code

msftrncs:
Any history item longer than your VS Code buffer (in lines), should cause the problem, when you use history search commands.  This is different than <kbd>UpArrow</kbd> or <kbd>DownArrow</kbd>, as those always put the cursor on the last line.  The search functions will try to place the cursor after the match, and if that match is off the buffer in VS Code, the exception occurs.

You can also erase your history, note its location is given in `Get-PSReadLineOption'.  You could delete the file or edit the file to remove entries.

Schwanowski:
ok thank you I will figure it out

daxian-dbw:
Close as a duplicate of #1083

msftrncs:
[...]

PR #1256, opened 2019-12-19
Fix the 'ArgumentOutOfRangeException' and 'IndexOutOfRangeException' when searching history with ctrl+r or ctrl+s

### LINKED ISSUE #1083: bck-i-search (Ctrl-R) quits accepting input when it finds a very long command

daxian-dbw:
Fix #1083

When using `ctrl+r` and `ctrl+s`, it's not uncommon to find a history item with multi-line text, and the matching string in the text, where the cursor is supposed to be moved to, will be scrolled up-off the buffer after rendering.
After rendering, if the cursor position to be set is already scrolled up-off the buffer (`point.Y < 0`), we should just move the cursor to the left-most position of the first line in buffer.

Before this fix, only 2 very specific scenarios where this may happen was covered. This fix replaces that with a general solution.

### DIFF
--- a/PSReadLine/History.cs
+++ b/PSReadLine/History.cs
@@ -657,6 +657,9 @@ namespace Microsoft.PowerShell
 
             if (newHistoryIndex >= 0 && newHistoryIndex <= _history.Count)
             {
+                // Set '_current' back to where it was when starting the first search, because
+                // it might be changed during the rendering of the last matching history command.
+                _current = _emphasisLength;
                 _currentHistoryIndex = newHistoryIndex;
                 var moveCursor = InViCommandMode()
                     ? HistoryMoveCursor.ToBeginning
--- a/PSReadLine/Render.cs
+++ b/PSReadLine/Render.cs
@@ -817,8 +817,8 @@ namespace Microsoft.PowerShell
                 //      to merge in the previous but still span to the current physical line). The cursor is supposed to
                 //      appear in the previous line (which is off the buffer).
                 //   3. Both 'bck-i-search' and 'fwd-i-search' may find a history command with multi-line text, and the
-                //      matching part in the text, where the cursor is supposed to be moved to, will be scroll up-off the
-                //      buffer after rendering.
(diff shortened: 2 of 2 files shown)


msftrncs:
This seems to also fix the <kbd>F8</kbd>/<kbd>Shift</kbd>-<kbd>F8</kbd> issue in #1202, except the cursor's final resting place seems to be a bit baffling, sometimes its on the first line of the buffer, sometimes on the middle of a line in the middle of the buffer, or if the history item returned is again small enough to fit entirely in the buffer, it will be on the last line of the history, where normally it would be on the first line. ....   it would seem this is a result updating `_current`, the search still seems to work correctly, but it hadn't expected the cursor to get moved away from the text entry point. 

In this case, as I have run some various Pester tests in the console, I used `describe` as the feed value for an F8 search and it returned several long scripts.  I sampled this in Windows Terminal and VS Code's terminal.

daxian-dbw:
@msftrncs Good catch! I pushed another commit to fix the cursor issue in `HistorySearchBackward` and `HistorySearchForward`, can you please take another look?

msftrncs (approved):
LGTM

daxian-dbw:
Thanks @msftrncs

PR #1259, opened 2019-12-20
Preserve and clear the saved current line properly

### LINKED ISSUE #1257: ArgumentOutOfRangeException in RevertLine after F8 then Shift-F8 after second search term was used
### Environment
PSReadLine: 2.0.0-rc1
PowerShell: 6.2.3
OS: Microsoft Windows 10.0.19536
Last 71 Keys
```
 d e s c Backspace Backspace Backspace Backspace Backspace Backspace t e s t Backspace Backspace Backspace Backspace Backspace Backspace Backspace h e l p Spacebar F8 F8 F8 F8 F8 Shift+F8 Shift+F8 Shift+F8 Shift+F8 Shift+F8 Shift+F8 Shift+F8 Shift+F8 Shift+F8 Shift+F8 Escape Escape Escape Escape d e s c r i b e Spacebar Shift+F8 F8 F8 F8 F8 F8 F8 F8 Shift+F8 Shift+F8 Shift+F8 Shift+F8 Shift+F8 Shift+F8 Shift+F8 Shift+F8 Escape
```

### Exception
```
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at Microsoft.PowerShell.PSConsoleReadLine.RevertLine(Nullable`1 key, Object arg)
```

Environment data
----------------
PS version: 6.2.3
PSReadline version: 2.0.0-rc1'; ProjectUri = 'https://github.com/PowerShell/PSReadLine
os: 10.0.19536.1000 (WinBuild.160101.0800)
PS file version: 6.2.3.0
HostName: ConsoleHost
BufferWidth: 120
BufferHeight: 9001

Steps to reproduce
I can get this exception to occur consistently.

1. Type a term. (I typed `help`)
2. Press <kbd>F8</kbd>, several times to move through several matches.
3. Press <kbd>Shift</kbd>-<kbd>F8</kbd> several times to revert back to the original input.
4. Press <kbd>Esc</kbd>.
5. Type another term. (I typed `describe`, from some Pester tests I had run)
6. Press <kbd>F8</kbd>, several times to move through several matches.
7. Press <kbd>Shift</kbd>-<kbd>F8</kbd> several times to revert back to the original input.

Note:  At this point, the original search term (in my case, `help`) was back on the command prompt, instead of the more recent term (`describe`).

8. Press <kbd>Esc</kbd>.

At this point the exception occurs.

In Windows Terminal, this causes the PowerShell Core tab to close completely, but in ConsoleHost it remained.

daxian-dbw:
Fix #1257

Major changes:
1. Preserve the saved current line and the history index while in a series of history commands.
    So when switching between different history command, such as from `F8`/`Shift+F8` (`HistorySearchBackward` and `HistorySearchForward`) to `UpArrow`/`DownArrow`  (recall history) or from `Ctrl+r`/`Ctrl+s` to `UpArrow`/`DownArrow`, the history command can continue to work as expected.
2. Clear the saved current line when we are out of a series of history commands.
    So the saved line will not leak to another unrelated history search.

### DIFF
--- a/PSReadLine/History.cs
+++ b/PSReadLine/History.cs
@@ -95,7 +95,7 @@ namespace Microsoft.PowerShell
         private int _getNextHistoryIndex;
         private int _searchHistoryCommandCount;
         private int _recallHistoryCommandCount;
-        private int _allHistoryCommandCount;
+        private int _anyHistoryCommandCount;
         private string _searchHistoryPrefix;
         // When cycling through history, the current line (not yet added to history)
         // is saved here so it can be restored.
@@ -511,7 +511,7 @@ namespace Microsoft.PowerShell
             // to check if we need to load history from another sessions now.
             MaybeReadHistoryFile();
 
-            _allHistoryCommandCount += 1;
+            _anyHistoryCommandCount += 1;
             if (_savedCurrentLine.CommandLine == null)
             {
                 _savedCurrentLine.CommandLine = _buffer.ToString();
--- a/PSReadLine/History.cs
+++ b/PSReadLine/History.cs
@@ -113,6 +113,13 @@ namespace Microsoft.PowerShell
             "password|asplaintext|token|key|secret",
             RegexOptions.Compiled | RegexOptions.IgnoreCase);
 
+        private void ClearSavedCurrentLine()
+        {
+            _savedCurrentLine.CommandLine = null;
+            _savedCurrentLine._edits = null;
+            _savedCurrentLine._undoEditIndex = 0;
+        }
+
         private AddToHistoryOption GetAddToHistoryOption(string line)
         {
             // Whitespace only is useless, never add.
@@ -217,9 +224,7 @@ namespace Microsoft.PowerShell
             // to recall the saved line.
             if (_getNextHistoryIndex == 0)
             {
-                _savedCurrentLine.CommandLine = null;
-                _savedCurrentLine._edits = null;
-                _savedCurrentLine._undoEditIndex = 0;
+                ClearSavedCurrentLine();
             }
             return result;
         }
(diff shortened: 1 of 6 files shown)


PSReadLine/History.cs
```diff
@@ -633,7 +638,7 @@ private void HistorySearch(int direction)
                     continue;
                 }
 
-                var line = newHistoryIndex == _history.Count ? _savedCurrentLine.CommandLine : _history[newHistoryIndex].CommandLine;
+                var line = _history[newHistoryIndex].CommandLine;
```
daxian-dbw, line 643: The `newHistoryIndex == _history.Count` is removed because that was already checked above:
```
if (newHistoryIndex < 0 || newHistoryIndex >= _history.Count)
{
    break;
}
```

msftrncs:
I added to #1257 some minor points.  While the changes to `RevertLine()` seem to prevent the exception, its not the only vector to the source of the issue.   

I'm wondering if `SaveCurrentLine()` needs to rely on something other than a line having not already having been saved ... such as `_searchHistoryCommandCount == 0`, to allow it to overwrite the previously saved line.

daxian-dbw:
Thanks for the additional details. You are right, it’s not just RevertLine, but a more general issue. Basically the saved line is not properly cleared after any history operation.
I will think a bit more and update the fix.

daxian-dbw:
@msftrncs I think the PR is ready for another review :) Happy new year!

msftrncs (approved):
All LGTM.

daxian-dbw:
Thanks @msftrncs!

Issue #590, opened 2017-11-22
Sample smart quote won't insert quotes when sitting on parentheses/braces


vr8hub:
Environment data
------------------
PS version: 5.1.16299.19
PSReadline version: 1.2
os: 10.0.16299.15 (WinBuild.160101.0800)
PS file version: 10.0.16299.15 (WinBuild.160101.0800)

Steps to Reproduce
--------------------
This may be WAD, but I thought I would ask. I searched all of the tickets that mentioned "quote," but didn't see anything related to this.

Using the smart quote function in the sample profile (I'm testing with the sample profile as is, this time with the two using's, I promise :) ), it won't insert quotes when the cursor is setting on a parenthesis/brace. But sitting on one of those is pretty common, especially since the function itself often inserts a matching one.

For example, typing "@{n=" results in
```
@{n=}
```
with the cursor sitting on the closing (right) brace. The next thing to type is a single quote, with the idea of having n='somename'. But, with the cursor hovering over that closing brace, the quote key does nothing (single or double, i.e. unshifted or shifted).

I'm not sure if there's a reason for not inserting a quote (or pair of quotes) in that situation, or if it's just a bug. If it's a bug, no problem, I'll wait for the fix, but if it's considered WAD, could you give a bit more information on the logic of why that shouldn't work? It's a bit inconvenient to have to type a space and then left-arrow just to be able to type the quote. And this situation happens a *lot*, at least to me.

Thanks!

lzybkr:
The state of the current sample is not ideal. I tried to make it smarter, but didn't thoroughly test it, so it's sometimes not great, and maybe worse than when it was less smart (you can dig through the history to try older versions.)

There's a good reason why it's still a sample!

vr8hub:
No worries. :) I’ll try to make some time to look at it and see if I can figure out something.

vr8hub:
This problem is in this section at the end of the key handler.
```
    if ($token.Extent.StartOffset -eq $cursor) {
```

That block only does something if the token is Generic or Identifier, otherwise it exits. Thus, none of the right-hand members of the paired characters (curly bracket, bracket, parenthesis) qualify, and therefore nothing happens (the block returns unconditionally, so it doesn't get to the "we failed to be smart" section following).

[...]

PR #1288, opened 2020-01-10
Allow SmartInsertQuote after InsertPairedBraces

theaquamarine:
Previously SmartInsertQuote would not insert a quote if a closing paren/curly was the highlighted token, eg if InsertPairedBraces had just inserted a closer - SmartInsertQuote will now handle this situation for parenthesis and curly brackets.
This commit does not allow SmartInsertQuote on a closing square bracket - that still makes it through to the "We failed to be smart, so just insert a single quote" part at the bottom.

### DIFF
--- a/PSReadLine/SamplePSReadLineProfile.ps1
+++ b/PSReadLine/SamplePSReadLineProfile.ps1
@@ -187,7 +187,7 @@ Set-PSReadLineKeyHandler -Key '"',"'" `
         }
     }
 
-    if ($null -eq $token) {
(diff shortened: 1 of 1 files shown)


theaquamarine:
This would partially resolve #590

daxian-dbw:
@theaquamarine Thanks for your contribution!
I pushed a commit to fix the misplacement of the `return` statement in `if ($token.Extent.StartOffset -eq $cursor)` block. Without this change, `SmartInsertQuote` doesn't insert anything in case the cursor is at the beginning of a token that is not `Generic` or `Identifier`, for example, when cursor is at the beginning of `$abc`.

daxian-dbw (approved):

PR #1289, opened 2020-01-10
Allow SmartInsertQuote to surround keywords

theaquamarine:
Previously it wasn't possible to type an apostrophe before a keyword with SmartInsertQuote on, now keywords are surrounded correctly.
`hasFlag()` requires .net framework 4.0 or .net core


daxian-dbw:
Oops, didn't know you fixed the return statement here.
I will revert my change in #1288 then.

daxian-dbw (approved):

PR #1293, opened 2020-01-12
Allow InsertPairedBraces to wrap selected text

theaquamarine:
Similar to SmartInsertQuote's existing ability.

### DIFF
--- a/PSReadLine/SamplePSReadLineProfile.ps1
+++ b/PSReadLine/SamplePSReadLineProfile.ps1
@@ -227,11 +227,24 @@ Set-PSReadLineKeyHandler -Key '(','{','[' `
         <#case#> '[' { [char]']'; break }
     }
 
-    [Microsoft.PowerShell.PSConsoleReadLine]::Insert("$($key.KeyChar)$closeChar")
+    $selectionStart = $null
+    $selectionLength = $null
+    [Microsoft.PowerShell.PSConsoleReadLine]::GetSelectionState([ref]$selectionStart, [ref]$selectionLength)
+
     $line = $null
     $cursor = $null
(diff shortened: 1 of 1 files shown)


PSReadLine/SamplePSReadLineProfile.ps1
```diff
@@ -227,11 +227,24 @@ Set-PSReadLineKeyHandler -Key '(','{','[' `
...
+    
+    if ($selectionStart -ne -1)
+    {
+      # Text is selected, wrap it in brackets
+      [Microsoft.PowerShell.PSConsoleReadLine]::Replace($selectionStart, $selectionLength, $key.KeyChar + $line.SubString($selectionStart, $selectionLength) + $closeChar)
+      [Microsoft.PowerShell.PSConsoleReadLine]::SetCursorPosition($selectionStart + $selectionLength + 2)
+    } else {
+      # No text is selected, insert a pair
+      [Microsoft.PowerShell.PSConsoleReadLine]::Insert("$($key.KeyChar)$closeChar")
+      [Microsoft.PowerShell.PSConsoleReadLine]::SetCursorPosition($cursor + 1)
```
daxian-dbw, line 246: Should be `$cursor - 1` as before the change. The expectation is that the subsequent input will happen within the braces.

daxian-dbw (changes requested):

PSReadLine/SamplePSReadLineProfile.ps1
daxian-dbw: Ignore this comment. `GetBufferState` is called before `Insert` now, so `$cursor + 1` is correct.

daxian-dbw (approved):

PR #1298, opened 2020-01-14
Add a -Chord param to Get-PSReadLineKeyHandler

### LINKED ISSUE #879: Add a -Key parameter to Get-PSReadLineKeyHandler
Instead of:

```powershell
Get-PSReadLineKeyHandler | ? Key -eq Enter
```

it would be nice to be able to do:

```powershell
Get-PSReadLineKeyHandler -Key Enter
```

both for convenience and for symmetry with `Set-PSReadLineKeyHandler`

Environment data
----------------

```powershell
```

theaquamarine:
Allow filtering of results by keys used in binding.

Closes #879

### DIFF
--- a/PSReadLine/Cmdlets.cs
+++ b/PSReadLine/Cmdlets.cs
@@ -809,7 +809,7 @@ namespace Microsoft.PowerShell
     [OutputType(typeof(KeyHandler))]
     public class GetKeyHandlerCommand : PSCmdlet
     {
-        [Parameter]
+        [Parameter(ParameterSetName = "FullListing")]
         public SwitchParameter Bound
         {
             get => _bound.GetValueOrDefault();
@@ -817,7 +817,7 @@ namespace Microsoft.PowerShell
         }
         private SwitchParameter? _bound;
 
-        [Parameter]
+        [Parameter(ParameterSetName = "FullListing")]
         public SwitchParameter Unbound
         {
             get => _unbound.GetValueOrDefault();
@@ -825,7 +825,8 @@ namespace Microsoft.PowerShell
         }
         private SwitchParameter? _unbound;
 
-        [Parameter]
+        [Parameter(ParameterSetName = "SpecificBindings", Position = 0, Mandatory = true)]
+        [ValidateNotNullOrEmpty]
         [Alias("Key")]
         public string[] Chord { get; set; }
 
--- a/PSReadLine/Cmdlets.cs
+++ b/PSReadLine/Cmdlets.cs
@@ -825,6 +825,10 @@ namespace Microsoft.PowerShell
         }
         private SwitchParameter? _unbound;
 
+        [Parameter]
+        [Alias("Key")]
+        public string[] Chord { get; set; }
+
         [ExcludeFromCodeCoverage]
         protected override void EndProcessing()
         {
@@ -845,7 +849,7 @@ namespace Microsoft.PowerShell
                 bound = false;
                 unbound = _unbound.Value.IsPresent;
             }
-            var groups = PSConsoleReadLine.GetKeyHandlers(bound, unbound).GroupBy(k => k.Group).OrderBy(g => g.Key);
+            var groups = PSConsoleReadLine.GetKeyHandlers(bound, unbound, Chord).GroupBy(k => k.Group).OrderBy(g => g.Key);
             foreach (var bindings in groups)
             {
                 WriteObject(bindings.OrderBy(k => k.Function), true);
--- a/PSReadLine/Cmdlets.cs
+++ b/PSReadLine/Cmdlets.cs
@@ -805,7 +805,8 @@ namespace Microsoft.PowerShell
         }
     }
 
-    [Cmdlet("Get", "PSReadLineKeyHandler", HelpUri = "https://go.microsoft.com/fwlink/?LinkId=528807")]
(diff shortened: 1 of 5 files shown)


PSReadLine/Options.cs
```diff
@@ -271,24 +271,41 @@ public static void RemoveKeyHandler(string[] key)
...
+        /// Helper function for the Get-PSReadLineKeyHandler cmdlet.
+        /// </summary>
+        /// <returns></returns>
+        public static IEnumerable<PowerShell.KeyHandler> GetKeyHandlers(bool includeBound, bool includeUnbound, string[] Chord)
         {
             var boundFunctions = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
+            var searchChords = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+            if (Chord != null) { foreach (var Key in Chord) { searchChords.Add(Key); }; }
```
daxian-dbw: This means you don't different keys like "Ctrl+k" from "Ctrl+K", but they are different key bindings.
For "Ctrl+k", it's triggered by pressing <kbd>Ctrl</kbd> and <kbd>k</kbd>
For "Ctrl+K", it's triggered by pressing <kbd>Ctrl</kbd>, <kbd>Shift</kbd>, and <kbd>k</kbd>

To accurately find the key, you will need to do the similar to [this code](https://github.com/PowerShell/PSReadLine/blob/03a0d3fe2b715eee66dcdc1304cac9e03456b434/PSReadLine/Options.cs#L178-L179), to get the `PSKeyInfo` that the specified string chord represents.

PSReadLine/Cmdlets.cs
```diff
@@ -825,6 +825,10 @@ public SwitchParameter Unbound
         }
         private SwitchParameter? _unbound;
 
+        [Parameter]
+        [Alias("Key")]
+        public string[] Chord { get; set; }
```
daxian-dbw, line 832: This should be put in a separate parameter set. When specifying `-Chord`, it implies searching from the bound bindings.
After adding this parameter, the syntax should look like below, and the `[-Bound] [-Unbound]` parameter set should be made the default.
```
PS:82> gcm Get-PSReadLineKeyHandler -Syntax

Get-PSReadLineKeyHandler [-Bound] [-Unbound] [<CommonParameters>]

Get-PSReadLineKeyHandler -Chord [<CommonParameters>]
```

daxian-dbw (changes requested):

PSReadLine/Cmdlets.cs
theaquamarine: I've added it to a separate ParameterSet, made it mandatory for that ParameterSet, and allowed it to be positional.

PSReadLine/Options.cs
theaquamarine: Implemented

PSReadLine/Options.cs
```diff
@@ -271,17 +271,64 @@ public static void RemoveKeyHandler(string[] key)
...
+                                {
+                                    Key = firstKey.KeyStr + "," + secondKey.KeyStr,
+                                    Function = entry.BriefDescription,
+                                    Description = entry.LongDescription,
+                                    Group = GetDisplayGrouping(entry.BriefDescription),
+                                };
+                            }
+                        }
+                    }
+                }
```
daxian-dbw: Key bindings in VI mode are not covered by this. Can you please add the support to VI mode too?

PSReadLine/Cmdlets.cs
```diff
@@ -809,22 +809,27 @@ public object GetDynamicParameters()
     [OutputType(typeof(KeyHandler))]
```
daxian-dbw, line 810: Can you please add `DefaultParameterSetName = "FullListing"` to the `Cmdlet(...)` attribute, as follows:
```c#
[Cmdlet("Get", "PSReadLineKeyHandler", DefaultParameterSetName = "FullListing", HelpUri = "https://go.microsoft.com/fwlink/?LinkId=528807")]
```

daxian-dbw (changes requested):

PSReadLine/Cmdlets.cs
```diff
@@ -809,22 +809,27 @@ public object GetDynamicParameters()
...
+        [Parameter(ParameterSetName = "FullListing")]
         public SwitchParameter Unbound
         {
             get => _unbound.GetValueOrDefault();
             set => _unbound = value;
         }
         private SwitchParameter? _unbound;
 
+        [Parameter(ParameterSetName = "SpecificBindings", Position = 0, Mandatory = true)]
+        [ValidateNotNullOrEmpty]
```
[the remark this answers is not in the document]
msftrncs: Does `Mandatory` also enforce `NotEmpty` or give a clear error message?  I have less experience in C# cmdlet writing, but in PowerShell, using the `Mandatory` property doesn't get quite the same affect as `ValidateNotNullOrEmpty()`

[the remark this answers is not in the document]
daxian-dbw: Good point. The parameter type is a string array, so `ValidateNotNullOrEmpty` does more than the default check from `mandatory`. This attribute should be kept.

daxian-dbw (approved):
LGTM!
Sorry for the long delay, and thanks a lot for your contribution!

PR #1300, opened 2020-01-16
Append space to self-contained completion results

theaquamarine:
Some completion result types can be reliably assumed to be the end
of an input chunk and followed by a space, so insert that space
automatically when appropriate.

eg a ProviderItem is unlikely to have anything except a space
following it, as it is the end of a path, so insert the space with
the path.

Implements #740, which has more details on the idea.

Currently, there are no tests written for this, but all existing ones pass. I'm also not sure if it should be behind an option?

### DIFF
--- a/PSReadLine/Completion.cs
+++ b/PSReadLine/Completion.cs
@@ -372,9 +372,15 @@ namespace Microsoft.PowerShell
         {
             var replacementText = completionResult.CompletionText;
             int cursorAdjustment = 0;
+            CompletionResultType[] selfContainedCompletionResultTypes = {CompletionResultType.Command,
+                CompletionResultType.Keyword, CompletionResultType.ParameterName, CompletionResultType.ProviderItem};
+
             if (completionResult.ResultType == CompletionResultType.ProviderContainer)
             {
                 replacementText = GetReplacementTextForDirectory(replacementText, ref cursorAdjustment);
+            } else if (selfContainedCompletionResultTypes.Contains(completionResult.ResultType))
+            {
+                replacementText += " ";
             }
             Replace(completions.ReplacementIndex, completions.ReplacementLength, replacementText);
             if (cursorAdjustment != 0)


PSReadLine/Completion.cs
```diff
@@ -372,9 +372,15 @@ private void DoReplacementForCompletion(CompletionResult completionResult, Comma
         {
             var replacementText = completionResult.CompletionText;
             int cursorAdjustment = 0;
+            CompletionResultType[] selfContainedCompletionResultTypes = {CompletionResultType.Command,
```
lzybkr, line 375: This should be a static array so you don't create a new array for every completion.

PSReadLine/Completion.cs
```diff
@@ -372,9 +372,15 @@ private void DoReplacementForCompletion(CompletionResult completionResult, Comma
...
+                CompletionResultType.Keyword, CompletionResultType.ParameterName, CompletionResultType.ProviderItem};
```
lzybkr, line 376: Many people do not use a space after some keywords, e.g. `if(`.

Similarly, one may not want a space after a filename, e.g. if I'm copying a file but adding an extension, e.g. `.bak`.

And infrequently one might want to type `:` after a parameter name.

lzybkr (commented):
This change would annoy a lot of people, it definitely can't be the default.

I'm not too excited about adding an option for this.

I also wonder if this belongs in PowerShell instead of PSReadLine. For example, why should completion be different in the VSCode editor or in the VSCode console window?

theaquamarine:
That's a very good point and an angle I'd not really considered, thanks!

daxian-dbw:
I agree that this should probably be done in PowerShell, and I transferred the issue #740 to https://github.com/PowerShell/PowerShell/issues/11651.
@theaquamarine Thanks for submitting this PR. If you want, feel free to submit a PR in PowerShell repo to address this in the tab completion API.

PR #1319, opened 2020-01-26
Copy PSReadLine.psd1 etc to output

theaquamarine:
This, combined with setting the debug actions as in the launchSettings attached, allows debugging in Visual Studio.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/PowerShell/PSReadLine/pull/1319)


daxian-dbw:
@theaquamarine Thanks for initiating this change!
I updated your changes to make it cleaner and work when targeting both .NET and .NET Core.

daxian-dbw (approved):

PR #1496, opened 2020-04-30
Add basic support for inline predictive suggestion

### LINKED ISSUE #1495: System.ArgumentOutOfRangeException PSReadLine: 2.1.0-beta1 PowerShell: 7.1.0-preview.2
## Environment

Name | Value
----- | -----
PS version | 7.1.0-preview.2
PSReadline version | 2.1.0-beta1
os | 10.0.19613.1000 (WinBuild.160101.0800)
PS file version | 7.1.0.0
HostName | ConsoleHost
BufferWidth | 127
BufferHeight | 1000

Last 200 Keys
```
 I n A r c h i v e F o r m a t ] : : X Z ) Ctrl+LeftArrow Ctrl+LeftArrow Ctrl+LeftArrow LeftArrow LeftArrow Shift+Ctrl+LeftArrow $ r . C o n t e n t Ctrl+z Escape # Backspace $ s RightArrow Spacebar RightArrow LeftArrow RightArrow Shift+Ctrl+LeftArrow Shift+Ctrl+LeftArrow Ctrl+RightArrow Ctrl+RightArrow RightArrow Spacebar LeftArrow LeftArrow Ctrl+LeftArrow Ctrl+LeftArrow Ctrl+LeftArrow Ctrl+LeftArrow Ctrl+LeftArrow Ctrl+LeftArrow LeftArrow Shift+LeftArrow Shift+LeftArrow Shift+LeftArrow Shift+LeftArrow Shift+LeftArrow Shift+LeftArrow Shift+LeftArrow Delete Delete Enter
 $ s e v Tab Enter
 UpArrow UpArrow Ctrl+a Ctrl+x N e w - O b j e c t Spacebar - T y p e N a m e Spacebar S e v e n Z i p E x t r a c t o r Spacebar - A r g u m e n t L i s t Spacebar @ ( $ s e v e n Z i p S t r e a m , [ S e v e n Z i p . I n A r c h i v e F o r m a t ] : : X Z ) Ctrl+LeftArrow Ctrl+LeftArrow Ctrl+LeftArrow Ctrl+LeftArrow Ctrl+RightArrow LeftArrow LeftArrow Ctrl+LeftArrow Ctrl+LeftArrow Ctrl+RightArrow [ ] Backspace Backspace Escape
```

## Exception
```
System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'length')
   at System.Text.StringBuilder.Remove(Int32 startIndex, Int32 length)
   at Microsoft.PowerShell.PSConsoleReadLine.EditItemInsertChar.Undo()
   at Microsoft.PowerShell.PSConsoleReadLine.RevertLine(Nullable`1 key, Object arg)
```

## Steps to reproduce

Investigating...

daxian-dbw:
This PR is to merge the existing predictive suggestion changes to the master branch.
Changes were refactored and cleaned up. Also tests are added.

- Currently, the only suggestion source is the PSReadLine history file.
- By default, the suggestion text is rendered with the `dark black` color `\x1b[38;5;238m`, which works OK with a black background.
  You can change it by running `Set-PSReadLineOption -Colors @{ Prediction = '<your-choice-of-color>' }`
- Parameter `-PredictionSource` is added to `Set-PSReadLineOption`, currently with the options `None` and `History`.
  The value is `None` by default, meaning that the predictive suggestion feature is disabled by default.
  A user can enable the prediction feature by set it to `History`. The feature cannot be enabled when the console output doesn't support VT.
- Make `AcceptSuggestion` and `AcceptNextSuggestionWord` bindable.
- Also fix #1495

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/PowerShell/PSReadLine/pull/1496)

### DIFF
--- a/PSReadLine/BasicEditing.cs
+++ b/PSReadLine/BasicEditing.cs
@@ -31,20 +31,26 @@ namespace Microsoft.PowerShell
             {
                 if (count <= 0)
                     return;
-            }
-            else
-            {
-                count = 1;
+                if (count > 1)
+                {
+                    var toInsert = new string(keyChar, count);
+                    if (_singleton._visualSelectionCommandCount > 0)
+                    {
+                        _singleton.GetRegion(out var start, out var length);
+                        Replace(start, length, toInsert);
+                    }
+                    else
+                    {
+                        Insert(toInsert);
+                    }
+                    return;
+                }
             }
 
             if (_singleton._visualSelectionCommandCount > 0)
             {
                 _singleton.GetRegion(out var start, out var length);
-                Replace(start, length, new string(keyChar, count));
-            }
-            else if (count > 1)
-            {
-                Insert(new string(keyChar, count));
+                Replace(start, length, new string(keyChar, 1));
             }
             else
             {
--- a/PSReadLine/BasicEditing.cs
+++ b/PSReadLine/BasicEditing.cs
@@ -274,7 +274,7 @@ namespace Microsoft.PowerShell
             SetCursorPosition(_current);
             if (_suggestionText != null)
             {
-                _suggestionText = null;
+                ResetSuggestion();
                 _console.BlankRestOfLine();
             }
 
--- a/PSReadLine/BasicEditing.cs
+++ b/PSReadLine/BasicEditing.cs
@@ -31,26 +31,20 @@ namespace Microsoft.PowerShell
             {
                 if (count <= 0)
                     return;
-                if (count > 1)
-                {
-                    var toInsert = new string(keyChar, count);
-                    if (_singleton._visualSelectionCommandCount > 0)
-                    {
-                        _singleton.GetRegion(out var start, out var length);
-                        Replace(start, length, toInsert);
-                    }
-                    else
-                    {
-                        Insert(toInsert);
-                    }
-                    return;
-                }
+            }
+            else
+            {
+                count = 1;
             }
 
             if (_singleton._visualSelectionCommandCount > 0)
             {
                 _singleton.GetRegion(out var start, out var length);
-                Replace(start, length, new string(keyChar, 1));
+                Replace(start, length, new string(keyChar, count));
+            }
+            else if (count > 1)
+            {
+                Insert(new string(keyChar, count));
             }
             else
             {
@@ -85,6 +79,8 @@ namespace Microsoft.PowerShell
         {
             _singleton.ClearStatusMessage(false);
             _singleton._current = _singleton._buffer.Length;
+
+            using var _ = _singleton.PredictionOff();
             _singleton.ForceRender();
 
             _singleton._console.Write("\x1b[91m^C\x1b[0m");
@@ -214,6 +210,8 @@ namespace Microsoft.PowerShell
 
         private bool AcceptLineImpl(bool validate)
         {
+            using var _ = PredictionOff();
+
             ParseInput();
             if (_parseErrors.Any(e => e.IncompleteInput))
             {
@@ -274,6 +272,12 @@ namespace Microsoft.PowerShell
 
             // Let public API set cursor to end of line incase end of line is end of buffer
             SetCursorPosition(_current);
+            if (_suggestionText != null)
+            {
+                _suggestionText = null;
+                _console.BlankRestOfLine();
+            }
+
             _console.Write("\n");
             _inputAccepted = true;
             return true;
--- a/PSReadLine/Changes.txt
+++ b/PSReadLine/Changes.txt
@@ -1,3 +1,26 @@
+### Version 2.1.0-beta1
+
+Pre-release notes:
+
+#### Prototype to support the fish-like auto-suggestion in PSReadLine
+
+Currently:
+
+- The only suggestion source is the PSReadLine history file.
+  The ultimate goal is to call into a prediction API in PowerShell engine that is powered by a prediction plugin.
+- By default, the suggestion text is rendered with the `dark black` color `\x1b[38;5;238m`, which works OK with a black background.
+  You can change it by running `Set-PSReadLineOption -Colors @{ Prediction = '<your-choice-of-color>' }`.
+- VI mode - now the suggestion is disabled when switching to the command mode, and enabled when switching to insert mode.
+- Very limited efforts have been spent on the VI/Emacs experience, so it could be pretty raw for them, and feedback is very welcome.
+
+When using this PSReadLine, please add the following key binding to your profile:
+
+    Set-PSReadLineKeyHandler -Key "Ctrl+f" -Function ForwardWord
+
+- Pressing `RightArrow` at the end of current typing will accept the suggestion text
+- Pressing `Ctrl+f` at the end of current typing will accept a word from the suggestion text
+- Pressing `Ctrl+z` will get you back to where you were before accepting all or part of the suggestion
+
 ### Version 2.0.1
 
 Bug fixes:
--- a/PSReadLine/Changes.txt
+++ b/PSReadLine/Changes.txt
@@ -2,24 +2,7 @@
 
 Pre-release notes:
 
-#### Prototype to support the fish-like auto-suggestion in PSReadLine
-
-Currently:
-
-- The only suggestion source is the PSReadLine history file.
-  The ultimate goal is to call into a prediction API in PowerShell engine that is powered by a prediction plugin.
-- By default, the suggestion text is rendered with the `dark black` color `\x1b[38;5;238m`, which works OK with a black background.
-  You can change it by running `Set-PSReadLineOption -Colors @{ Prediction = '<your-choice-of-color>' }`.
-- VI mode - now the suggestion is disabled when switching to the command mode, and enabled when switching to insert mode.
-- Very limited efforts have been spent on the VI/Emacs experience, so it could be pretty raw for them, and feedback is very welcome.
-
-When using this PSReadLine, please add the following key binding to your profile:
-
-    Set-PSReadLineKeyHandler -Key "Ctrl+f" -Function ForwardWord
-
-- Pressing `RightArrow` at the end of current typing will accept the suggestion text
-- Pressing `Ctrl+f` at the end of current typing will accept a word from the suggestion text
-- Pressing `Ctrl+z` will get you back to where you were before accepting all or part of the suggestion
+* Experimental support for fish-like suggestions in PSReadLine.
 
 ### Version 2.0.1
 
--- a/PSReadLine/Cmdlets.cs
+++ b/PSReadLine/Cmdlets.cs
@@ -63,10 +63,15 @@ namespace Microsoft.PowerShell
         MemoryAndFile
     }
 
-    public enum PredictionStyle
+    public enum PredictionSource
     {
         None,
-        Concise,
+        History,
+    }
+
+    public enum PredictionViewStyle
+    {
+        Default,
     }
 
     public class PSConsoleReadLineOptions
@@ -138,7 +143,12 @@ namespace Microsoft.PowerShell
 
         public const HistorySaveStyle DefaultHistorySaveStyle = HistorySaveStyle.SaveIncrementally;
 
-        public const PredictionStyle DefaultPredictionStyle = PredictionStyle.Concise;
+        /// <summary>
+        /// The predictive suggestion feature is disabled by default.
+        /// </summary>
+        public const PredictionSource DefaultPredictionSource = PredictionSource.None;
+
+        public const PredictionViewStyle DefaultPredictionViewStyle = PredictionViewStyle.Default;
 
         /// <summary>
         /// How long in milliseconds should we wait before concluding
@@ -167,7 +177,8 @@ namespace Microsoft.PowerShell
             HistorySearchCaseSensitive = DefaultHistorySearchCaseSensitive;
             HistorySaveStyle = DefaultHistorySaveStyle;
             AnsiEscapeTimeout = DefaultAnsiEscapeTimeout;
-            PredictionStyle = DefaultPredictionStyle;
+            PredictionSource = DefaultPredictionSource;
+            PredictionViewStyle = DefaultPredictionViewStyle;
 
             var historyFileName = hostName + "_history.txt";
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
@@ -323,9 +334,14 @@ namespace Microsoft.PowerShell
         public HistorySaveStyle HistorySaveStyle { get; set; }
 
         /// <summary>
+        /// Sets the source to get predictive suggestions.
+        /// </summary>
+        public PredictionSource PredictionSource { get; set; }
+
+        /// <summary>
         /// How the predictive suggestion is rendered.
         /// </summary>
-        public PredictionStyle PredictionStyle { get; set; }
+        public PredictionViewStyle PredictionViewStyle { get; set; }
 
         /// <summary>
         /// How long in milliseconds should we wait before concluding
@@ -708,12 +724,20 @@ namespace Microsoft.PowerShell
         public ScriptBlock ViModeChangeHandler { get; set; }
 
         [Parameter]
-        public PredictionStyle PredictionStyle
+        public PredictionSource PredictionSource
         {
-            get => _predictionStyle.GetValueOrDefault();
-            set => _predictionStyle = value;
+            get => _predictionSource.GetValueOrDefault();
+            set => _predictionSource = value;
         }
-        internal PredictionStyle? _predictionStyle;
+        internal PredictionSource? _predictionSource;
+
+        [Parameter]
+        public PredictionViewStyle PredictionViewStyle
+        {
+            get => _predictionViewStyle.GetValueOrDefault();
+            set => _predictionViewStyle = value;
+        }
+        internal PredictionViewStyle? _predictionViewStyle;
 
         [Parameter]
         public Hashtable Colors { get; set; }
--- a/PSReadLine/Cmdlets.cs
+++ b/PSReadLine/Cmdlets.cs
@@ -69,11 +69,6 @@ namespace Microsoft.PowerShell
         History,
     }
 
-    public enum PredictionViewStyle
-    {
-        Default,
-    }
-
     public class PSConsoleReadLineOptions
     {
         public const ConsoleColor DefaultCommentColor   = ConsoleColor.DarkGreen;
@@ -148,8 +143,6 @@ namespace Microsoft.PowerShell
         /// </summary>
         public const PredictionSource DefaultPredictionSource = PredictionSource.None;
 
-        public const PredictionViewStyle DefaultPredictionViewStyle = PredictionViewStyle.Default;
-
         /// <summary>
         /// How long in milliseconds should we wait before concluding
         /// the input is not an escape sequence?
@@ -178,7 +171,6 @@ namespace Microsoft.PowerShell
             HistorySaveStyle = DefaultHistorySaveStyle;
             AnsiEscapeTimeout = DefaultAnsiEscapeTimeout;
             PredictionSource = DefaultPredictionSource;
-            PredictionViewStyle = DefaultPredictionViewStyle;
 
             var historyFileName = hostName + "_history.txt";
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
@@ -337,11 +329,6 @@ namespace Microsoft.PowerShell
         /// Sets the source to get predictive suggestions.
         /// </summary>
         public PredictionSource PredictionSource { get; set; }
-
-        /// <summary>
-        /// How the predictive suggestion is rendered.
-        /// </summary>
-        public PredictionViewStyle PredictionViewStyle { get; set; }
 
         /// <summary>
         /// How long in milliseconds should we wait before concluding
@@ -730,14 +717,6 @@ namespace Microsoft.PowerShell
             set => _predictionSource = value;
         }
         internal PredictionSource? _predictionSource;
-
-        [Parameter]
-        public PredictionViewStyle PredictionViewStyle
-        {
-            get => _predictionViewStyle.GetValueOrDefault();
-            set => _predictionViewStyle = value;
-        }
-        internal PredictionViewStyle? _predictionViewStyle;
 
         [Parameter]
         public Hashtable Colors { get; set; }
--- a/PSReadLine/Cmdlets.cs
+++ b/PSReadLine/Cmdlets.cs
@@ -63,6 +63,12 @@ namespace Microsoft.PowerShell
         MemoryAndFile
     }
 
+    public enum PredictionStyle
+    {
+        None,
+        Concise,
+    }
+
     public class PSConsoleReadLineOptions
     {
         public const ConsoleColor DefaultCommentColor   = ConsoleColor.DarkGreen;
@@ -77,6 +83,10 @@ namespace Microsoft.PowerShell
         public const ConsoleColor DefaultMemberColor    = ConsoleColor.Gray;
         public const ConsoleColor DefaultEmphasisColor  = ConsoleColor.Cyan;
         public const ConsoleColor DefaultErrorColor     = ConsoleColor.Red;
+
+        // Use dark black by default for the suggestion text.
+        // Find the most suitable color using https://stackoverflow.com/a/33206814
+        public const string DefaultPredictionColor = "\x1b[38;5;238m";
 
         public static EditMode DefaultEditMode = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
             ? EditMode.Windows
@@ -128,6 +138,8 @@ namespace Microsoft.PowerShell
 
         public const HistorySaveStyle DefaultHistorySaveStyle = HistorySaveStyle.SaveIncrementally;
 
+        public const PredictionStyle DefaultPredictionStyle = PredictionStyle.Concise;
+
         /// <summary>
         /// How long in milliseconds should we wait before concluding
         /// the input is not an escape sequence?
@@ -155,6 +167,7 @@ namespace Microsoft.PowerShell
             HistorySearchCaseSensitive = DefaultHistorySearchCaseSensitive;
             HistorySaveStyle = DefaultHistorySaveStyle;
             AnsiEscapeTimeout = DefaultAnsiEscapeTimeout;
+            PredictionStyle = DefaultPredictionStyle;
 
             var historyFileName = hostName + "_history.txt";
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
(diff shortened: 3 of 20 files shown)


daxian-dbw:
@lzybkr @SteveL-MSFT Gentle ping :) Please take a look when you have time. Thanks!

docs/Set-PSReadLineOption.md
```diff
@@ -612,6 +615,28 @@ Accept pipeline input: false
...
+### -PredictionStyle
+
+Specifies how PSReadLine should render the predictive suggestions.
+
+Valid values are:
+
+-- None: disable the predictive suggestion feature
+
+-- Concise: render the predictive suggestion in a concise view
+
```
lzybkr, line 627: Without knowing what other styles will be added, it's not clear that `Concise` is sufficiently descriptive. I'd probably be happier if it was called `Default` (preferred) or maybe `Simple`.

docs/Set-PSReadLineOption.md
```diff
@@ -612,6 +615,28 @@ Accept pipeline input: false
...
+```yaml
+Type: PredictionStyle
+Parameter Sets: (All)
+Aliases:
+
+Required: False
+Position: Named
+Default value: Concise
+Accept pipeline input: false
```
lzybkr, line 636: I really think this is a bad default.

docs/Set-PSReadLineOption.md
```diff
@@ -612,6 +615,28 @@ Accept pipeline input: false
 Accept wildcard characters: false
 ```
 
+### -PredictionStyle
```
lzybkr: >PredictionStyle [](start = 5, length = 15)

This parameter suggests something other than rendering, e.g. prediction from history or some other magic. How about `PredictionViewStyle`?

PSReadLine/BasicEditing.cs
```diff
@@ -31,26 +31,20 @@ public static void SelfInsert(ConsoleKeyInfo? key = null, object arg = null)
             {
                 if (count <= 0)
                     return;
-                if (count > 1)
```
lzybkr: Am I correct that these changes have nothing to do with predictive suggestions?

If so, can you move this change to a new PR?

PSReadLine/Changes.txt
```diff
@@ -1,3 +1,26 @@
+### Version 2.1.0-beta1
+
+Pre-release notes:
+
+#### Prototype to support the fish-like auto-suggestion in PSReadLine
+
```
lzybkr: You've released it, so it's not a prototype.

How about `Experimental support for fish-like suggestions in PSReadLine`.

PSReadLine/Changes.txt
```diff
@@ -1,3 +1,26 @@
...
+Currently:
+
+- The only suggestion source is the PSReadLine history file.
+  The ultimate goal is to call into a prediction API in PowerShell engine that is powered by a prediction plugin.
```
lzybkr: >The ultimate goal is to call into a prediction API in PowerShell engine that is powered by a prediction plugin. [](start = 2, length = 111)

I don't think future plans belong in the change log.

lzybkr: Actually, I don't think most of this section belongs in the change log - documentation should live elsewhere.

---
In reply to: [420564331](https://github.com/PowerShell/PSReadLine/pull/1496#discussion_r420564331) [](ancestors = 420564331)

PSReadLine/Cmdlets.cs
```diff
@@ -63,6 +63,12 @@ public enum AddToHistoryOption
         MemoryAndFile
     }
 
+    public enum PredictionStyle
+    {
+        None,
+        Concise,
```
lzybkr: >Concise [](start = 8, length = 7)

Consider `Default` instead.

PSReadLine/Cmdlets.cs
```diff
@@ -63,6 +63,12 @@ public enum AddToHistoryOption
         MemoryAndFile
     }
 
+    public enum PredictionStyle
```
lzybkr: >PredictionStyle [](start = 16, length = 15)

Consider `PredictionDisplayStyle` or `PredictionViewStyle` instead.

PSReadLine/Cmdlets.cs
```diff
@@ -128,6 +138,8 @@ public class PSConsoleReadLineOptions
 
         public const HistorySaveStyle DefaultHistorySaveStyle = HistorySaveStyle.SaveIncrementally;
 
+        public const PredictionStyle DefaultPredictionStyle = PredictionStyle.Concise;
```
lzybkr: I have concerns with this default.

PSReadLine/Suggestion.cs
```diff
@@ -0,0 +1,61 @@
+using System;
+
+namespace Microsoft.PowerShell
+{
+    public partial class PSConsoleReadLine
+    {
+        private class SuggestionModeRestore : IDisposable
+        {
```
lzybkr: See `UseViCommandModeTables` for an alternative pattern in use in this code base.

PSReadLine/History.cs
```diff
@@ -677,6 +679,21 @@ private void HistorySearch(int direction)
             }
         }
 
+        private string GetHistorySuggestion(string text)
+        {
```
lzybkr, line 689: I'd like to see comments here for possible improvements, e.g. not using StartsWith and instead extracting a partial pipeline/statement from a longer pipeline or multiline command.

PSReadLine/PublicAPI.cs
```diff
@@ -235,5 +235,34 @@ public static bool TryGetArgAsInt(object arg, out int numericArg, int defaultNum
             numericArg = 0;
             return false;
         }
+
+        /// <summary>
+        /// Accept the suggestion text if there is one.
+        /// </summary>
+        public static void AcceptSuggestion()
+        {
```
lzybkr: Is this meant to be bindable? If so, it has the wrong signature.

PSReadLine/PublicAPI.cs
```diff
@@ -235,5 +235,34 @@ public static bool TryGetArgAsInt(object arg, out int numericArg, int defaultNum
...
+                using var _ = _singleton.PredictionOff();
+                Replace(0, _singleton._buffer.Length, _singleton._suggestionText);
+            }
+        }
+
+        /// <summary>
+        /// Accept the current or the next suggestion text word.
+        /// </summary>
+        public static void AcceptNextSuggestionWord(int numericArg)
+        {
```
lzybkr: Is this meant to be bindable? If so it has the wrong signature.

PSReadLine/Render.cs
```diff
@@ -310,38 +311,61 @@ void MaybeEmphasize(int i, string currColor)
...
+                }
+            }
+
+            if (inSelectedRegion)
+            {
+                _consoleBufferLines[currentLogicalLine].Append("\x1b[0m");
+            }
+
+            inSelectedRegion = false;
+            if (suggestion != null)
```
lzybkr, line 342: This isn't necessary as it's not read anymore, but if you really want it, at least move it inside the previous `if` block that writes out the reset CSI.

lzybkr: Oh, I see, it's needed because of the local function call to UpdateColorsIfNecessary - so just move it.

---
In reply to: [420571809](https://github.com/PowerShell/PSReadLine/pull/1496#discussion_r420571809) [](ancestors = 420571809)

PSReadLine/Render.cs
```diff
@@ -310,38 +311,61 @@ void MaybeEmphasize(int i, string currColor)
...
+                {
+                    if (charToRender == '\n')
+                    {
+                        currentLogicalLine += 1;
+                        if (currentLogicalLine == _consoleBufferLines.Count)
+                        {
+                            _consoleBufferLines.Add(new StringBuilder(COMMON_WIDEST_CONSOLE_WIDTH));
+                        }
+                        _consoleBufferLines[currentLogicalLine].Append(Options.ContinuationPrompt);
+                    }
```
lzybkr: The continuation prompt could change the color so you'll need to restore the prediction color.

PSReadLine/Render.cs
```diff
@@ -310,38 +311,61 @@ void MaybeEmphasize(int i, string currColor)
...
+            }
+
+            if (inSelectedRegion)
+            {
+                _consoleBufferLines[currentLogicalLine].Append("\x1b[0m");
+            }
+
+            inSelectedRegion = false;
+            if (suggestion != null)
+            {
```
lzybkr, line 343: I'm slightly worried about duplicated logic in this loop and the main rendering loop (which already has special code to cover text after the last token for cases where the parser fails to produce enough tokens).

At least consider refactoring the inner loop into a function that could be shared.

lzybkr:
>         private bool LineIsMultiLine()

Duplicate logic - you should just call your new function with _buffer as the parameter.

---
Refers to: PSReadLine/Render.cs:1211 in e876ffc. [](commit_id = e876ffcd323e8dd2d4fa8e543b1635e60cfd79ee, deletion_comment = False)

lzybkr:
>         {

Duplicate logic - just call your new function with _buffer as the first parameter.

---
Refers to: PSReadLine/Words.cs:92 in e876ffc. [](commit_id = e876ffcd323e8dd2d4fa8e543b1635e60cfd79ee, deletion_comment = False)

PSReadLine/Suggestion.cs
```diff
@@ -0,0 +1,61 @@
...
+                _suggestionText = null;
+                _lastRenderedTextHash = 0;
+                return null;
+            }
+
+            try
+            {
+                uint textHash = _lastRenderedTextHash;
+                _lastRenderedTextHash = FNV1a32Hash.ComputeHash(text);
+                if (textHash != _lastRenderedTextHash)
```
lzybkr: Why not just compare strings? Hashing has to look at every character but string comparison can return quickly, e.g. if lengths don't match or at the first mismatch.

PSReadLine/Words.cs
```diff
@@ -94,6 +94,57 @@ private bool InWord(int index, string wordDelimiters)
...
+        {
+            char c = text[index];
+            return !char.IsWhiteSpace(c) && wordDelimiters.IndexOf(c) < 0;
+        }
+
+        private int FindForwardSuggestionWordPoint(int currentIndex, string wordDelimiters)
+        {
+            System.Diagnostics.Debug.Assert(
+                _suggestionText != null && _suggestionText.Length > _buffer.Length,
+                "Caller needs to make srue the suggestion text exist.");
```
lzybkr: >srue [](start = 38, length = 4)

spelling

docs/Set-PSReadLineOption.md
daxian-dbw: Think about it more, I do want to use it beyond rendering. The idea is to have `Concise` (can be named `Default` as you suggested) and `List` options.
- When `Default` is in use, PSReadLine will use history as the only source for the inline suggestion. The suggestion text will be rendered inline just like what the prototype does today.
- When `List` is in use, besides history, PSReadLine will also call into a new API in `S.M.A`, which will call into a potential prediction plugin, to get suggestions. The suggestion results (up to 2 results from history + up to 4 results from the API) will be rendered as a drop down list, which can be navigated by <kbd>UpArrow</kbd> and <kbd>DownArrow</kbd> and selected by <kbd>Enter</kbd> or <kbd>RightArrow</kbd>.

The 2nd part is still in design and will raise privacy concerns as PSReadLine will be sending the user input to somewhere (in the cloud?). But it should clarify that this option does suggest something more than rendering. The doc needs to be updated for sure.

rkeithhill: Do you want to mix the display style and prediction source?  Seems like those could be two different settings.  Maybe I'd like to see a list of predictions from just my PSRL history?  Or maybe I only want to see the simple/single view but from an alternate source (not yet implemented)?

docs/Set-PSReadLineOption.md
daxian-dbw: I'm fine to disable the prediction by default. Will update the code.

PSReadLine/BasicEditing.cs
daxian-dbw: You are right. It's a small refactoring of the existing code. I will move it to a new PR.

PSReadLine/Changes.txt
daxian-dbw: Updated.

PSReadLine/Changes.txt
daxian-dbw: I will remove this section.

PSReadLine/Cmdlets.cs
daxian-dbw: I will update to use `Default`

PSReadLine/Cmdlets.cs
daxian-dbw: Given the clarification above, I prefer to keep the name as is.

PSReadLine/Cmdlets.cs
daxian-dbw: I will disable the prediction by default.

PSReadLine/Suggestion.cs
daxian-dbw: Good to know this. I will switch to that pattern.

docs/Set-PSReadLineOption.md
daxian-dbw: I will change it to `Default`

docs/Set-PSReadLineOption.md
daxian-dbw: When the suggestion is from a different source, there is no guarantee the suggestion can be rendered inline (simple/single view), because the returned results may not be prefixed with what the user have typed on the console. But maybe it's better to have 2 separate settings anyways, it's more clear that way.

PSReadLine/Cmdlets.cs
daxian-dbw: Renamed to `PredictionViewStyle`

PSReadLine/History.cs
daxian-dbw: Will add the comment.

PSReadLine/PublicAPI.cs
daxian-dbw: I think it's desirable to make it bindable. I will update the code.

PSReadLine/PublicAPI.cs
daxian-dbw: Same here, will update the code.

PSReadLine/Render.cs
daxian-dbw: Moved to the if block.

PSReadLine/Render.cs
daxian-dbw: Fixed by removing the duplicate logic.

PSReadLine/Words.cs
daxian-dbw: Fixed.

PSReadLine/Render.cs
daxian-dbw: Very valid concern. I have refactored the code to render one char into a function.

PSReadLine/Suggestion.cs
daxian-dbw: `_lastRenderedTextHash` is not really that useful. It saves a call to `GetHistorySuggestion` only in some cases where text is pasted to the console. I removed it for now.

daxian-dbw:
@lzybkr I'm not sure what these 2 comments are referring to: https://github.com/PowerShell/PSReadLine/pull/1496#issuecomment-624470594 https://github.com/PowerShell/PSReadLine/pull/1496#issuecomment-624470853. Can you please clarify them a bit?

Other than that, I'm almost done addressing the comments, and will push the changes tomorrow.

PSReadLine/Words.cs
```diff
@@ -94,6 +94,57 @@ private bool InWord(int index, string wordDelimiters)
             return !char.IsWhiteSpace(c) && wordDelimiters.IndexOf(c) < 0;
         }
 
+        private bool InWord(string text, int index, string wordDelimiters)
```
lzybkr: Logic is duplicated with the method above. The method above should call this method instead.

daxian-dbw: I refactored this method to directly take a char, and use it in the above method.

PSReadLine/BasicEditing.cs
daxian-dbw: @lzybkr The changes in `SelfInsert` is the only refactoring code that is not directly related to the predictive suggestion feature.
Given it's small enough, is it OK to keep it in this PR?

daxian-dbw:
@lzybkr I think I addressed all comments. A summary:
1. Two settings are used: `-PredictionSource` and `PredictionViewStyle`
    `PredictionSource` -- options are `None` and `History` today;
    `PredictionViewStyle` -- option is `Default` only today.
2. Make the predictive suggestion feature turned off by default (`PredictionSource` is none by default)
3. Make `AcceptSuggestion` and `AcceptNextSuggestionWord` bindable.
4. Refactor to use the existing patterns as well as avoid redundant logic.
5. Add more tests.

PSReadLine/BasicEditing.cs
lzybkr: Well it's not really good engineering to include unrelated fixes with new features. For example, if one wanted to revert the feature, they'd also lose this fix.

PSReadLine/Cmdlets.cs
```diff
@@ -63,6 +63,17 @@ public enum AddToHistoryOption
         MemoryAndFile
     }
 
+    public enum PredictionSource
+    {
+        None,
+        History,
+    }
+
+    public enum PredictionViewStyle
```
lzybkr: I don't think this enum should be added until there is a need - right now there is no option as there is only 1 alternative.

PSReadLine/Cmdlets.cs
```diff
@@ -78,6 +89,10 @@ public class PSConsoleReadLineOptions
         public const ConsoleColor DefaultEmphasisColor  = ConsoleColor.Cyan;
         public const ConsoleColor DefaultErrorColor     = ConsoleColor.Red;
 
+        // Use dark black by default for the suggestion text.
+        // Find the most suitable color using https://stackoverflow.com/a/33206814
+        public const string DefaultPredictionColor = "\x1b[38;5;238m";
```
lzybkr, line 89: See `LegacyWin32Console` which won't handle this color. Mapping arbitrary colors to a `ConsoleColor` seems problematic - it might be best to disable predictive display (at least as currently implemented) when using the `LegacyWin32Console`.

PSReadLine/Cmdlets.cs
```diff
@@ -309,6 +333,16 @@ public object ContinuationPromptColor
...
 
+        /// <summary>
+        /// Sets the source to get predictive suggestions.
+        /// </summary>
+        public PredictionSource PredictionSource { get; set; }
+
+        /// <summary>
+        /// How the predictive suggestion is rendered.
+        /// </summary>
+        public PredictionViewStyle PredictionViewStyle { get; set; }
```
lzybkr: Can we remove this until there is a need?

PSReadLine/Render.cs
```diff
@@ -1194,6 +1200,16 @@ private bool LineIsMultiLine()
             return false;
         }
 
+        private bool LineIsMultiLine(string text)
```
lzybkr: Line 1193 is `LineIsMultiLine()` (no arguments). It should call this function passing `_buffer`.

PSReadLine/Suggestion.cs
```diff
@@ -0,0 +1,119 @@
+using System;
+
+namespace Microsoft.PowerShell
+{
+    public partial class PSConsoleReadLine
+    {
+        private string _suggestionText;
+        private string _userInputText;
```
lzybkr: This field needs a better name or a good comment. I could get confused and think it's the current command line.

PSReadLine/Suggestion.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            return _suggestionText?.Substring(text.Length);
+        }
+
+        /// <summary>
+        /// Accept the suggestion text if there is one.
+        /// </summary>
+        public static void AcceptSuggestion(ConsoleKeyInfo? key = null, object arg = null)
+        {
+            if (!TryGetArgAsInt(arg, out var numericArg, 1))
```
lzybkr: I think I'd reject any non-null arg. It would be playing with fire to work like other args (take the nth suggestion - crazy and confusing).

You could also just ignore the arg assuming you have no ideas how you might use the arg.

PSReadLine/Suggestion.cs
```diff
@@ -0,0 +1,119 @@
...
+                Replace(0, _singleton._buffer.Length, _singleton._suggestionText);
+            }
+        }
+
+        /// <summary>
+        /// Accept the current or the next suggestion text word.
+        /// </summary>
+        public static void AcceptNextSuggestionWord(ConsoleKeyInfo? key = null, object arg = null)
+        {
+            if (!TryGetArgAsInt(arg, out var numericArg, 1))
```
lzybkr, line 78: I think I'd reject any non-null arg. It would be playing with fire to work like other args (take the nth suggestion - crazy and confusing).

You could also just ignore the arg assuming you have no ideas how you might use the arg.

lzybkr (changes requested):

PSReadLine/Render.cs
daxian-dbw: Passing in `_buffer` would mean calling `_buffer.ToString()` when it's not really necessary. I don't think that's preferable.

PSReadLine/Cmdlets.cs
daxian-dbw: Fair. I will remove this enum and the corresponding option.

PSReadLine/Cmdlets.cs
daxian-dbw: Good point. I recall seeing issue reporting it doesn't work on 2012R2. How about throwing an error when setting `PredictionSource` to something other than `None` when `LegacyWin32Console` is in use?

PSReadLine/Cmdlets.cs
daxian-dbw: Will do.

PSReadLine/Suggestion.cs
daxian-dbw: Will rename the field.

PSReadLine/Suggestion.cs
daxian-dbw: Didn't think much about it, but yeah, it feels confusing to allow accepting the nth suggestion.
I will ignore the arg.

PSReadLine/BasicEditing.cs
daxian-dbw: Understood. I will revert changes in this method and submit a separate PR.

PSReadLine/Suggestion.cs
daxian-dbw: I feel it's useful to allow accepting more than 1 words from the suggestion text.

PSReadLine/Render.cs
lzybkr: You could accept a Span to avoid the `ToString` call, right?

PSReadLine/Suggestion.cs
lzybkr: Got it.

PSReadLine/Render.cs
daxian-dbw: You cannot get a span/readonlyspan from a `StringBuilder`. I updated the code to use `line.Contains('\n')` and have this method removed.

daxian-dbw:
@lzybkr All your comments are addressed. The refactoring changes to `SelfInsert` was separated out to #1510. 
Summaries:
1. `PredictionViewStyle` and its related code is removed until another view is implemented.
2. Not support the predictive suggestion feature when the legacy console is in use.
3. Ignore `arg` for `AcceptSuggestion`
4. Removed `LineIsMultiLine(string text)`
5. Renamed the field to `_lastUserInput`

lzybkr (approved):

JustinGrote:
@daxian-dbw I had a question about this. There is the new `AcceptSuggestion` bind, but even if it is unbound, `ForwardChar` bound to rightarrow still does the completion. Is this intended or am I missing something?

daxian-dbw:
@JustinGrote The function `ForwardChar` has `AcceptSuggestion` built in, and similarly, `ForwardWord` has `AcceptNextSuggestionWord` built in.
The functions `AcceptSuggestion` and `AcceptNextSuggestionWord` are made public and bindable so as to allow users to use different key bindings to accept suggestion text.
For example, [here is a sample](https://github.com/PowerShell/PSReadLine/blob/master/PSReadLine/SamplePSReadLineProfile.ps1#L584) to make `RightArrow` to accept next word instead the entire suggestion text.

nippyin:
When i run this command Set-PSReadLineOption -PredictionSource History. I expect following settings to be persistent. PredictionSource                       : History

However, after VScode restart or new instance launch this is again set to None.
image

JustinGrote:
@nippyin add it to your `$profile`

nippyin:
> @nippyin add it to your `$profile`
Thanks Justin, can you share me the syntax please.

rkeithhill:
The syntax is the same - just put that command in your profile script located at this path `$profile.CurrentUserAllHosts`. Well, execute that to get the path.  For me, that path is:
```
C:\Users\Keith\Documents\PowerShell\profile.ps1
```
It will be a bit different for you (different username at least).

If the file doesn't exist, create it.  Then place this line in the file:
```powershell
Set-PSReadLineOption -PredictionSource History
```
Now, whenever you start PowerShell, it will have this setting configured.

nippyin:
> he syntax is the same - j

this is nice, however is there VScode settings for this where we can set the parameter value as History? something similar to "powershell.codeFormatting.autoCorrectAliases": true

rkeithhill:
The above would apply to all PowerShell hosts including the VSCode terminal and PS Integrated Console. To be clear, this is PSReadLine module configuration - not specific to VSCode.  That said, if for some reason you only want this setting configured in the PS Integrated Console in VSCode, then put the command in the PSIC's profile script: `$HOME\Documents\PowerShell\Microsoft.VSCode_profile.ps1`.

musm:
This is a really nice feature, many thanks.

Is it not possible to have `<tab>` complete the suggested completion, instead of having to use the `<right>` arrow key?

JustinGrote:
@musm you can customize the keybindings
https://docs.microsoft.com/en-us/powershell/module/psreadline/set-psreadlinekeyhandler?view=powershell-7

musm:
Right now mine is set to `Function` yet that doesn't seem to want to complete predictive suggestions. Any chance you know why or which keybinding I need to change it to? The documentation isn't clear on this one.

It looks like you need to set `AcceptSuggestion ` Is it possible to set the keybinding to frist try to  AcceptSuggestion  (f it exists) and if not do a `Complete`?

SSvilen:
Hello,

I saw Jason Helmick showing PredictionViewStyle during the session with Jeffrey on Ignite, but I don't see that feature anywhere - even in the RC. Will this be implemented at all?

JustinGrote:
@SSvilen my understanding is that it is on a private branch and will surface for 2.2

PR #1511, opened 2020-05-10
Simplify the detection of Windows Terminal

daxian-dbw:
Use the environment variable `WT_SESSION` instead of walk up the process chain.

```
WT_SESSION=e1d171d9-4739-4cd0-b086-50d729f39631
```

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/PowerShell/PSReadLine/pull/1511)

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_Report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_Report.md
@@ -33,16 +33,8 @@ Please run the script in the PowerShell session where you ran into the issue and
 
 & {
     $hostName = $Host.Name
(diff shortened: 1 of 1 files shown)


rkeithhill:
An issue with this approach is that env vars get inherited so the presence of this env var does not necessarily mean you are running in WT.  For example, start WT then start up a (VS) code session from WT.  In the VSCode terminal, note that `WT_SESSION` is defined but that terminal is not WT.

daxian-dbw:
@rkeithhill Good catch. Both `WT_SESSION` and `TERM_PROGRAM` are way less useful in deciding where the shell is running in considering the environment variable inheritance 😦
Closing this PR.

PR #1517, opened 2020-05-12
Relax the sensitive words we filter

### LINKED ISSUE #1459: DefaultAddToHistoryHandler filters too aggressively
## Description of the new feature/enhancement

The default history filter as implemented in #1058 is nice but including `key` causes too many false positives, as `key` is often used in a non-sensitive context e.g. as a synonym for ID.

I would suggest removing it but the question of course is whether minimizing false positives or false negatives should have priority here. IMO the filter should be less aggressive (i.e. filter less) by default for the following reasons:

- Not being able to find some commands in the history is confusing.
- There are bound to be many commands with secrets that will not be caught by the filter anyway.
- There was no filter for a long time, and e.g. bash does not have one either, so there should be a certain level of awareness about not entering secrets directly in commands.

daxian-dbw:
Fix #1459

Change `key` to `apikey` to reduce false positives.
`key` as a sensitive word is too aggressive. Some false positives that one can easily run into:
- `Set/Get/Remove-PSReadLineKeyHandler`
- `-KeyFilePath` parameter on all PowerShell remoting cmdlets, such as `Enter/New-PSSession`, `Invoke-Command`
- `-Key` parameter on CIM cmdlets such as `New/Get-CimInstance`, which specifies the properties that are used as keys.
- `Update-TypeData -DefaultKeyPropertySet`
-  `-Key` and `-SecureKey` parameter on `ConvertTo/ConvertFrom-SecureString` accept a byte array and a `SecureString` respectively, so they don't need to be removed from the history.

`-NuGetApiKey` is used by `Publish-Module/Script` to accept a sensitive string, and `apikey` can continue to filter those cmdlets out from the history.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/PowerShell/PSReadLine/pull/1517)

### DIFF
--- a/PSReadLine/History.cs
+++ b/PSReadLine/History.cs
@@ -111,7 +111,7 @@ namespace Microsoft.PowerShell
 
         // Pattern used to check for sensitive inputs.
         private static readonly Regex s_sensitivePattern = new Regex(
-            "password|asplaintext|token|key|secret",
+            "password|asplaintext|token|apikey|secret",
             RegexOptions.Compiled | RegexOptions.IgnoreCase);
 
         private void ClearSavedCurrentLine()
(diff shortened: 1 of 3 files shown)


lzybkr (approved):
Maybe add a note in `Changes.txt`?

daxian-dbw:
@lzybkr I usually update `Changes.txt` before a release for all fixes included in the release.

PR #1527, opened 2020-05-14
useLocalScope to true so that it works in CLM

TylerLeonhardt:
`.AddScript("0")` dot sources the script. Since it's created in C#, the script is considered trusted... You can't dot source trusted scripts into a runspace that is running in ConstrainedLanguage mode otherwise it throws an exception.

This allows PSReadLine to be loaded in the PowerShell extension for VS Code when running in ConstrainedLanguage mode.

### DIFF
--- a/PSReadLine/ReadLine.cs
+++ b/PSReadLine/ReadLine.cs
@@ -653,7 +653,7 @@ namespace Microsoft.PowerShell
             {
                 try
                 {
-                    var results = ps.AddScript("$Host").Invoke<PSHost>();
+                    var results = ps.AddScript("$Host", useLocalScope: true).Invoke<PSHost>();
                     PSHost host = results.Count == 1 ? results[0] : null;
                     hostName = host?.Name;
                 }
--- a/PSReadLine/ReadLine.cs
+++ b/PSReadLine/ReadLine.cs
@@ -230,7 +230,7 @@ namespace Microsoft.PowerShell
                             if (ps == null)
                             {
                                 ps = System.Management.Automation.PowerShell.Create(RunspaceMode.CurrentRunspace);
-                                ps.AddScript("0");
+                                ps.AddScript("0", useLocalScope: true);
                             }
 
                             // To detect output during possible event processing, see if the cursor moved


TylerLeonhardt:
cc @daxian-dbw if this could be backported to 2.0.x I'd really appreciate it. Then I can ship a PowerShell stable extension with ConstrainedLanguage mode support with PSRL.

daxian-dbw:
@TylerLeonhardt The change is benign, but there is another use of `AddScript` [here](https://github.com/PowerShell/PSReadLine/blob/master/PSReadLine/ReadLine.cs#L656) that happens at the initialization of PSReadLine. Since you didn't change that one, I assume it didn't cause a problem, so I'm not sure if your change in this PR is the root cause of your problem or something else more fundamental is wrong.

TylerLeonhardt:
@daxian-dbw that code you shared is wrapped in a try/catch. The error is probably swallowed. This change I made is the only change needed to prevent a "classic PSRL exception" from showing up in my console over and over again.

TylerLeonhardt:
After my last commit, the history file is "Visual Studio Code Host_history.txt"

daxian-dbw:
When a script runs from within a nested pipeline, the script will be treated as `FullLanguage`, see the code [here](https://github.com/PowerShell/PowerShell/blob/8bd96e03a9d8f34bf99171de407abf5263d17901/src/System.Management.Automation/engine/hostifaces/Command.cs#L470-L474).

PSReadLine is invoked directly from within PS extension, whose pipeline thread (default Runspace) runs in `ConstrainedLanguage` when the machine is locked down.
`.AddScript` will make the script `FullLanguge` but the default Runspace is `ConstrainedLanguage`, so the invocation will fail due to `AddScript(string)` is dot-sourcing.

So adding `useLocalScope: true` is necessary for PS extension to use PSReadLine in constrained language mode.
I will publish version 2.0.2 to service this change. @TylerLeonhardt Can you please verify if there is anything else in PSReadLine that will be required for it to work in PSES in CLM?

daxian-dbw (approved):

TylerLeonhardt:
@daxian-dbw can you get that 2.0.2 version out for me?

daxian-dbw:
@TylerLeonhardt Will do. I'm currently busy on the .NET CAP work, can it wait till early next week?

TylerLeonhardt:
Yeah that's fine

PR #1626, opened 2020-06-30
Teach PSReadline to not force the background color during render

### LINKED ISSUE #830: Stop setting the background color when erasing
As of Windows 10 build 18298, the [new “Terminal” settings](https://blogs.msdn.microsoft.com/commandline/2018/12/10/new-experimental-console-features/) allow setting _default_ ForegroundColor and BackgroundColor values which are _separate_ from the 16 color "ConsoleColor" palette.

PSReadLine doesn't use the ANSI/VT escape sequence for _default_ when erasing text, but instead overwrites the line with `$Host.UI.RawUI.BackgroundColor` (e.g. when pressing backspace, or when using the arrow keys to navigate through the history buffer with commands of different lengths -- worse when some have multiple lines).

PSReadLine needs to start using the "default" color, instead of specifying a color.

DHowett:
# PR Summary

Upcoming versions of the Windows Console will be able to differentiate
a color set by the Win32 API from a color set by VT. This code
transforms a Win32 color into a VT color, and then uses that VT color
during rendering. When an enlightened console host receives this VT, it
may handle it differently (translate it differently, reverse it
differently, store it differently).

I've switched the _defaults_ over to "39;49" (the well-supported SGR
codes for "foreground, background to default") and taught the two
console translators how to handle them. I went back to the Windows
source code and confirmed: *conhost supports 39;49 everywhere
`ENABLE_VIRTUAL_TERMINAL_PROCESSING` is supported.*

The first commit that introduced 39/49 is in Windows 10 TH2, the very
same release that introduced VT handling.

There's a lot more information in https://github.com/PowerShell/PSReadLine/issues/830#issuecomment-650508857

Fixes #830.
Related to microsoft/terminal#5952

## PR Checklist

- [x] Summarized changes
- [x] Make sure you've added one or more new tests (tests updated to handle this)
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/PowerShell/PSReadLine/pull/1626)


DHowett:
I'm surprised about this failure. I ran `./build.ps1 -test` and I saw none of these failures locally...

DHowett:
Alright, I went with a different approach. SGR `39` and `49` exist to restore the console color to the default. I've taught the mock and legacywin32 consoles how to handle this (in the same way as the windows console handles it), and switched the "reset to the default colors" sequence to set 39;49.

~**This does change the behavior of Render**. Where it would previously enforce the current colors, it now prefers the _default_ colors.~ This does not appear to be true; I cannot get PSReadline (before my change) to respect the active colors.

~For what reasons were we keeping the "current" colors? Are they incontrovertible?~

DHowett:
Tested in the legacy console:
image

daxian-dbw (approved):
@DHowett Thanks for the contribution!

PR #1694, opened 2020-07-25
[vi-mode] 'dd' now handles single line or multiline buffers consistently

### LINKED ISSUE #1693: [vi-mode] 'dd' handles single line and multiline buffers differently - code should be simplified.
## Proposed change

The 'dd' operation behaves differently for single line an multiline buffers whereas the behaviour should be consistent with vim.
The added benefit would be to simplify the code, albeit while slightly changing the behaviour when pasting the deleted line.

## Proposed technical implementation details (optional)

- Get rid of the `LineIsMultiline` check.
- Make use of the various functions `GetLogicalLineCount`, `GetLogicalLineNumber`, etc.
- Use the internal `DeleteLineImpl` handler that supports line boundaries and hooks into the Undo/Redo mechanism.

springcomp:
# PR Summary

Fix #1693
The code has been simplified so that the 'dd' command now handles single line or multiline buffers consistently.

I believe the documentation for the `DeleteLine` function to be correct. The concept of _line_ is what has changed with this PR, as well as previous and upcoming work on improving multiline buffer handling in vi-mode.

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/PowerShell/PSReadLine/pull/1694)

### DIFF
--- a/PSReadLine/ReadLine.vi.cs
+++ b/PSReadLine/ReadLine.vi.cs
@@ -722,36 +722,25 @@ namespace Microsoft.PowerShell
         /// </summary>
         public static void DeleteLine(ConsoleKeyInfo? key = null, object arg = null)
         {
-            if (_singleton.LineIsMultiLine())
+            var lineCount = _singleton.GetLogicalLineCount();
+            var lineIndex = _singleton.GetLogicalLineNumber() - 1;
+
+            TryGetArgAsInt(arg, out var requestedLineCount, 1);
+
+            var deletePosition = DeleteLineImpl(lineIndex, requestedLineCount);
+
+            // goto the first character of the first remaining logical line
+            var newCurrent = deletePosition + 1;
+
+            if (lineIndex + requestedLineCount >= lineCount)
             {
-                var lineCount = _singleton.GetLogicalLineCount();
-                var lineIndex = _singleton.GetLogicalLineNumber() - 1;
-
-                TryGetArgAsInt(arg, out var requestedLineCount, 1);
-
-                var deletePosition = DeleteLineImpl(lineIndex, requestedLineCount);
-
-                // goto the first character of the first remaining logical line
-                var newCurrent = deletePosition + 1;
-
-                if (lineIndex + requestedLineCount >= lineCount)
-                {
-                    // if the delete operation has removed all the remaining lines
-                    // goto the first character of the previous logical line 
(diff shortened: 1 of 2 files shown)


PSReadLine/ReadLine.vi.cs
```diff
@@ -764,20 +753,19 @@ private static int DeleteLineImpl(int lineIndex, int lineCount)
         {
             var range = _singleton._buffer.GetRange(lineIndex, lineCount);
 
-            var deleteText = _singleton._buffer.ToString(range.Offset, range.Count);
-
-            _clipboard.LinewiseRecord(deleteText);
+            _clipboard.LinewiseRecord(_singleton._buffer.ToString(range.Offset, range.Count));
```
daxian-dbw: @springcomp The change to `DeleteLineImpl` doesn't seem to be needed. Was it an oversight?

daxian-dbw (changes requested):

test/YankPasteTest.VI.cs
```diff
@@ -51,8 +51,8 @@ public void ViPasteAfterDelete()
             Test("abcd", Keys(
                 "abcd", _.Escape,
                 "dd", CheckThat(() => AssertLineIs("")), CheckThat(() => AssertCursorLeftIs(0)),
-                'p', CheckThat(() => AssertLineIs("abcd")), CheckThat(() => AssertCursorLeftIs(3)),
-                'P', CheckThat(() => AssertLineIs("abcabcdd")), CheckThat(() => AssertCursorLeftIs(6)),
+                'p', CheckThat(() => AssertLineIs("\nabcd")), CheckThat(() => AssertCursorLeftIs(3)),
+                'P', CheckThat(() => AssertLineIs("\nabcd\nabcd")), CheckThat(() => AssertCursorLeftIs(3)),
```
daxian-dbw: Maybe use `continuationPrefixLength` too as in the below test?

PSReadLine/ReadLine.vi.cs
springcomp: It was indeed. I will revert the changes.

daxian-dbw (approved):

PR #1848, opened 2020-10-04
Fix #1393: avoid dead key check for known unambiguous control characters

### LINKED ISSUE #1393: Ctrl-C doesn't work if PowerShell was started with Russian keyboard layout
## Environment

```none
PS version: 7.0.0
PSReadline version: 2.0.0
PS file version: 7.0.0.0
HostName: ConsoleHost
BufferWidth: 170
BufferHeight: 32766
```

## Steps to reproduce

1. Enable Russian keyboard layout on Windows (I use Alt-Shift to do that)
2. Start PowerShell Core 7.0 (which includes PSReadLine 2.0)
3. Press Ctrl-C in the shell

## Expected behavior

Ctrl-C should terminate current command/input/whatever, it should work as Ctrl-C usually do.

## Actual behavior

Russian letter "с" getting inserted into the command line, and no actual Ctrl-C behavior happens.

Ctrl-C is still broken even if I switch to the English keyboard layout (then English letter "c" gets inserted inctead of actual Ctrl-C behavior heppening).

ForNeVeR:
# PR Summary

This introduces the fix suggested by @lzybkr in https://github.com/PowerShell/PSReadLine/issues/1393#issuecomment-686622043. I've verified that the wrong behavior is gone after the fix.

## PR Checklist

- [x] Summarized changes
- [ ] ~~Make sure you've added one or more new tests~~
  - I'm not sure that the changes are easily (if at all) testable; on my machine, many of the tests are failing no matter what and I'm not sure if they should or not
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - [ ] ~~Documentation needed at [PowerShell-Docs](https://github.com/MicrosoftDocs/PowerShell-Docs)~~
        - [ ] ~~Doc Issue filed:~~

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/PowerShell/PSReadLine/pull/1848)


ForNeVeR:
I could use some help with tests.

- `Test.en_US_Windows.CharacterSearchBackwardEmacs` and `Test.en_US_Windows.ShowKeyBindings` already fail on my machine even before the changes
- for other three failed tests, I couldn't able to understand what's going on (yet), and I couldn't (yet) find what's been broken in the actual shell

Considering the (very poor!) overall state of the tests on my machine (I have 88 tests failing out of 842), I'm not even sure the test failure is not spurious.

ForNeVeR:
Regarding other failures: I believe that the test keymap data is filled incorrectly, and that incorrect data was earlier fixed by the code I've turned off in this PR.

For example, let's take a look at the test `Test.en_US_Windows.YankLastNth`. It fails on the first check, namely, when pressing `Ctrl-Alt-y`.

On my actual system in runtime, this generates a key event with `KeyChar` of `25` (or at least that's what I see in the breakpoint set at the line I've changed), which then gets properly converted to `y`, so the code works in the actual console.

But `KeyInfo-en-US-windows.json` (the test data file) contains the following:

```
    {
        "Key":  "Ctrl+Alt+y",
        "KeyChar":  "\u0000",
        "ConsoleKey":  "Y",
        "Modifiers":  "Alt, Control"
    },
```

The same data (with `KeyChar = '\0'`) was regenerated by `Get-KeyInfoData.ps1` script for me, so, something's probably not right here. This test data, in turn, generates not the same `ConsoleKeyInfo` for tests, and that's why the test fails.

Any ideas on how to proceed?

daxian-dbw:
I'm using the en-US keyboard layout.
All tests passed without the change in this PR. When applying the change from this PR, I got the same failures as in the CI.

For `Ctrl-Alt-y`, I got the exact `KeyChar`, `ConsoleKey` and `Modifiers` as shown the above comment.

image

daxian-dbw:
The change is a breaking change. At least for key chords like `Ctrl+1` and `Ctrl+2`, they will be interpreted to `Ctrl+@` with this change, so any bindings with those chords won't work after this change.

Den-dp:
Is there something we can do in order to make it a non-breaking change?

ForNeVeR:
I really could use some help from the team. I'm not able to fix the tests, since they don't work on my computer in the first place.

Probably, we could invent another approach to hotkey handling, but I'm not that savvy with the Windows console API w.r.t. hotkeys, unfortunately.

terekhov-d:
Any update on this?

daxian-dbw:
Superseded by #3786

Issue #262, opened 2015-07-05
Set-PSReadlineOption -HistorySaveStyle SaveAtExit not working on Windows 10 build 10162


rkeithhill:
This option hasn't been working in the last couple of Windows 10 builds dating back to build 10130 I think.  I can workaround this for now by not using this option and letting it save incrementally.

gwojan:
I don't think it works correctly on Windows 7 either. I recently started using -SaveAtExit with the latest builds and my history wasn't saved at all.

rkeithhill:
@gwojan  Thanks for confirming the issue.

daxian-dbw:
@rkeithhill Can you please confirm if this is still an issue with the latest PSReadLine module?

ghost:
This issue is closed because it has been marked as requiring author feedback but has not had any activity for **7 days**. If you think the issue is still relevant, please reopen and provide your feedback.

rkeithhill:
@daxian-dbw AFAICT nothing is saved in the history file if you use `SaveAtExit`. 🤷‍♂️

daxian-dbw:
@rkeithhill Thanks for checking and confirming.

puxu-msft:
same issue, 2.2.5 and win11.

Shuangcheng-Ni:
I've tested this command with PowerShell 5.1 and 7.3 on Windows 11. Is this the expected behavior?
| PowerShell Version \ exit by | `exit` command | close console window |
|:---------------------:|:-----------------:|:-------------------------:|
| 5.1 | not save | not save |
| 7.3 | not save | save |

StevenBucher98:
I tested this and have had the following behaviors with PSReadline 2.3.3 and Windows 11:

PowerShell Version \ exit by | exit command | close console window | close console window Terminal
-- | -- | -- | --
5.1 | not save | save | save 
7.3 | not save | save | save

Personally I would expect history to be saved when `exit` is typed, unsure if this is a technical limitation @daxian-dbw?

Also I get no history saved when I do `SaveAtExit` on Mac.

PR #1909, opened 2020-10-28
Add the prediction `ListView` and also hook up with the `CommandPrediction` APIs introduced in PS 7.1

### LINKED ISSUE #1591: Predictive text behaves something odd
PSReadLine is behaving something odd with the predictions, when the prediction is in multiple lines and if we are not continuing with the predicted command then it leaves the marks from the 2nd line in the screen, notice the same in the 2nd Pic and 1st pic shows the prediction.

image

image

## Environment

```none
PS version: 7.0.1
PSReadline version: 2.1.0-beta2
os: 10.0.19041.1 (WinBuild.160101.0800)
PS file version: 7.0.1.0
BufferWidth: 209
BufferHeight: 52
```

## Steps to reproduce

You can see in the pics that I have attached.

## Expected behavior

It should wipe the remaining predicted text
## Actual behavior
It leaves the predicted text from the 2nd line on the screen

daxian-dbw:
# PR Summary

## Feature Overview

Major feature level changes:
1. Add the `ListView` to the predictive intellisense feature in PSReadLine.
2. Hook up PSReadLine with the `CommandPrediction` APIs introduced in PS 7.1, so when using PSReadLine with PS 7.1, a user can import a predictor module and PSReadLine can render the suggestions returned from the predictor module.

There will be 4 options for the `PredictionSource`: `None`, `History`, `Plugin`, `HistoryAndPlugin`.
- `None` -- disable the predictive intellisense feature, which is the default.
- `History` -- enable the predictive intellisense feature, and use the PSReadLine history as the only source.
- `Plugin` -- enable the predictive intellisense feature, and use the plugins (`CommandPrediction`) as the only source.
- `HistoryAndPlugin` -- enable the predictive intellisense feature, and use both history and plugin as the sources.

There will be 2 views for the `PredictionViewStyle`: `InlineView` and `ListView`.
- `InlineView` -- the style as existing today, similar as in fish shell and zsh.
- `ListView` -- suggestions are rendered in a drop down list, and users can select using <kbd>UpArrow</kbd> and <kbd>DownArrow</kbd>. 

tabcom2

Two more color settings were added to support customization of the `ListView`:
- `ListPredictionColor` -- set color for the leading `>` character and the trailing source name, e.g. `[History]`. By default, it uses `DarkYellow` as the foreground color.
- `ListPredictionSelectedColor` -- set color for indicating a list item is selected. By default, it uses `DarkBlack` as the background color.

image

## Handling of Commonly Used Keys

To improve the user experience, some commonly used keys/functionalities are specially handled:
1. <kbd>Escape</kbd> (`RevertLine`) -- in the list view, it will revert to the original line, no matter whether or not a list item is selected, and clear the list view. So <kbd>UpArrow</kbd> and <kbd>DownArrow</kbd> can then be used to navigate history commands as normal.

tabcom2

2. <kbd>Ctrl+z</kbd> (`Undo`) -- in the list view, it will revert to the original line, no matter whether or not a list item is selected, but keep the list view.

tabcom2

3. <kbd>Alt+n</kbd> (`DigitArgument`) -- in the list view, the digit argument won't refresh the suggestion results, and it can be used together with <kbd>UpArrow</kbd> and <kbd>DownArrow</kbd> to speed up the selection.

tabcom2

4. Moving cursor, or selection operations such as <kbd>Shift+LeftArrow</kbd>, <kbd>Shift+Home</kbd>, etc. won't refresh the suggestion results. This is very necessary for a good user experience.

tabcom2

5. Switching between `InlineView` and `ListView` with a hotkey <kbd>F2</kbd> (`SwitchPredictionView`).

tabcom2

## Feedback Trigger

The `CommandPrediction` has 2 APIs for the host client to provide feedback to the predictor plugin/module:
`OnSuggestionAccepted`, `OnCommandLineAccepted`.

`OnCommandLineAccepted` is triggered when the current command line is accepted (`AcceptLine`). Call to `AcceptLine` due to <kbd>Ctrl+c</kbd> is filtered out.

`OnSuggestionAccepted` is triggered differently based on the prediction view in use:
- `InlineView` -- callback is triggered when either the suggestion is accepted via <kbd>RightArrow</kbd>, or the next word of the suggestion is accepted via <kbd>Ctrl+f</kbd> (`ForwardWord`).
- `ListView` -- callback is triggered when
   - a list item is selected and then any edit is done on the selected item. For example, typing new characters, deleting characters, moving cursor or selecting part of the item and then replace with new typing or delete the selected range.
   - a list item is selected and then the user types <kbd>Enter</kbd> to accept the item.

## Work with downlevel PowerShell

On supported downlevel PowerShell versions, such as Windows PowerShell 5.1 and PowerShell 7.0:

- Only `History` can be used as the prediction source. The `Plugin` and `HistoryAndPlugin` options for `PredictionSource` only work with PS 7.1+.
- `ListView` works on all supported downlevel PowerShell versions when using the `History` prediction source.

image
image

We are still shipping only a single `Microsoft.PowerShell.PSReadLine2.dll` that is built targeting `net461`. To bridge it with different PowerShell runtimes (PS7.1+ vs. downlevel), I introduced the `Microsoft.PowerShell.PSReadLine.Polyfiller.dll`. The module structure is shown as follows:

```none
C:\REPOS\PSREADLINE\BIN\DEBUG\PSREADLINE
│   Changes.txt
│   License.txt
│   Microsoft.PowerShell.PSReadLine2.dll
│   PSReadLine.format.ps1xml
│   PSReadLine.psd1
│   PSReadLine.psm1
│   SamplePSReadLineProfile.ps1
│   System.Runtime.InteropServices.RuntimeInformation.dll
│
├───net461
│       Microsoft.PowerShell.PSReadLine.Polyfiller.dll
└───net5.0
        Microsoft.PowerShell.PSReadLine.Polyfiller.dll
```

- The `net461` version of the `Polyfiller.dll` contains the stub definitions of `CommandPrediction` and related types.
- The `net5.0` version of the `Polyfiller.dll` contains type forwarders that forwards the `CommandPrediction` and related types to the 7.1+ version of `System.Management.Automation.dll`.

By loading the right `Polyfiller.dll`, the `Microsoft.PowerShell.PSReadLine2.dll` built against `net461` can work on all supported versions of PowerShell.

To make sure the right `Polyfiller.dll` is loaded, PSReadLine registers a handler to `AppDomain.CurrentDomain.AssemblyResolve` upon module loading and unregisters the handler during module unloading. The handler decides which `Polyfiller.dll` to use based on the current .NET version (`Environment.Version`).

## Build and Tests

Build scripts were updated to build PSReadLine with multi-target `net461` and `net5.0`. Also, the artifacts are packaged in the structure as shown above.

12 xUnit tests were added and each of them contains quite amount of sub tests. The new test code is more than 1300 line.
To test the plugin scenario, I made the use of the new `CommandPrediction` APIs mockable, by adding corresponding member methods to `IPSConsoleReadLineMockableMethods`.

Also fix #1591

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [x] Doc Issue filed: https://github.com/MicrosoftDocs/PowerShell-Docs/issues/6812

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/PowerShell/PSReadLine/pull/1909)


PSReadLine/Prediction.Entry.cs
```diff
@@ -0,0 +1,351 @@
...
+                                    .Append(_singleton._options.EmphasisColor)
+                                    .Append(SuggestionText, InputMatchIndex, input.Length)
+                                    .EndColorSection(selectionHighlighting)
+                                    .Append(SuggestionText, rightStartindex, rightStrLen)
+                                    .Append("...");
+                                break;
+                            }
+
+                            int leftPlusRightLenInCells = leftMidLenInCells + midRightLenInCells - midLenInCells * 2;
+                            if (leftPlusRightLenInCells <= textWidth - 7)
```
SteveL-MSFT, line 230: Seems like `7` should be a const defined somewhere

PSReadLine/Prediction.Entry.cs
```diff
@@ -0,0 +1,351 @@
...
+                                //   '...LLLLLLLLLLLLLLMMMMMMMM...MMMMMMMMRRRRRRRRRRRRRRRR...'
+                                int midRemainingLenInCells = textWidth / 3 - 3;
+                                int midLeftCellLen = midRemainingLenInCells / 2;
+                                int midRightCellLen = midRemainingLenInCells - midLeftCellLen;
+
+                                int midLeftStrLen = SubstringLengthByCells(SuggestionText, InputMatchIndex, midLeftCellLen);
+                                int midRightStrLen = SubstringLengthByCellsFromEnd(SuggestionText, rightStartindex - 1, midRightCellLen);
+                                int leftStrLen = SubstringLengthByCellsFromEnd(SuggestionText, InputMatchIndex - 1, midRemainingLenInCells);
+                                int rightStrLen = SubstringLengthByCells(SuggestionText, rightStartindex, midRemainingLenInCells);
+                                int spacesNeeded = textWidth - midRemainingLenInCells * 3 - 9;
```
SteveL-MSFT: Where does the `9` come from?

PSReadLine/Prediction.Views.cs
```diff
@@ -0,0 +1,780 @@
...
+            /// </remarks>
+            private void AggregateSuggestions()
+            {
+                var results = GetPredictionResults();
+                if (results?.Count > 0)
+                {
+                    try
+                    {
+                        _listItems ??= new List<SuggestionEntry>();
+                        _cacheList1 ??= new List<int>();
```
SteveL-MSFT: Is there something more descriptive than cachelist1 and 2?  or have a comment describing how these are used if they are two caches of the same thing.

PSReadLine/Prediction.Views.cs
```diff
@@ -0,0 +1,780 @@
...
+                {
+                    var item = _listItems[_selectedIndex];
+                    if (item.PredictorId != Guid.Empty)
+                    {
+                        _singleton._mockableMethods.OnSuggestionAccepted(item.PredictorId, item.SuggestionText);
+                    }
+                }
+            }
+
+            internal override void Clear(bool cursorAtEol)
```
SteveL-MSFT, line 527: Instead of passing in `cursorAtEol`, would it be better to have a helper calculate it?

PSReadLine/Render.Helper.cs
```diff
@@ -0,0 +1,161 @@
...
+namespace Microsoft.PowerShell
+{
+    public partial class PSConsoleReadLine
+    {
+        private void WriteBlankLines(int count)
+        {
+            _console.BlankRestOfLine();
+            for (int i = 1; i < count; i++)
+            {
+                _console.Write("\n");
```
SteveL-MSFT, line 16: Environment.Newline?

SteveL-MSFT (changes requested):

PaulHigin (changes requested):

PSReadLine/Render.Helper.cs
```diff
@@ -0,0 +1,161 @@
...
+
+            // The following is based on http://www.cl.cam.ac.uk/~mgk25/c/wcwidth.c
+            // which is derived from http://www.unicode.org/Public/UCD/latest/ucd/EastAsianWidth.txt
+
+            bool isWide = c >= 0x1100 &&
+                (c <= 0x115f || /* Hangul Jamo init. consonants */
+                    c == 0x2329 || c == 0x232a ||
+                    (c >= 0x2e80 && c <= 0xa4cf &&
+                    c != 0x303f) || /* CJK ... Yi */
+                    (c >= 0xac00 && c <= 0xd7a3) || /* Hangul Syllables */
```
iSazonov, line 89: Is performance important here?
```suggestion
                    (((uint)c - (uint)0xac00) <= ((uint)0xd7a3) - (uint)0xac00) || /* Hangul Syllables */
```
We could use helper method https://source.dot.net/#System.Private.CoreLib/UnicodeUtility.cs,141

theJasonHelmick (approved):
Updated PM RFC #262 to reflect property and value changes, including keybindings, described here.

PSReadLine/Prediction.Entry.cs
```diff
@@ -0,0 +1,351 @@
+/********************************************************************++
+Copyright (c) Microsoft Corporation.  All rights reserved.
```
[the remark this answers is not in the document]
daxian-dbw: The PSReadLine is licensed under `BSD 2-Clause "Simplified" License` instead of MIT. Shall I change the header to the following?
```
// Copyright (c) Microsoft Corporation.
// Licensed under the BSD 2-Clause "Simplified" License.
```

PSReadLine/Prediction.Entry.cs
daxian-dbw: `7` is completely a guessing number, and it's explained in the comment right below it:
> ... if the (left+right) portions are not too long -- namely we can reserve 7 cells at least for the mid portion, ...

So I think I will just keep it as is without defining a const for it.

PSReadLine/Prediction.Entry.cs
daxian-dbw: 9 is actually the length of 3 ellipsis strings, namely 3 * 3. I made changes to make this more clear.

PSReadLine/Cmdlets.cs
```diff
@@ -66,8 +66,16 @@ public enum AddToHistoryOption
 
     public enum PredictionSource
     {
-        None,
-        History,
+        None = 1,
+        History = 2,
+        Plugin = 4,
+        HistoryAndPlugin = 6,
```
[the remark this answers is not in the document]
daxian-dbw: With `None = 0`, you will always have `HasFlag(None)` or `(value & None) == None` to be true. That's what I tried to avoid.

PSReadLine/Prediction.Views.cs
daxian-dbw: This parameter exists for perf reason. To clear the view, we need to get the `CursorTop` as if the cursor is at the end of the input text. But in some common scenarios, the cursor is already at the end and hence we can avoid calculating that again.

PSReadLine/Render.Helper.cs
daxian-dbw: PSReadLine always uses `\n` for rendering. The only exception is in `CaptureScreen`, which copies the buffer cells to clipboard on Windows  platform.

PSReadLine/OnImportAndRemove.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            AppDomain.CurrentDomain.AssemblyResolve -= ResolveAssembly;
+        }
+
+        /// <summary>
+        /// Load the correct 'Polyfiller' assembly based on the runtime.
+        /// </summary>
+        private static Assembly ResolveAssembly(object sender, ResolveEventArgs args)
+        {
+            if (args.Name != "Microsoft.PowerShell.PSReadLine.Polyfiller, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null")
```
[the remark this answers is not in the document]
daxian-dbw: It surely will when releasing 2.2.0-beta1. But it will not be strongly signed, if that's what you meant to ask.

PSReadLine/Prediction.Entry.cs
```diff
@@ -0,0 +1,351 @@
...
+    {
+        /// <summary>
+        /// This type represents an individual suggestion entry.
+        /// </summary>
+        private struct SuggestionEntry
+        {
+            internal readonly Guid PredictorId;
+            internal readonly string Source;
+            internal readonly string SuggestionText;
+            internal readonly int InputMatchIndex;
```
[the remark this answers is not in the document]
daxian-dbw: I planned to make the list item generated from `GetListItemText(...)` to be cacheable, but didn't get to it. That would make the struct mutable, so let's keep it as is for now.

PSReadLine/Prediction.Entry.cs
```diff
@@ -0,0 +1,351 @@
...
+                    .Append(textMetadataColor);
+
+                if (sourceStrLen == Source.Length)
+                {
+                    line.Append(Source);
+                }
+                else
+                {
+                    line.Append(Source, 0, sourceStrLen)
+                        .Append("...");
```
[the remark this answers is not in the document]
daxian-dbw: Agreed. Will use a const string.

PSReadLine/Render.Helper.cs
daxian-dbw: This is existing code (moved to the new file), and thus I'd like to not change it in this PR.
How much is the `(value - lower) <= (upper - lower)` faster than `value >= lower && value <= upper`?

PSReadLine/Prediction.Views.cs
daxian-dbw: I cannot find a suitable descriptive name for those 2 cache lists, but I will add comments here to explain what they are for.

PSReadLine/Disposable.cs
```diff
@@ -6,6 +6,13 @@ internal sealed class Disposable : IDisposable
     {
         private Action m_onDispose;
 
+        internal static readonly Disposable NonOp = new Disposable();
+
+        private Disposable()
```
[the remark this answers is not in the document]
daxian-dbw: I want to hide it from being used elsewhere in PSReadLine, and thus made it private. Except for this readonly `NonOp`, all usage of `Disposable` should go through the parameterized constructor.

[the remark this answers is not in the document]
PaulHigin: Ok, but I think you can remove the null assignment.

PSReadLine.build.ps1
```diff
@@ -120,22 +134,30 @@ task LayoutModule BuildMainModule, {
...
         # ensure files have \r\n line endings as the signing tool only uses those endings to avoid mixed endings
         $content = Get-Content -Path $file -Raw
         Set-Content -Path (Join-Path $targetDir (Split-Path $file -Leaf)) -Value (ConvertTo-CRLF $content) -Force
     }
 
+    if ($Framework -eq "net461") {
+        if (-not (Test-Path "$targetDir/net461")) {
+            New-Item "$targetDir/net461" -ItemType Directory -Force > $null
+        }
+        if (-not (Test-Path "$targetDir/net5.0")) {
```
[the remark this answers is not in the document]
daxian-dbw: According to [The future of .NET Standard](https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/), you can think of .NET 5 as .NET Standard vNext.

> .NET 5 and all future versions will always support .NET Standard 2.1 and earlier. The only reason to retarget from .NET Standard to .NET 5 is to gain access to more runtime features, language features, or APIs. **So, you can think of .NET 5 as .NET Standard vNext.**

> To summarize:
> - Use net5.0 for code sharing moving forward.

Given that, I will keep the folder name as `net5.0` for now. If things change, we can revisit this before 2.2.0 GA.

PSReadLine/OnImportAndRemove.cs
```diff
@@ -0,0 +1,37 @@
...
+            {
+                return null;
+            }
+
+            string root = Path.GetDirectoryName(typeof(OnModuleImportAndRemove).Assembly.Location);
+            string subd = (Environment.Version.Major >= 5) ? "net5.0" : "net461";
```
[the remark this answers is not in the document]
daxian-dbw: Addressed above.

daxian-dbw:
@SteveL-MSFT @PaulHigin and @iSazonov Thanks for the review! I addressed all comment except for the header to be used for new code in PSReadLine. Please take another look when you get a chance.

PSReadLine/Disposable.cs
```diff
@@ -6,6 +6,13 @@ internal sealed class Disposable : IDisposable
...
+        {
+            m_onDispose = null;
```
PaulHigin, line 13: Nitpick.  This member assignment is not necessary and can be removed.

PaulHigin (approved):
Just one minor comment.

PSReadLine/Render.Helper.cs
iSazonov: This patterns comes from .Net repo. They uses it broadly in all char operations. It use 2 operation instead of 3 ones in first pattern and more fast commands. It is huge optimization relatively (up to 30%) and some ns in absolute values.

Another way they uses is 2- or 3-level translation tables.

PSReadLine/OnImportAndRemove.cs
```diff
@@ -0,0 +1,37 @@
...
+            string path = Path.Combine(root, subd, "Microsoft.PowerShell.PSReadLine.Polyfiller.dll");
+
+            return Assembly.LoadFrom(path);
```
iSazonov, line 34: It seems it is one time operation. Can we exclude the event handler and load the dll in an init code at startup?

iSazonov (commented):
Interesting, how many does MSFT team want to support for Windows PowerShell? I mean polyfill.dll could be enhanced in separate repo and reused as nuget package.

PSReadLine/OnImportAndRemove.cs
daxian-dbw: Registering the `AssemblyResolve` event make it able to load the assembly lazily, which is especially desired for a module that gets loaded at startup.

SeeminglyScience:
> I'm just wondering if PowerShell Polyfill will be a separate project so that other developer can benefit from it.

That's sort of difficult to do well without also shipping the 7.1 version of the polyfill *with* 7.1 itself.  Otherwise you'd either need a proxy assembly that registers the resolve event handler, or require the package consumer to set one up themselves.

I think there is potentially a lot of value in shipping the `TypeForwardedTo` version of these assemblies with the version they're introduced, but that could get unmaintainable quickly.  Every new API (or rather, every version with a new API) would need it's own assembly, in-box type forward version, and nuget polyfill version.

It's worth discussing but probably out of scope for this PR.

Polyfill/CommandPrediction.cs
```diff
@@ -0,0 +1,135 @@
...
+        /// <summary>
+        /// Gets the tooltip of the suggestion.
+        /// </summary>
+        public string ToolTip { get; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="PredictiveSuggestion"/> class.
+        /// </summary>
+        /// <param name="suggestion">The predictive suggestion text.</param>
+        public PredictiveSuggestion(string suggestion)
```
SeeminglyScience, line 62: Consider decorating statically visible members with `HiddenAttribute`.  Since these will be visible from PowerShell, adding `Hidden` will reduce the chance of folks trying to use it.

PSReadLine/OnImportAndRemove.cs
daxian-dbw: It is critical too. People complain that starting a pwsh session is way slower than Windows PowerShell, and that's for interactive use. Also, the predictive intellisense feature is off by default.
The test we have are focusing on automation scenarios.

daxian-dbw:
@SeeminglyScience @iSazonov We can use PSReadLine as an experiment to see how the `polyfiller` idea works. Foreseeably, the `polyfiller` assemblies shipped in PSReadLine will be updated as new types/members exposed in PowerShell engine in 7.2.

Polyfill/CommandPrediction.cs
daxian-dbw: `HiddenAttribute` added to properties and methods of the stub types.

daxian-dbw:
@SteveL-MSFT Can you please review again? Thanks!

PSReadLine/Prediction.Entry.cs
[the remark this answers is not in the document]
SteveL-MSFT: Yes, let's use that

SteveL-MSFT (approved):
The copyright headers can be updated as a separate PR

PSReadLine/Prediction.Entry.cs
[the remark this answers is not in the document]
daxian-dbw: Will address in a separate PR.

PR #1923, opened 2020-11-03
Fix IDE0044: Add readonly modifier

xtqqczze:
https://docs.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0044

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/PowerShell/PSReadLine/pull/1923)

### DIFF
--- a/PSReadLine/ConsoleLib.cs
+++ b/PSReadLine/ConsoleLib.cs
@@ -13,10 +13,10 @@ namespace Microsoft.PowerShell.Internal
         // custom ReadKey implementation. This is not a public API, but it is
         // part of a private contract with that project.
         #pragma warning disable CS0649
-        private static Func<bool, ConsoleKeyInfo> _readKeyOverride;
+        private static readonly Func<bool, ConsoleKeyInfo> _readKeyOverride;
         #pragma warning restore CS0649
 
-        private static Lazy<Func<bool, ConsoleKeyInfo>> _readKeyMethod = new Lazy<Func<bool, ConsoleKeyInfo>>(
+        private static readonly Lazy<Func<bool, ConsoleKeyInfo>> _readKeyMethod = new Lazy<Func<bool, ConsoleKeyInfo>>(
             () => _readKeyOverride == null ? Console.ReadKey : _readKeyOverride);
 
         public int CursorLeft
--- a/PSReadLine/ReadLine.cs
+++ b/PSReadLine/ReadLine.cs
@@ -46,8 +46,8 @@ namespace Microsoft.PowerShell
 
         private bool _delayedOneTimeInitCompleted;
 
-        private IPSConsoleReadLineMockableMethods _mockableMethods;
-        private IConsole _console;
+        private readonly IPSConsoleReadLineMockableMethods _mockableMethods;
+        private readonly IConsole _console;
         private ICharMap _charMap;
         private Encoding _initialOutputEncoding;
         private bool _skipOutputEncodingChange;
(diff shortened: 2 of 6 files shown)


PSReadLine/ConsoleLib.cs
```diff
@@ -13,10 +13,10 @@ internal class VirtualTerminal : IConsole
         // custom ReadKey implementation. This is not a public API, but it is
         // part of a private contract with that project.
         #pragma warning disable CS0649
-        private static Func<bool, ConsoleKeyInfo> _readKeyOverride;
+        private static readonly Func<bool, ConsoleKeyInfo> _readKeyOverride;
         #pragma warning restore CS0649
 
-        private static Lazy<Func<bool, ConsoleKeyInfo>> _readKeyMethod = new Lazy<Func<bool, ConsoleKeyInfo>>(
+        private static readonly Lazy<Func<bool, ConsoleKeyInfo>> _readKeyMethod = new Lazy<Func<bool, ConsoleKeyInfo>>(
             () => _readKeyOverride == null ? Console.ReadKey : _readKeyOverride);
```
daxian-dbw, line 20: These 2 cannot have `readonly` added as they are actually a private contract between PSReadLine and PowerShellEditorService. They will be assigned via reflection in PSES.

PSReadLine/ReadLine.cs
```diff
@@ -46,16 +46,16 @@ public partial class PSConsoleReadLine : IPSConsoleReadLineMockableMethods
...
+        private readonly IConsole _console;
         private ICharMap _charMap;
         private Encoding _initialOutputEncoding;
         private bool _skipOutputEncodingChange;
         private EngineIntrinsics _engineIntrinsics;
         private Thread _readKeyThread;
         private AutoResetEvent _readKeyWaitHandle;
         private AutoResetEvent _keyReadWaitHandle;
-        private AutoResetEvent _forceEventWaitHandle;
+        private readonly AutoResetEvent _forceEventWaitHandle;
```
daxian-dbw, line 58: These 3 should be reverted. PSReadLine xUnit test needs to set `_mockableMethods` and `_console` via reflection, and PSES needs to set `_forceEventWaitHandle` via reflection.

daxian-dbw:
Set readonly field via reflection doesn't work in .NET Core 3.0 and later versions, see https://stackoverflow.com/a/934942:
> In dotnet core 3.0 this is no longer possible. A System.FieldAccessException is thrown saying: "Cannot set initonly static field 'bar' after type 'Foo' is initialized."

Given the feedback, I will close this PR in favor of #1984, which is based on changes in this PR.

xtqqczze:
> Set readonly field via reflection doesn't work in .NET Core 3.0 and later versions, see https://stackoverflow.com/a/934942:
> 
> > In dotnet core 3.0 this is no longer possible. A System.FieldAccessException is thrown saying: "Cannot set initonly static field 'bar' after type 'Foo' is initialized."

@daxian-dbw I compiled and ran the code from https://stackoverflow.com/a/934944 with .NET Core 3.0, .NET Core 3.1 and .NET 5.0 and didn't see the `System.FieldAccessException`.

```csharp
using System;
using System.Reflection;

public class Test
{
    private readonly string foo = "Foo";

    public static void Main()
    {
        Test test = new Test();
        FieldInfo field = typeof(Test).GetField
            ("foo", BindingFlags.Instance | BindingFlags.NonPublic);
        field.SetValue(test, "Hello");
        Console.WriteLine(test.foo);
    }        
}
```

According to [SetValue documentation](https://docs.microsoft.com/dotnet/api/system.reflection.fieldinfo.setvalue), if you attempt to set a value on a **static**, readonly field, an exception is thrown in .NET Core 3.0 and later.

See also:
* https://github.com/dotnet/runtime/issues/11571
* https://github.com/dotnet/dotnet-api-docs/issues/3851

PR #2012, opened 2020-12-03
[vi-mode] Correctly closing unbalanced start edit undo groups

### LINKED ISSUE #1281: InvalidOperationException thrown when typing `cffcc` in command mode
Environment data
----------------

```
PS version: 7.0.0-rc.1
PSReadline version: 2.0.0-rc1
PS file version: 7.0.0.0
HostName: ConsoleHost
BufferWidth: 145
BufferHeight: 9001
```

1. run `Set-PSReadLineOption -EditMode Vi`
2. type `abc`
3. type <kbd>Esc</kbd>
4. type `cffcc`, exception will be thrown

### Exception
```
System.InvalidOperationException: Operation is not valid due to the current state of the object.
   at Microsoft.PowerShell.PSConsoleReadLine.StartEditGroup()
   at Microsoft.PowerShell.PSConsoleReadLine.GroupUndoHelper.StartGroup(Action`2 instigator, Object instigatorArg)
   at Microsoft.PowerShell.PSConsoleReadLine.ViReplaceLine(Nullable`1 key, Object arg)
   at Microsoft.PowerShell.PSConsoleReadLine.ViChord(Nullable`1 key, Object arg)
```

springcomp:
# PR Summary

Fixes #1281.

Adds an `else` block in each occurrence where a `_groupUndoHelper.StartGroup()` is made but ultimately unbalanced (not ended) because no matching characters were found. As a result, the next time this operation is attempted PSRL no longer crashes due to not supporting nested undo groups.

Changes implementations of the following commands:

- <kbd>c</kbd>, <kbd>f</kbd>, …
- <kbd>c</kbd>, <kbd>F</kbd>, …
- <kbd>c</kbd>, <kbd>t</kbd>, …
- <kbd>c</kbd>, <kbd>T</kbd>, …

When authoring this PR I introduced a regression that boils down to the function `ViInsertWithAppend`.
This method does not start an undo group, and therefore, when undoing, only the last character appended is removed from the buffer.

Adding a call to `_groupUndoHelper.StartGroup()`, however, triggers a nested call, from one of the modified commands above as well as many other occasions, such as `ViAppendLine` for instance. One attempt would be to just ignore the nested call and simply pretend we are in the same, current group. But I propose another attempt which is to replace all the calls to `ViInsertWithAppend` which are done internally by a dedicated method that does the same, but without the call to `_groupUndoHelper.StartGroup()`.

Then I noticed that all the occurrences of calls to `ViInsertWithAppend` are in a block that looks like:

```
if (some condition)
{
  ViInsertWithAppend(key, arg);
}
else
{
  ViInsertMode(key, arg);
}
```

That’s why I propose a new internal method that does this pattern in place of each occurrence of this block in the code.

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not expecting documentation change.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/PowerShell/PSReadLine/pull/2012)


PSReadLine/Replace.vi.cs
```diff
@@ -151,16 +151,12 @@ private static void ViReplaceWord(ConsoleKeyInfo? key, object arg)
...
-            }
-            else
-            {
-                ViInsertMode(key, arg);
-            }
+
+            ViInsertModeAndForwardChar(
+                () => (_singleton._current == _singleton._buffer.Length - 1
+                    && !_singleton.IsDelimiter(_singleton._lastWordDelimiter, _singleton.Options.WordDelimiters)
+                    && _singleton._shouldAppend),
```
daxian-dbw: Why is the lambda needed? Why not calculate the bool result here and pass that in?

springcomp: Good catch, no real reason. I will refactor.

PSReadLine/ReadLine.vi.cs
```diff
@@ -611,8 +611,19 @@ public static void ViInsertAtEnd(ConsoleKeyInfo? key = null, object arg = null)
         /// </summary>
         public static void ViInsertWithAppend(ConsoleKeyInfo? key = null, object arg = null)
         {
+            _singleton._groupUndoHelper.StartGroup(ViInsertWithAppend, arg);
```
daxian-dbw, line 576: This is a public exposed method that can be accociated with any keybindings. If we start an editing group here, where will `EndGroup` be called?

I don't quite understand why `StartGroup` is added here. You mentioned in the PR descirption that:
> When authoring this PR I introduced a regression that boils down to the function ViInsertWithAppend.
**This method does not start an undo group, and therefore, when undoing, only the last character appended is removed from the buffer.**

Is the bold part an issue today without your changes? or is it as a result of the changes in this PR?

springcomp: In fact, the first commit introduces a regression. Closing a group if not finding a matching character in <kbd>c</kbd><kbd>f</kbd><kbd>?</kbd> _is_ the regression.

Indeed, this sequence is followed by a call to `ViInsertWithAppend` which relies on the group started in `ViReplaceToChar`.

On the one hand, not closing the group makes PSRL crash if the same sequence is retried (because nested groups are not supported). On the other hand, closing the group prevents _appends_ from recording the whole future sequence of inserts to undo.

So I'm not sure how best to solve this problem.

Perhaps we should not throw an exception when a group is started without closing the previous one. Just carrying along with recording sequences to undo ?

springcomp: @daxian-dbw also, I mostly did the change out of symmetry with the method `ViInsertWithDelete` which, itself, starts a group. So I think the same argument should apply.

If this is a breaking change, I’m convinced it is a _necessary_ change.
What do you think ?

daxian-dbw: Sorry for the looooonnnnggg delay in reviewing this PR :(
While re-reading the code changes, I think maybe the best way to solve this problem is to properly support nested groups. It's a way bigger change I guess but I just wonder maybe that's the right way forward if we want to eliminate all issues related to unbalanced edit groups ...

springcomp: OK, I started from scratch and ended up supporting multiple starts for one single end.
I think there can only be one edit group at any single time. It’s just that methods may be called in multiple orders.

This allows to support adding a start in the public facing method without impact.

I fear the that undo support is a bit brittle as it now needs a ton of states to track.
Please, let me know what you feel.

springcomp: The new change supports closing all the "starts" so that would no longer be an issue.

PSReadLine/UndoRedo.cs
```diff
@@ -59,19 +59,13 @@ private void SaveEditItem(EditItem editItem)
...
         }
 
-        private void StartEditGroup()
+        internal void StartEditGroup()
         {
-            if (_editGroupStart != -1)
-            {
-                // Nesting not supported.
-                throw new InvalidOperationException();
-            }
```
daxian-dbw: Without this check, for the edit groups started by calling `StartEditGroup` (without going through `GroupUndoHelper.StartGroup`), the current edit is not saved anywhere and will be lost. That doesn't seem right.

test/UndoRedoTest.cs
```diff
@@ -56,5 +58,31 @@ public void ProperlySetEditGroupStartMarkWhenLoopInHistory()
...
+                    PSConsoleReadLine._singleton.EndEditGroup();
+                })));
+
+            Test("012 45", Keys(
+                "012 45", _.Escape,
+                "b", // back one word
+                CheckThat(() =>
+                {
+                    PSConsoleReadLine._singleton.StartEditGroup();
+                    PSConsoleReadLine._singleton.StartEditGroup();
```
daxian-dbw: I don't understand the purpose of this test. The second `StartEditGroup` call will override the 1st one -- `_editGroupStart` set in the first call will be lost.

test/BasicEditingTest.VI.cs
```diff
@@ -1121,5 +1121,18 @@ public void ViInsertModeMoveCursor()
...
+        [SkippableFact]
+        public void ViDefect1281()
+        {
+            TestSetup(KeyMode.Vi);
+
+            Test("bcd", Keys(
+                "abcdabcd", _.Escape,
+                "Bc2tb", _.Escape, CheckThat(() => AssertCursorLeftIs(0)), 
+                "c2tb", CheckThat(() => AssertLineIs("bcd")),
+                "c2tb" // should not crash
```
daxian-dbw: Can you please add comment about what each input sequence does in VI mode? For example, a comment about what 
functions will be triggered by `Bc2tb`.

PSReadLine/UndoRedo.cs
daxian-dbw: If I understand correctly, `_groupUndoHelper` and `_groupUndoStates` are used only when in VI mode, and the starting nested edit group issue is only relevant in VI modes.
If so, then maybe we should only skip the `_editGroupStart != -1` check in `GroupUndoHelper.StartGroup`, but not skipping generally in `StartEditGroup`.

PSReadLine/UndoRedo.vi.cs
```diff
@@ -37,11 +47,50 @@ public void EndGroup()
                 if (_singleton._editGroupStart >= 0)
                 {
                     _singleton.EndEditGroup(_instigator, _instigatorArg);
+
+                    if (_singleton._groupUndoStates.Count > 0)
+                    {
+                        var groupUndoState = _singleton._groupUndoStates.Pop();
+                        _singleton._editGroupStart = groupUndoState.EditGroupStart;
+                        groupUndoState.GroupUndoHelper.EndGroup();
+                    }
```
daxian-dbw: Can you please change this to a loop instead of doing it recursively?

test/UndoRedoTest.cs
springcomp: Indeed, I have added comments to clarify.
The purpose of the pull request is to properly support multiple starts with a single end.

test/BasicEditingTest.VI.cs
springcomp: ✔

PSReadLine/UndoRedo.cs
springcomp: I have re-introduced a protection in the general `StartEditGroup()` method.

PSReadLine/UndoRedo.vi.cs
springcomp: I have refactored the code.

This manifested a mix of usage for the `GroupUndoHelper` class like so:

```c#
this.EndEditGroup(_instigator, _instigatorArg);
```

and
```c#
_singleton._groupUndoHelper.EndEditGroup(
  _singleton._groupUndoHelper._instigator,
  _singleton._groupUndoHelper._instigatorArg);
```

I figured to make the code consistent with a single style.

PSReadLine/UndoRedo.cs
daxian-dbw: > I have re-introduced a protection in the general StartEditGroup() method.

@springcomp the latest updates in the PR doesn't show a change in the `StartEditGroup()` method. Did you push all changes in this branch?

test/UndoRedoTest.cs
daxian-dbw: Same here -- I don't see the comments added here. Can you please check if you pushed all changes in this branch?

test/BasicEditingTest.VI.cs
daxian-dbw: Same here too :) command not added in the latest changes.

springcomp: Could not find the code – so that must have been my mistake.
Sorry for the confusion. 🤦‍♂️
This is pushed now.

test/UndoRedoTest.cs
springcomp: Sorry for the confusion.
I have pushed explanatory comments.

PSReadLine/UndoRedo.cs
springcomp: Sorry for the confusion.

Taking a step back and re-assessing the original feedback, I’m not actually in favor of the requested change.
My understanding is that this would allow nested (pending) edit groups in VI mode only, but still protected the general case from unwantingly starting nested groups.

I have pushed a last commit to address this feedback but this leads to code that is – in my opinion – not very elegant.

Please, let me know what you think.

PSReadLine/UndoRedo.vi.cs
```diff
@@ -3,24 +3,36 @@
...
+                    // A nested "start" of a group is being made, so we
+                    // need to record the state of the preceding start.
+                    _singleton._groupUndoStates.Push(
+                        new GroupUndoState(_singleton._groupUndoHelper, _singleton._editGroupStart));
+
+                    _singleton._editGroupStart = -1;
+                }
+
                 _instigator = instigator;
                 _instigatorArg = instigatorArg;
```
daxian-dbw, line 37: It doesn't seem to be right to override the existing `_instigator` and `_instigatorArg`. Consider `ViReplaceToChar` as an example, it calls `ViInsertWithAppend` internally, which also starts a group with your change, however, when redoing, it should be calling the instigator set by `ViReplaceToChar` instead of the one set by `ViInsertWithAppend`, because the use didn't explicitly trigger `ViInsertWithAppend`.

daxian-dbw (commented):
@springcomp Thank you so much for your patience.

I spent nearly 4 hours today reviewing the changes and thinking about what the right thing should be. Overall, I think allowing nested edit groups adds too much complexity and may make it more error-prone when writing code, because it will be harder to notice when a nested edit group was made by mistake. It also makes it harder to reason the code, because part of the code base assumes the edit group is always balanced while the other part of the code base allows unbalanced edit groups.

I wonder whether it's possible to make the VI mode works as expected while requiring balanced edit groups.

Take `ViReplaceToChar` as an example (_triggered by <kbd>cf</kbd> in command mode_), it calls `ViInsertWithAppend` (_trigged by <kbd>a</kbd> in command mode_) internally which also starts an edit group with your change in this PR (_expected behavior when it's explicitly triggered by user_). In this case, `ViInsertWithAppend` should be changed to the following, and `ViReplaceToChar` should call `ViInsertWithAppendImpl` instead.

```
  public static void ViInsertWithAppend(ConsoleKeyInfo? key = null, object arg = null)
  {
      _singleton._groupUndoHelper.StartGroup(ViInsertWithAppend, arg);
      ViInsertWithAppendImpl(key, arg);
  }

  private static void ViInsertWithAppendImpl(ConsoleKeyInfo? key = null, object arg = null)
  {
      ViInsertMode(key, arg);
      ForwardChar(key, arg);
  }
```

daxian-dbw:
@springcomp Given the concerns described in https://github.com/PowerShell/PSReadLine/pull/2012#pullrequestreview-1683512679 and that the PR #3845 was merged, I will close this one. But as always, I really appreciate your consistent contribution to PSRL!

Issue #1674, opened 2020-07-19
[vi-mode] Backwards delete removes the character under the cursor, which is incorrect


springcomp:
## Environments

PS version: 5.1.19041.1
PSReadline version: 2.0.0-beta2
os: 10.0.19041.322 (WinBuild.160101.0800)
PS file version: 10.0.19041.322 (WinBuild.160101.0800)
HostName: ConsoleHost (Windows Terminal)
BufferWidth: 120
BufferHeight: 30

PS version: 7.1.0-preview.2
PSReadline version: 2.0.1
os: 10.0.19041.322 (WinBuild.160101.0800)
PS file version: 7.1.0.0
HostName: ConsoleHost (Windows Terminal)
BufferWidth: 120
BufferHeight: 30

## Steps to reproduce

- Type the following sequence: <kbd>a</kbd><kbd>b</kbd><kbd>c</kbd><kbd>d</kbd><kbd>e</kbd><kbd>f</kbd><kbd>g</kbd><kbd>Esc</kbd>
Thus, typing `abcdefg` and positioning the cursor at the end of the line (on the `g` character).

- Then, type <kbd>d</kbd><kbd>F</kbd><kbd>a</kbd> to delete up to the first `a` character.

## Expected behavior

The entire word should be deleted, except the `g` character itself.

## Actual behavior

Notice that the whole word is deleted, thus leaving an empty buffer.

Issue #584, opened 2017-11-14
Add "text object" support for Vi mode


gwojan:
It would be really nice if Vi mode supported text objects so things like:

```
ci{
da{
ci[
ci"
ci'
etc...
```
were possible.

lzybkr:
Indeed - I might have mentioned this to @srdubya a long time ago. I might even switch to vi mode if this was implemented. (I use vim or vim bindings in my editors.)

gwojan:
Yeah, my workaround when I really want to get nasty is set ```$env:VISUAL = 'vim.exe'``` and then edit the whole thing inside vim proper...

I'm even using the ```wasavi``` extension for chrome and editing this comment inside a vim-like editor. Yes, I'm sad... 😉

springcomp:
I’m working on this issue.

This will probably need to be broken up into several pull requests.
This is how I see tackling this subject in order:

- [x] _n_ <kbd>d</kbd><kbd>i</kbd><kbd>w</kbd>: deletes inner words. #2059. 🍾
- [ ] _n_ <kbd>c</kbd><kbd>i</kbd><kbd>w</kbd>: changes inner words.
- [ ] _n_ <kbd>d</kbd><kbd>i</kbd><kbd>W</kbd>: deletes inner WORDs.
- [ ] _n_ <kbd>c</kbd><kbd>i</kbd><kbd>W</kbd>: changes inner WORDs.
- [x] <kbd>d</kbd><kbd>i</kbd> [ <kbd>"</kbd> | <kbd>'</kbd> ]: deletes inner strings. #3791.
- [ ] <kbd>c</kbd><kbd>i</kbd> [ <kbd>"</kbd> | <kbd>'</kbd> ]: changes inner strings.
- [ ] <kbd>d</kbd><kbd>i</kbd> [ <kbd>(</kbd> | <kbd>{</kbd> | <kbd>[</kbd> ]: deletes inner parentheses. 
- [ ] <kbd>c</kbd><kbd>i</kbd> [ <kbd>(</kbd> | <kbd>{</kbd> | <kbd>[</kbd> ]: changes inner parentheses. 

- [ ] _n_ <kbd>d</kbd><kbd>a</kbd><kbd>w</kbd>: deletes outer words.
- [ ] _n_ <kbd>c</kbd><kbd>a</kbd><kbd>w</kbd>: changes outer words.
[...]

PR #2022, opened 2020-12-06
Refactor usage of clipboard

springcomp:
# PR Summary

While working on various upcoming PRs (notably #1674 and #584) I noticed this would be easier if some code cleanup/refactoring had taken place before,

I noticed a common pattern while trying to remove some text while supporting undo:

```
_singleton.SaveToClipboard(start, count, instigator, arg);
_singleton.SaveEditItem(start, count, EditItemCreate(_clipboard, start, instigator, arg));
_singleton.Remove(position, length);
```

Therefore, this PR introduces a new method to promote this pattern and refactors all ocurrences to use the new method.
While doing so, I also slightly refactor one or two obvious related changes to the clipboard handling.

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/PowerShell/PSReadLine/pull/2022)


PSReadLine/YankPaste.vi.cs
```diff
@@ -70,6 +70,26 @@ private void SaveLinesToClipboard(int lineIndex, int lineCount)
...
 
+        /// <summary>
+        /// Removes a portion of text from the buffer
+        /// and saves it to the clipboard in order to support undo.
+        /// </summary>
+        /// <param name="start"></param>
+        /// <param name="count"></param>
+        /// <param name="instigator"></param>
+        /// <param name="arg"></param>
+        private void RemoveTextToClipboard(int start, int count, Action<ConsoleKeyInfo?, object> instigator = null, object arg = null)
```
daxian-dbw: `_clipboard` is not a good variable name IMO. It's actually VI specific, but the name is too general.
How about we rename this method to be `RemoveTextAndSaveToViRegister`?

PSReadLine/YankPaste.vi.cs
```diff
@@ -70,6 +70,26 @@ private void SaveLinesToClipboard(int lineIndex, int lineCount)
             _clipboard.LinewiseRecord(_buffer.ToString(range.Offset, range.Count));
         }
 
+        /// <summary>
+        /// Removes a portion of text from the buffer
+        /// and saves it to the clipboard in order to support undo.
```
daxian-dbw: I think the undo is supported by `_singleton.SaveEditItem`, not `_singleton.SaveToClipboard`. Maybe update the comment to:
```suggestion
        /// Remove a portion of text from the buffer,  save it to the vi register,
        /// and also save it to the edit list to support undo.
```

PSReadLine/YankPaste.vi.cs
```diff
@@ -70,6 +70,26 @@ private void SaveLinesToClipboard(int lineIndex, int lineCount)
...
+        /// </summary>
+        /// <param name="start"></param>
+        /// <param name="count"></param>
+        /// <param name="instigator"></param>
+        /// <param name="arg"></param>
+        private void RemoveTextToClipboard(int start, int count, Action<ConsoleKeyInfo?, object> instigator = null, object arg = null)
+        {
+            _singleton.SaveToClipboard(start, count);
+            _singleton.SaveEditItem(EditItemDelete.Create(
+                _clipboard,
```
daxian-dbw: Please use `_clipboard.RawText` here to improve the readability. It's not obvious that `ViRegister` implements an implicit converter to string, so it would cost more time to someone who is reading this code to realize that.

PSReadLine/YankPaste.vi.cs
```diff
@@ -70,6 +70,26 @@ private void SaveLinesToClipboard(int lineIndex, int lineCount)
...
+                start,
+                instigator,
+                arg
+                ));
```
daxian-dbw: nitpick:
```suggestion
                arg));
```

daxian-dbw (changes requested):
Thanks for the refactoring, look mostly good to me, with a few comments.

PSReadLine/YankPaste.vi.cs
springcomp: Thanks for the suggestion.

As a result, and following changes made in this PR, the implicit converter is no longer needed and is therefore removed.

PSReadLine/YankPaste.vi.cs
springcomp: I chose `RemoveTextToViRegister` if that is OK for you.

springcomp: I also introduced an extra commit to rename `_clipboard` to `_viRegister`.
Please, let me know if you want this change or not.

PSReadLine/BasicEditing.cs
```diff
@@ -178,9 +169,7 @@ private void DeleteCharImpl(int qty, bool orExit)
                 {
                     qty = Math.Min(qty, _singleton._buffer.Length - _singleton._current);
 
-                    SaveEditItem(EditItemDelete.Create(_buffer.ToString(_current, qty), _current, DeleteChar, qty));
-                    SaveToClipboard(_current, qty);
-                    _buffer.Remove(_current, qty);
+                    RemoveTextToClipboard(_current, qty, DeleteChar);
```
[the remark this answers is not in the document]
springcomp: Done.

springcomp:
@daxian-dbw I believe I have addressed all feedback.
Please, would you re-review?

daxian-dbw (approved):
LGTM. I will help resolve the conflict.

PR #2045, opened 2020-12-16
[vi-mode] Undo now leaves the cursor under the position at the start of the deletion

### LINKED ISSUE #1673: [vi-mode] Undoing a delete moves the cursor at the end of the newly pasted text.
## Environments

PS version: 5.1.19041.1
os: 10.0.19041.322 (WinBuild.160101.0800)
PS file version: 10.0.19041.322 (WinBuild.160101.0800)
BufferWidth: 120
BufferHeight: 30

PS version: 7.1.0-preview.2
PSReadline version: 2.0.1
os: 10.0.19041.322 (WinBuild.160101.0800)
PS file version: 7.1.0.0
BufferWidth: 120
BufferHeight: 30

## Steps to reproduce

- Type the following sequence: <kbd>o</kbd><kbd>n</kbd><kbd>e</kbd><kbd>Esc</kbd><kbd>0</kbd>
Thus, typing `one` and positioning the cursor at the beginning of the line.

- Then, type <kbd>x</kbd> to delete one character.
- Finally, type <kbd>u</kbd>to undo.

## Expected behavior

The cursor should stay at the beginning of the line. In fact, the cursor should stay at the start position of the deleted text.

## Actual behavior

Notice that the cursor is now on the `n` letter, instead of the `o` letter, at the beginning of the line.

### LINKED ISSUE #2408: Prediction Feature is broken for current line whenever the S command of vi EditMode is used (to clear line)
Searched for anything related to S command of vi EditMode breaking Prediction but didn't find anything. Sorry it that's a duplicate even so.

## Environment
I've reproduced the error on both environments below, both with the native PSReadline version (2.1, I believe) and after upgrading it to the beta version:

PS version: 7.1.3
PSReadline version: 2.2.0-beta2
os: 10.0.19041.320 (WinBuild.160101.0800)
PS file version: 7.1.3.0
BufferWidth: 237
BufferHeight: 61

PS version: 7.1.3
PSReadline version: 2.2.0-beta2
os: Linux ABELHA1674 4.19.128-microsoft-standard #1 SMP Tue Jun 23 12:58:10 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
PS file version: 7.1.3.0
HostName: ConsoleHost
BufferWidth: 237
BufferHeight: 61

## Exception report

No exception, I'm afraid.

## Steps to reproduce

Set VI Edit mode:
``` powershell
Set-PSReadlineOption -EditMode vi
```
Set History as PredictionSource
``` powershell
Set-PSReadLineOption -PredictionSource History
```
Write anything, then press ESC and S (to clear whole line)
Try to use prediction on the cleared line (such as typing Set-PSRead and pressing righ to accept suggestion)

## Expected behavior

The suggestion is accepted just as it happens on a new line (without using the S to clear the line before)

## Actual behavior

Instead of accepting the suggestion it appers to not do anything at all (visually) and if you stay on the edit mode it really does nothing. But if you press right (to accept the suggestion), then ESC to go to normal mode it skips to the end of the line (to where it would go if it had accepted the suggestion) without visually including the suggestion and if you press enter it will run the suggestion. If you type right then ESC then start typing again, eventually it updates visually as well (including the suggestion and what you just typed).

If you discard or run that line the next line works as expected as long as you don't run the "S" command again - so for now my workaround has been to use "S" followed by Enter to create new line (which is quite a pain).

## Summary

The Prediction feature breaks whenever you clear the line (at least with the S command of vi EditMode). Creating new line whenever that happens seems to "fix" it.

springcomp:
# PR Summary

Fixes #1673
Fixes #2408

This PR changes undo so that a deleted text is inserted in the buffer and resetting the cursor at the position where it was when deletion of text occurred. I believe this is also making it more consistent with how most text editors behave.

To implement this PR I also slightly changed the `AcceptSuggestion` and `AcceptNextSuggestionWord` methods which were doing a replacement of the buffer. This PR now makes those methods add the appropriate portion of text so that undoing the operation behaves more inline with most users’ expectations. Externally, the behaviour of those methods has not changed.

This PR also changes `SwapCharacters` to make it an _atomic_ operation while still moving the cursor after swapping characters. This preserves the original behaviour. However, it slightly alters the way the cursor is now positioned when undo occurs, but I’m not sure whether this can be seen as a breaking change or not.

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/PowerShell/PSReadLine/pull/2045)

### DIFF
--- a/PSReadLine/Prediction.cs
+++ b/PSReadLine/Prediction.cs
@@ -69,13 +69,17 @@ namespace Microsoft.PowerShell
             Prediction prediction = _singleton._prediction;
             if (prediction.ActiveView is PredictionInlineView inlineView && inlineView.HasActiveSuggestion)
             {
+                var start = inlineView.InputText.Length;
+
                 // Ignore the visual selection.
                 _singleton._visualSelectionCommandCount = 0;
 
                 inlineView.OnSuggestionAccepted();
 
                 using var _ = prediction.DisableScoped();
-                Replace(0, _singleton._buffer.Length, inlineView.SuggestionText);
+
+                _singleton._current = start;
+                Insert(inlineView.SuggestionText.Substring(_singleton._current));
             }
         }
 
@@ -102,17 +106,21 @@ namespace Microsoft.PowerShell
         {
             if (_singleton._prediction.ActiveView is PredictionInlineView inlineView && inlineView.HasActiveSuggestion)
             {
+                var start = Math.Max(0, inlineView.InputText.Length);
+                var index = _singleton._buffer.Length;
+
                 // Ignore the visual selection.
                 _singleton._visualSelectionCommandCount = 0;
 
-                int index = _singleton._buffer.Length;
                 while (numericArg-- > 0 && index < inlineView.SuggestionText.Length)
                 {
                     index = inlineView.FindForwardSuggestionWordPoint(index, _singleton.Options.WordDelimiters);
                 }
 
                 inlineView.OnSuggestionAccepted();
-                Replace(0, _singleton._buffer.Length, inlineView.SuggestionText.Substring(0, index));
+
+                _singleton._current = start;
+                Insert(inlineView.SuggestionText.Substring(start, index - start));
             }
         }
 
--- a/PSReadLine/Prediction.cs
+++ b/PSReadLine/Prediction.cs
@@ -69,8 +69,6 @@ namespace Microsoft.PowerShell
             Prediction prediction = _singleton._prediction;
             if (prediction.ActiveView is PredictionInlineView inlineView && inlineView.HasActiveSuggestion)
             {
-                var start = inlineView.InputText.Length;
-
                 // Ignore the visual selection.
                 _singleton._visualSelectionCommandCount = 0;
 
@@ -78,7 +76,7 @@ namespace Microsoft.PowerShell
 
                 using var _ = prediction.DisableScoped();
 
-                _singleton._current = start;
+                _singleton._current = _singleton._buffer.Length;
                 Insert(inlineView.SuggestionText.Substring(_singleton._current));
             }
         }
@@ -106,12 +104,11 @@ namespace Microsoft.PowerShell
         {
             if (_singleton._prediction.ActiveView is PredictionInlineView inlineView && inlineView.HasActiveSuggestion)
             {
-                var start = Math.Max(0, inlineView.InputText.Length);
-                var index = _singleton._buffer.Length;
-
                 // Ignore the visual selection.
                 _singleton._visualSelectionCommandCount = 0;
 
+                int start = _singleton._buffer.Length;
+                int index = start;
                 while (numericArg-- > 0 && index < inlineView.SuggestionText.Length)
                 {
                     index = inlineView.FindForwardSuggestionWordPoint(index, _singleton.Options.WordDelimiters);
--- a/PSReadLine/UndoRedo.cs
+++ b/PSReadLine/UndoRedo.cs
@@ -258,45 +258,40 @@ namespace Microsoft.PowerShell
             private readonly string _deletedString;
             private readonly int _deleteStartPosition;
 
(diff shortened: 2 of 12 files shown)


PSReadLine/UndoRedo.cs
```diff
@@ -278,7 +278,7 @@ public static EditItem Create(string str, int position, Action<ConsoleKeyInfo?,
             public override void Undo()
             {
                 _singleton._buffer.Insert(_deleteStartPosition, _deletedString);
-                _singleton._current = _deleteStartPosition + _deletedString.Length;
+                _singleton._current = _deleteStartPosition;
```
daxian-dbw: This change doesn't seem right for some basic text-deleting scenario, for example, in "Windows" edit mode:

_Before the change_, selecting part of text from left to right, then "Delete", then "Ctrl+z"
demo1

Or, selecting from right to left, then "Delete", then "Ctrl+z"
demo1

_After the change_, the cursor will always at <kbd>c</kbd> after "Ctrl+z"

This is a pretty noticeable breaking change. Personally, I'm very used to having "Ctrl+z" to place the curosr at the end for scenarios like the above 2 examples, especially the 2nd one.

------------

Maybe, what we need is something like `_insertAnchor` in `EditItemInsertLines`, to instruct where `EditItemDelete.Undo` should place the cursort at. Maybe just a `bool` field to indicate whether the cursor should be placed at the beginning position of the deleted text, or the end of it (default should be the end so no breaking changes).

daxian-dbw (changes requested):

PSReadLine/Prediction.cs
```diff
@@ -45,13 +45,17 @@ public static void AcceptSuggestion(ConsoleKeyInfo? key = null, object arg = nul
...
                 // Ignore the visual selection.
                 _singleton._visualSelectionCommandCount = 0;
 
                 inlineView.OnSuggestionAccepted();
 
                 using var _ = prediction.DisableScoped();
-                Replace(0, _singleton._buffer.Length, inlineView.SuggestionText);
+
+                _singleton._current = start;
+                Insert(inlineView.SuggestionText.Substring(_singleton._current));
```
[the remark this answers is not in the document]
springcomp: As alluded to in the PR description, I also slightly changed the `AcceptSuggestion` and `AcceptNextSuggestionWord` methods which were doing a replacement of the buffer. This PR now makes those methods add the appropriate portion of text so that undoing the operation behaves more inline with most users’ expectations. Externally, the behaviour of those methods has not changed.

It simplifies the code and has the same external behaviour while not using a delete operation at all. This therefore makes it invariant in the face of whether to customize the undo mechanism with `EditItemDelete` or not in other parts of the code.

PSReadLine/UndoRedo.cs
springcomp: I’ve just added a `bool adjustCursor` flag as you suggested to the `EditItemDelete.Create` method.
This flag defaults to `true` meaning that the cursor moves at the end of the re-inserted text after undoing a delete operation.

In VI-mode specific files, such as `ReadLine.vi.cs` and `Replace.vi.cs`, this flag is explicitely overridden and set to `false`.

In other parts of the code, where the operation is shared between Vi-mode and Windows-mode, this flag is computed using the expression `Options.EditMode != EditMode.Vi` which keeps the behaviour unchanged for Windows (an potentially Emacs) mode.

Finally, a call in `Kill` has not been changed, thus taking advantage of the default value to keep the existing behaviour unchanged for Emacs-mode.

daxian-dbw (approved):
@springcomp Sorry for the looooong delay. I rebased your code and made some refactoring and updates. Will merge this PR shortly.

springcomp:
@daxian-dbw  thanks a lot 🙏. I never lost hope 😉.

ghost:

PR #2934, opened 2021-10-27
Generate `OnIdle` event only if the editing buffer is empty

daxian-dbw:
# PR Summary

Generate `OnIdle` event only if the editing buffer is empty. But if there is any other event subscribers, we still run the dummy pipeline to allow potential event handling.

`OnIdle` event by definition is not an event that need urgent processing, and sometimes it's used by module/script to output a message, e.g. `Az.Predictor` uses it to write out a ask-for-survey message. Having this event to fire when the user already input something in the buffer is not a good experience for the user. See the following screenshot as an example.

image

With this PR, we generate the `OnIdle` event only if the editing buffer is empty. But for any other event subscribers, we still run the dummy pipeline to allow potential event processing for them even if the editing buffer is not empty.

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [x] Doc Issue filed: https://github.com/MicrosoftDocs/PowerShell-Docs/issues/8270

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/2934)

### DIFF
--- a/PSReadLine/ReadLine.cs
+++ b/PSReadLine/ReadLine.cs
@@ -218,18 +218,27 @@ namespace Microsoft.PowerShell
                         bool runPipelineForEventProcessing = false;
                         foreach (var sub in eventSubscribers)
                         {
-                            runPipelineForEventProcessing = true;
                             if (sub.SourceIdentifier.Equals(PSEngineEvent.OnIdle, StringComparison.OrdinalIgnoreCase))
                             {
-                                // There is an OnIdle event.  We're idle because we timed out.  Normally
+                                // If the buffer is not empty, let's not consider we are idle because the user
+                                // is in the middle of typing something.
+                                if (_singleton._buffer.Length > 0)
+                                {
+                                    continue;
+                                }
+
+                                // There is an OnIdle event. We timed out and the buffer is empty.  Normally
                                 // PowerShell generates this event, but PowerShell assumes the engine is not
                                 // idle because it called PSConsoleHostReadLine which isn't returning.
                                 // So we generate the event instead.
+                                runPipelineForEventProcessing = true;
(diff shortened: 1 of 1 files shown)


ghost:

lzybkr:
I understand the issue you are trying to address, I'm not sure this is the right fix. For example, without PSReadLine, the experience is still ugly.

I think it'd be better to find a better solution than using OnIdle for any sort of UI experience.

daxian-dbw:
Right, totally agree that using `OnIdle` event for UI experience is far from desired. The good news is that `Az.Predictor` depends on PSReadLine to work :)

lzybkr:
Fine, but this change will probably affect people not using `Az.Predictor`. People will randomly see their `OnIdle` events not firing and not understand the root cause, or waste a lot of time tracking down it down.

daxian-dbw:
Yes, that's absolutely possible. But I think the impact should be quite low, as `OnIdle` event will be fired when there's a timeout and the buffer is empty, and that situation happens a lot. I will keep an eye on the feedback, and I can revert this change if there is a scenario that got blocked by this change.

daxian-dbw:
I also opened the doc issue https://github.com/MicrosoftDocs/PowerShell-Docs/issues/8270 to have this behavior documented (not now, but I will submit a doc PR after a while if there is no feedback about this change).

lzybkr:
I would also mention it as a low risk breaking change in the release notes for better visibility.

daxian-dbw:
Done, updated the GitHub release page: https://github.com/PowerShell/PSReadLine/releases/tag/v2.2.0-beta4

PR #3039, opened 2021-11-15
Override Console.WriteLine

dkattan:
# PR Summary

Added overridable hooks for 
  - Console.Write
  - Console.WriteLine
  - Console.SetWindowPosition
  - Console.SetCursorPosition

This is useful in situations where you want to provide your own handler for output, like a web terminal.

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3039)

### DIFF
--- a/PSReadLine/ConsoleLib.cs
+++ b/PSReadLine/ConsoleLib.cs
@@ -12,12 +12,40 @@ namespace Microsoft.PowerShell.Internal
         // These two fields are used by PowerShellEditorServices to inject a
         // custom ReadKey implementation. This is not a public API, but it is
         // part of a private contract with that project.
-        #pragma warning disable CS0649
+#pragma warning disable CS0649
         private static Func<bool, ConsoleKeyInfo> _readKeyOverride;
-        #pragma warning restore CS0649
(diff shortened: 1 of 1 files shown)


daxian-dbw:
@dkattan What you are trying to do is to define private contracts that can only be accessed through reflection. I don't think that's what you really want. The existing `_readKeyOverride` is a very special thing that PowerShell VSCode extension requires to work, and we are not going to accept other private contracts.

If you really need to override all those methods, you would be better to implement the [interface IConsole](https://github.com/PowerShell/PSReadLine/blob/master/PSReadLine/PublicAPI.cs#L39), and then replace the console used by PSReadLine with [code like this](https://github.com/PowerShell/PSReadLine/blob/9c816be20f23c7a11056ed0b940d52f09b942b14/test/UnitTestReadLine.cs#L551-L558).

PR #3078, opened 2021-12-02
Fix control symbols when changing layout

### LINKED ISSUE #2865: Cyrillic breaks CTRL+ combos
## Environment

```none
PS version: 7.1.4
PSReadline version: 2.1.0
os: 10.0.19041.1165 (WinBuild.160101.0800)
PS file version: 7.1.4.0
HostName: ConsoleHost
BufferWidth: 120
BufferHeight: 9001
```

## Steps to reproduce

Open powershell 7.1.4 in windows terminal.
In cyrillic keymap press CTRL+С (C cyrillic)
CTRL not working and terminal show C
CTRL+A and CTRL+C not working after it in both keymaps

There is also another cases but it hard to reproduce.

## Actual behavior

CTRL+C working correctly

243083df:
# PR Summary
toUnicode sticks with first encoutered locale. During some research i use toUnicodeEx and provide current used locale to it.
Resolve #2865
## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3078)


PSReadLine/Keys.cs
```diff
@@ -147,7 +157,8 @@ internal static void TryGetCharFromConsoleKey(ConsoleKeyInfo key, ref char resul
             {
                 flags |= (1 << 2); /* If bit 2 is set, keyboard state is not changed (Windows 10, version 1607 and newer) */
             }
-            int charCount = ToUnicode(virtualKey, scanCode, state, chars, chars.Length, flags);
+            uint layout = GetKeyboardLayout(GetWindowThreadProcessId(GetForegroundWindow(), 0));
```
lzybkr: This always felt like a bug in console applications. I believe the conhost process receives the keyboard layout change notifications, but the `ToUnicode` call happens in a different process that never sees layout change notifications. I may have even opened a bug on `ToUnicode` that might have been resolved as won't fix over concerns about breaking something.

This change may seem work well enough in practice, but there is a race condition:

* Type a character
* New popup window gets created by some background process
* Call to `GetForegroundWindow` sees this new window instead of the console/vscode/etc. window.

I'm also not excited about seeing `GetWindowThreadProcessId` called on every key - it seems like it should be cached. Or maybe better - can we find the conhost/openconsole process instead? That process should see the correct layout, right?

243083df: @lzybkr I create replacement for GetForegroundWindow.  What do you think about it?
```
// aka GetForegroundWindow
static Process findParentWithWindowHandle()
        {
            var current = Process.GetCurrentProcess();
            while (current != null)
            {
                if (current.MainWindowHandle != IntPtr.Zero)
                {
                    break;
                }
                current = GetParentProcess(current.Handle);
            }
            return current;
        }

 static Process GetParentProcess(IntPtr handle)
        {
            ProcessInformation processInformation = new ProcessInformation();

            int returnLength;
            int status = NtQueryInformationProcess(handle, 0, ref processInformation, Marshal.SizeOf(processInformation), out returnLength);
            if (status != 0)
                throw new Win32Exception(status);

            try
            {
                return Process.GetProcessById(processInformation.InheritedFromUniqueProcessId.ToInt32());
            }
            catch (ArgumentException)
            {
                // not found
                return null;
            }
        }
        [StructLayout(LayoutKind.Sequential)]
        public struct ProcessInformation
        {
            internal IntPtr Reserved1;
            internal IntPtr PebBaseAddress;
            internal IntPtr Reserved2_0;
            internal IntPtr Reserved2_1;
            internal IntPtr UniqueProcessId;
            internal IntPtr InheritedFromUniqueProcessId;

        }
```

lzybkr: That looks like it would work well in Windows Terminal, VSCode, and the ISE, but it won't work with conhost because the parent will be explorer. conhost.exe is a sibling process.

lzybkr: I wonder if you can use something like [GetProcessHandle](https://docs.microsoft.com/en-us/windows/win32/winauto/getprocesshandlefromhwnd)([GetConsoleWindow](https://docs.microsoft.com/en-us/windows/console/getconsolewindow)()) to get the correct conhost process - I expect this wouldn't work in Windows Terminal.

243083df: I research this theme deeper and find out that all approaches that find window that subsribed to locale change does not work in all situations.

243083df: I found that replacing https://github.com/PowerShell/PSReadLine/pull/3078/files#diff-9ab1ecb1e5d0ebd879a483eada4f6fcf2b407c662752f52bea29e45fd143f1a5R242
to c == '\0' do not break test.
And so far i didn't find dead key what start with different from '\0' code. Can it help?

243083df:
This is all what can i do. Better solution immposible with my skills.

lzybkr:
@DHowett - maybe you have some insights here?

DHowett:
Adding this to my list for tomorrow. My team is starting talent reviews, so I can't promise that I'll get to it timely.

243083df:
any update?

243083df:
bump

243083df:
How can i build custom verion of PSReadLine and install to PS?

kun1ck:
> How can i build custom verion of PSReadLine and install to PS?

Hello, is there a working solution now?

ForNeVeR:
@lzybkr @DHowett, sorry, folks, but could someone of you please take a look at this? Or direct us to someone who can review this at the moment, please. We're looking for feedback on this PR.

243083df:
> Hello, is there a working solution now?
This is fine to me, but i dont know how to install it to system

kun1ck:
> Adding this to my list for tomorrow. My team is starting talent reviews, so I can't promise that I'll get to it timely.

Hello, how soon can you check it out?

daxian-dbw:
Superseded by #3786. See https://github.com/PowerShell/PSReadLine/pull/3786#discussion_r1367402659 for the proposal to solve the two concerns raised in https://github.com/PowerShell/PSReadLine/pull/3078#discussion_r761562034.

PR #3274, opened 2022-04-14
Respect cancellation in `ReadOneOrMoreKeys()`

### LINKED ISSUE #1754: Resolve cancellation race with PSReadLine
When canceling the `ReadKey` method sent to PSReadLine, we observed that there's a race where PSReadLine is still consuming (and using) the dummy character we're returning when the method is canceled:

https://github.com/PowerShell/PowerShellEditorServices/blob/312051bbdaab5953f419c4f115ab4f322c512856/src/PowerShellEditorServices/Services/PowerShell/Host/PsesInternalHost.cs#L878-L908

While the chosen dummy character _should_ be a no-op, as observed in https://github.com/PowerShell/vscode-powershell/issues/3225#issuecomment-1097590406 it can come through as an `à` and in https://github.com/PowerShell/vscode-powershell/issues/3881#issuecomment-1097616974 it comes through as an `"`.

@daxian-dbw, @SeeminglyScience and I discovered this when trying to just use `default` for the returned value, and we think that around [here](https://github.com/PowerShell/PSReadLine/blob/f46f15d2d634e2060bc0eabe4c81fc13a5a64a3a/PSReadLine/ReadLine.cs#L97) "PSRL is probably checking if key returned, *then* checking if cancelled. Meaning if we have a key it doesn't check if cancelled."

andyleejordan:
Because .NET's . `Console.ReadKey()` is uncancellable, when a hosting application cancels PSReadLine, it also has to send a key press to get `ReadKey` to return. In this case, we want to ignore the key press (and a user certainly would not be expecting it, as `ReadLine` has already been canceled).

This fixes https://github.com/PowerShell/PowerShellEditorServices/issues/1754, at least as tested on Windows and macOS with PowerShell 7.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3274)

### DIFF
--- a/PSReadLine/ReadLine.cs
+++ b/PSReadLine/ReadLine.cs
@@ -91,7 +91,7 @@ namespace Microsoft.PowerShell
             return runspace?.ConnectionInfo != null;
         }
 
-        private void ReadOneOrMoreKeys()
+        private void ReadOneOrMoreKeys(CancellationToken cancellationToken)
         {
             _readkeyStopwatch.Restart();
             while (_console.KeyAvailable)
@@ -144,6 +144,14 @@ namespace Microsoft.PowerShell
                 while (_charMap.KeyAvailable)
                 {
                     var key = PSKeyInfo.FromConsoleKeyInfo(_charMap.ReadKey());
+                    if (cancellationToken.IsCancellationRequested)
+                    {
+                        // If PSReadLine is running under a host that can cancel it, the
+                        // cancellation will come at a time when ReadKey is stuck waiting for input.
+                        // The next key press will be used to force it to return, and so we want to
+                        // discard this key since we were already canceled.
+                        continue;
+                    }
                     _lastNKeys.Enqueue(key);
                     _queuedKeys.Enqueue(key);
                 }
@@ -160,7 +168,7 @@ namespace Microsoft.PowerShell
                     break;
 
                 var localCancellationToken = _singleton._cancelReadCancellationToken;
-                ReadOneOrMoreKeys();
+                ReadOneOrMoreKeys(localCancellationToken);
                 if (localCancellationToken.IsCancellationRequested)
                 {
                     continue;


andyleejordan:
@daxian-dbw do you know of any scenarios where `ReadLine(...)` gets canceled? PSES sends a cancellation token when we hook into it, and so cancel it, but...does anyone else?

Like, this is the entry point for PowerShell itself:

https://github.com/PowerShell/PSReadLine/blob/f46f15d2d634e2060bc0eabe4c81fc13a5a64a3a/PSReadLine/ReadLine.cs#L318-L328

And `_defaultCancellationToken` is:

https://github.com/PowerShell/PSReadLine/blob/f46f15d2d634e2060bc0eabe4c81fc13a5a64a3a/PSReadLine/ReadLine.cs#L43

Which doesn't look like it ever gets canceled (except maybe on disposal). (Also, why isn't it just literally `default` which creates a `CancellationToken.None`?)

Soo with the note here:

https://github.com/PowerShell/PSReadLine/blob/f46f15d2d634e2060bc0eabe4c81fc13a5a64a3a/PSReadLine/ReadLine.cs#L330-L339

This makes me think that PSES is quite possibly the _only_ "custom PSHost implementations that require the ability to cancel ReadLine." Which would mean this is quite a safe change for a v2.2.3?

/cc @SeeminglyScience

SeeminglyScience:
> (Also, why isn't it just literally `default` which creates a `CancellationToken.None`?)

Basically because `CancellationToken.None` has a global `WaitHandle` so if anyone ever does `CancellationToken.None.WaitHandle.Set()` PSRL would have a real rough time.

> This makes me think that PSES is quite possibly the _only_ "custom PSHost implementations that require the ability to cancel ReadLine." Which would mean this is quite a safe change for a v2.2.3?

It's the only one I know of (I mean, it's why I added the API in the first place) but it was purposefully designed as a public API as I can imagine other scenarios. That said, it's pretty unlikely anyone else is using it unless @daxian-dbw happens to know of someone.

PSReadLine/ReadLine.cs
```diff
@@ -144,6 +144,14 @@ private void ReadOneOrMoreKeys()
                 while (_charMap.KeyAvailable)
                 {
                     var key = PSKeyInfo.FromConsoleKeyInfo(_charMap.ReadKey());
+                    // If PSReadLine is running under a host that can cancel it, the cancellation
+                    // will come at a time when ReadKey is stuck waiting for input. The next key
+                    // press will be used to force it to return, and so we want to discard this key
+                    // since we were already canceled.
+                    if (cancellationToken.IsCancellationRequested)
+                    {
+                        continue;
```
andyleejordan, line 153: Perhaps this should be `break`?

SeeminglyScience: > Perhaps this should be `break`?

I was debating leaving that feedback, but I think maybe it *should* drain? If the key press is translated as an escape sequence (like an arrow key with VTINPUT/non-Windows) then we don't want to abort mid-way through processing the sequence, right?

andyleejordan: I guess it's possible, since .NET does what it wants with the key press we send. And yes, PSES is sending `\0` but that's not a guarantee necessarily.

SeeminglyScience: I'm more imagining that the only place that `break` vs `continue` would matter here is if the pending key press was an escape sequence. In this scenario I'd guess that it's a race condition between a cancellation happening and a key press.

In that scenario if we `break` instead we'll get the remainder of the escape sequence and `\0` next `ReadLine` invocation (if I understand the code correctly)

daxian-dbw: Yeah, I guess ~`break`~ `continue` would be fine.

PSReadLine/ReadLine.cs
```diff
@@ -144,6 +144,14 @@ private void ReadOneOrMoreKeys()
                 while (_charMap.KeyAvailable)
                 {
                     var key = PSKeyInfo.FromConsoleKeyInfo(_charMap.ReadKey());
+                    // If PSReadLine is running under a host that can cancel it, the cancellation
+                    // will come at a time when ReadKey is stuck waiting for input. The next key
+                    // press will be used to force it to return, and so we want to discard this key
+                    // since we were already canceled.
+                    if (cancellationToken.IsCancellationRequested)
```
daxian-dbw, line 147: Minor comment: maybe update this code a little bit to the following:
```
ConsoleKeyInfo kyeInfo = _charMap.ReadKey();
if (cancellationToken.IsCancellationRequested)
{
    // If PSReadLine is running under a host that can cancel it, the cancellation
    // will come at a time when ReadKey is stuck waiting for input. The next key
    // press will be used to force it to return, and so we want to discard this key
    // since we were already canceled.
    continue;
}

var key = PSKeyInfo.FromConsoleKeyInfo(kyeInfo);
_lastNKeys.Enqueue(key);
_queuedKeys.Enqueue(key);
```

daxian-dbw (approved):
LGTM, with a couple comments.

daxian-dbw:
> This makes me think that PSES is quite possibly the only "custom PSHost implementations that require the ability to cancel ReadLine." Which would mean this is quite a safe change for a v2.2.3?

I don't know of any other scenario where the `ReadLine` API with a cancellation token is used. I guess this is a safe change.

PSReadLine/ReadLine.cs
andyleejordan: 👍

andyleejordan:
Thanks @daxian-dbw! Do you think we could get a 2.2.3 release this week?

daxian-dbw:
Yeah, we can do a 2.2.3 release this week.

PR #3332, opened 2022-06-01
Add the feature of multiple keyword interactive search.

Andy-AO:
# PR Summary

- Adds the feature of multiple keyword interactive search.
- Refactoring related code, mainly to split the oversized classes and encapsulate the repeated code.

https://user-images.githubusercontent.com/38101170/171310015-bfc13b1a-c3d2-4f92-915a-9480f3de30d4.mp4

Forgive me for accidentally using the IDE's automatic cleaning program.

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

Added two new functions, the documentation will be supplemented later.
- ReverseSearchHistoryMultiKeyword
- ForwardSearchHistoryMultiKeyword


Andy-AO:
I also plan to make more improvements to the search function.
- Circular search
- Show search position in prompt
- Provide a list of results

daxian-dbw:
@Andy-AO Thanks for submitting the PR. There are too many style changes in the PR which makes it impossible to review functionality related changes. We don't accept any style changes, so please remove all style changes and only keep the functionality related code.

Andy-AO:
> @Andy-AO Thanks for submitting the PR. There are too many style changes in the PR which makes it impossible to review functionality related changes. We don't accept any style changes, so please remove all style changes and only keep the functionality related code.

OK. I will try.
I think I need to look at the relevant blocks one by one and decide whether to keep them or change them manually.
What git commands or working methods should I use? What do you suggest?

daxian-dbw:
I don't have any suggestions for tool or commands, and I'm afraid you may need to manually revert all changes that are not directly related to the new functionality.

Andy-AO:
> I don't have any suggestions for tool or commands, and I'm afraid you may need to manually revert all changes that are not directly related to the new functionality.

Can we set an automatic formatting tool? Many projects will run the automatic formatting program when submitting, so as to avoid many problems.

daxian-dbw:
It's not just formatting, but also all sorts of coding style changes, like changing `string varname` to `var varname`, moving class fields or properties, changing `if/while/for/... { xxx }` to `if/while/for/... xxx;`, and etc. All those coding style changes should be reverted, and only keep the functionality related changes.

Andy-AO:
> It's not just formatting, but also all sorts of coding style changes, like changing `string varname` to `var varname`, moving class fields or properties, changing `if/while/for/... { xxx }` to `if/while/for/... xxx;`, and etc. All those coding style changes should be reverted, and only keep the functionality related changes.

OK.

But, the purpose of moving class fields or properties is to adjust the responsibilities of classes, not meaningless, this is a refactoring technique, the goal is to reduce the complexity of the code.

It cannot exist independently of my function addition.

Andy-AO:
> `string varname` to `var varname`

This does not belong to the format in the normal sense. It is an automated code cleaning function. It executes predetermined rules.

Andy-AO:
> It's not just formatting, but also all sorts of coding style changes, like changing `string varname` to `var varname`, moving class fields or properties, changing `if/while/for/... { xxx }` to `if/while/for/... xxx;`, and etc. All those coding style changes should be reverted, and only keep the functionality related changes.

Since there is no way to separate refactoring from function addition, I have no way to undo the operation of reassigning class responsibilities. Do you think that such refactoring techniques as moving fields and extracting classes are also unacceptable?

Andy-AO:
@daxian-dbw 
Refactoring (in this case, mainly extracting classes to move fields and renaming them) is important for modifying the project to reduce the complexity of the code and make subsequent changes less difficult.

The changes are spread throughout the project, and I feel helpless about it, because the current situation of this project is that most of the code is located in one class, which means that when I extract the class, many files will be affected.

But there is no way to avoid this change as long as the bad design of oversized classes is dropped, which is a necessary cost.

If you want to review the feature additions, then you only need to look at the following files; other changes due to splitting classes are of little importance.

- HistoryTest.cs
- Emphasis.cs
- Renderer.cs
- HistorySearcherModel.cs
- HistorySearcherReadLine.cs
- Renderer.DataBuilder.cs

> like changing `if/while/for/... { xxx }` to `if/while/for/... xxx`;

I'm sorry about these changes, there's really no need for them, it was caused by my mistake and it's blocking your review, if I had to roll it back manually then the amount of work would be unimaginable.

But I think we just need to solve the side effect caused by this error, that is, the obstacle to its review.

The solution is simple, that is, use Rider to do the same operation on the original code in advance, so that most of the obstacles will be removed.

As for the correctness of these automated cleaning, I think there is no doubt. Besides, there are automated tests to ensure.

I have pushed new changes. You can review the relevant content based on the information I have provided.

Thank you very much！

daxian-dbw:
image

@Andy-AO With the size of changes in this PR, it's impossible to review it. Even if the refactoring is meaningful, we don't accept this magnitude of refactoring.
I'm open to replace the existing `string.Contains` searching with a fuzz searching such as the algorithm used in [fzf](https://github.com/junegunn/fzf), so it doesn't have to be a separate mode, which will simplify the code changes.

Andy-AO:
Thank you for your reply. I am willing to participate in this project. PowerShell is a good tool.

I agree with you that you can implement similar things without changing the current design.

Too large classes, too long functions and repetitive logic deserve our attention, because they will bring long-term adverse effects.

Sometimes we need to take a longer view. Although this revision seems to have a large scope of influence, as the coupling is gradually removed, the scope of influence of subsequent revisions will become smaller and clearer.

Changes will help we quickly add subsequent functions. I have many plans, and these modifications have laid a solid foundation for the implementation of those plans.

We can all feel the fear of modification, but the statistical data is problematic. It is just a mechanical calculation difference, without considering the two special cases I mentioned. In fact, the modification is much smaller than the above.

There are automated tests, we should trust them. These modification has passed all tests and it is safe.

daxian-dbw:
Thanks for your passion on this project, but we are not going to take refactoring in such magnitude. Automation tests don't have 100% coverage, changes in this size will definitely break something. Also, I personally don't think the current design and code structure prevent further feature works in any ways, and thus I don't think a fundamental refactoring is needed.

Andy-AO:
I understand your concerns, I'll be using this branch for the long term and continue to add more useful features to better demonstrate the potential of this design improvement.

PR #3351, opened 2022-06-21
Enable IntelliSense prediction by default

daxian-dbw:
# PR Summary

Enable intellisense prediction by default:
- when VT is not supported, `PredictionSource` is set to `None` by default
- when VT is supported and it's running with PowerShell 7.2+, `PredictionSource` is set to `HistoryAndPlugin` by default
- when VT is supported and it's running with PowerShell prior to 7.2, `PredictionSource` is set to `History` by default

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [x] Doc Issue filed: https://github.com/MicrosoftDocs/PowerShell-Docs/issues/8943

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3351)

### DIFF
--- a/PSReadLine/Cmdlets.cs
+++ b/PSReadLine/Cmdlets.cs
@@ -165,7 +165,7 @@ namespace Microsoft.PowerShell
         /// </summary>
         public const int DefaultAnsiEscapeTimeout = 100;
 
-        public PSConsoleReadLineOptions(string hostName)
+        public PSConsoleReadLineOptions(string hostName, bool usingLegacyConsole)
         {
             ResetColors();
             EditMode = DefaultEditMode;
@@ -185,7 +185,11 @@ namespace Microsoft.PowerShell
             HistorySearchCaseSensitive = DefaultHistorySearchCaseSensitive;
             HistorySaveStyle = DefaultHistorySaveStyle;
             AnsiEscapeTimeout = DefaultAnsiEscapeTimeout;
-            PredictionSource = DefaultPredictionSource;
+            PredictionSource = usingLegacyConsole
+                ? PredictionSource.None
+                : Environment.Version.Major < 6
+                    ? PredictionSource.History
+                    : PredictionSource.HistoryAndPlugin;
             PredictionViewStyle = DefaultPredictionViewStyle;
             MaximumHistoryCount = 0;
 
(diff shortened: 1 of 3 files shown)


daxian-dbw:
@andschwa Can you please review this PR when you get a chance? Thanks!

PSReadLine/Cmdlets.cs
```diff
@@ -185,7 +185,11 @@ public PSConsoleReadLineOptions(string hostName)
             HistorySearchCaseSensitive = DefaultHistorySearchCaseSensitive;
             HistorySaveStyle = DefaultHistorySaveStyle;
             AnsiEscapeTimeout = DefaultAnsiEscapeTimeout;
-            PredictionSource = DefaultPredictionSource;
+            PredictionSource = usingLegacyConsole
+                ? PredictionSource.None
+                : Environment.Version.Major < 6
```
andyleejordan, line 190: Just double-checking here @daxian-dbw, your PR description says:

> when VT is supported and it's running with PowerShell 7.2+, PredictionSource is set to HistoryAndPlugin by default
when VT is supported and it's running with PowerShell prior to 7.2, PredictionSource is set to History by default

But this is checking `Version.Major < 6`

So 7 and up (theoretically including 7.0 and 7.1, not just 7.2+, also any old versions of 6 that may still be out there) are going to be `HistoryAndPlugin`. Is that what you intend?

andyleejordan (changes requested):

PSReadLine/Cmdlets.cs
daxian-dbw: It's `Environment.Version.Major`, meaning the major version of the .NET Core. PowerShell 7.2 comes with .NET 6, so if `Environment.Version.Major < 6`, the PowerShell version is lower than 7.2, and we will use `History` as the default prediction source.

andyleejordan: Got it, got it. Thanks for checking!

andyleejordan (approved):

daxian-dbw:
@andschwa Thanks for the review!

PR #3435, opened 2022-09-15
Changes F7 sample to use Out-ConsoleGridView

tig:
# PR Summary

Updates the PSReadLine sample code to use `Out-ConsoleGridView` instead of the defunct (and not cross platform) `Out-GridView`.

Note that the old sample supported multi-select and the F7History.ps1 sample does not. It is possible to modify it to support multi-select, but doing so exposes a big question:

- In what order should the selected history items be executed. Neither `Out-GridView` or `Out-ConsoleGridView` support a way for the user to determine this. 

## PR Checklist

- [ X] PR has a meaningful title
- [ X] Summarized changes
- [ X] Make sure you've tested these changes in terminals that PowerShell is commonly used in (i.e. conhost.exe, Windows Terminal, Visual Studio Code Integrated Terminal, etc.)
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - [] Documentation needed at [PowerShell-Docs](https://github.com/MicrosoftDocs/PowerShell-Docs)
        - [ ] Doc Issue filed: 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3435)

### DIFF
--- a/PSReadLine/SamplePSReadLineProfile.ps1
+++ b/PSReadLine/SamplePSReadLineProfile.ps1
@@ -20,62 +20,90 @@ Set-PSReadLineOption -HistorySearchCursorMovesToEnd
 Set-PSReadLineKeyHandler -Key UpArrow -Function HistorySearchBackward
 Set-PSReadLineKeyHandler -Key DownArrow -Function HistorySearchForward
 
-# This key handler shows the entire or filtered history using Out-GridView. The
+# Use F7 as "Show Command History" in Powershell
+#
+# Requires `Out-ConsoleGridView` from [PowerShell/GraphicalTools](https://github.com/PowerShell/GraphicalTools). 
+#
+#  1. Install `ConsoleGuiTools` by typing the command `Install-Module -Name Microsoft.PowerShell.ConsoleGuiTools`
+#  2. Include the `F7History.ps1` script in your $profile:
+#
+#      # Include (don't RUN) the script by prefixing it with `. `.
+#      . ~/source/F7History/F7History.ps1
+#
+# This key handler shows the entire or filtered history using Out-ConsoleGridView. The
 # typed text is used as the substring pattern for filtering. A selected command
-# is inserted to the command line without invoking. Multiple command selection
-# is supported, e.g. selected by Ctrl + Click.
-Set-PSReadLineKeyHandler -Key F7 `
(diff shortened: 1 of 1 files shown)


daxian-dbw:
As a sample, the benefit of using `Out-GridView` is that it's available by default. While for `Out-ConsoleGridView`, it requires a user to install that module first. So, for the purpose of a sample, we should continue to use `Out-GridView` here.

However, if you can publish your handler script that uses `Out-ConsoleGridView` as a gist, we may be able to point to it in the comment of the `F7` example.

tig:
> As a sample, the benefit of using `Out-GridView` is that it's available by default. While for `Out-ConsoleGridView`, it requires a user to install that module first. So, for the purpose of a sample, we should continue to use `Out-GridView` here.
> 
> However, if you can publish your handler script that uses `Out-ConsoleGridView` as a gist, we may be able to point to it in the comment of the `F7` example.

Ok, but two points that might influence you:

1) `Out-GridView` only works on Windows and Mac. Not Linux.
2) `Out-GridView` is deprecated by Microsoft.

daxian-dbw:
I'm not aware that `Out-GridView` is to be deprecated. But you are right on the 1st point. So, if you can publish the "F7 with Out-ConsoleGridView" gist, we can consider point to it in the sample.

tig:
I forgot to mention: I've recently packaged F7History up as a PowerShell Gallery-installable module:

https://github.com/gui-cs/F7History/

With this, perhaps this PR is really about just pointing to the F7History repo as a good example?

Install `F7History` from the [PowerShell Gallery](https://www.powershellgallery.com/packages/F7History/).

```ps1
Install-Module -Name "F7History"
```

Add a line to import `F7History` in your PowerShell `$profile`:

```ps1
Import-Module -Name "F7History"
```

To change the key bindings, use the `-ArgumentList` parameter when importing the module. For example, to use `F6` and `Shift-F6` instead of `F7` and `Shift-F7`:

```ps1
Import-Module -Name "F7History" -ArgumentList  @{Key = "F6"; AllKey = "Shift-F6"}
```

daxian-dbw:
The comment of the `F7` sample was updated to point to the `F7History` module.

PR #3493, opened 2022-11-23
Change default color for inline prediction to `dim`

### LINKED ISSUE #3477: Predictive IntelliSense's text color is inaccessible
### Prerequisites

### Exception report

```console
N/A
```

### Screenshot

cc @daxian-dbw @SteveL-MSFT 

I just started seeing this feature described in https://github.com/PowerShell/PSReadLine/issues/1468 and n noticed that it has hardcoded indexed color 238 (`#444444`), if you only set the foreground color to a fixed color it will have wildly varying levels of contrast depending on the terminal theme's background color, unless you also set the background color.

Prediction looks like it's regular input with `#fff` bg, `#444` fg:

image

xterm.js' default demo theme is extremely low contrast, `#2D2E2C` bg, `#F8F8F8` fg:

image

Prediction is invisible with `#444` bg color:

image

I notice you can set the color manually via `Set-PSReadLineOption -Colors @{ Prediction = '#8A0303'}` but saying this is a good default and enough is a copout imo. Some users won't even know it's a feature because they can't see the text, and some users will think the shell is broken because it looks like regular text.

VS Code's minimum contrast feature actually prevents this contrast issue, however it brings another problem where it can more easily look like the default text color:

image

Considering this case, I think you should use the dim attribute by default to style the text while still allowing manual tweaking in psreadline options and in xterm.js we should exclude the dim attribute from the minimum contrast demands (or halve them).

### Environment data

```console
PS Version: 7.3.0
PS HostName: ConsoleHost
PSReadLine Version: 2.2.6
PSReadLine EditMode: Windows
OS: 10.0.22621.1 (WinBuild.160101.0800)
BufferWidth: 148
BufferHeight: 19
```

### Steps to reproduce

Type `git`

### Expected behavior

The prediction has reasonable contrast and is differentiated enough from the default foreground color, regardless of the terminal's background color.

### Actual behavior

It can range from completely invisible to identical to the regular input.

SteveL-MSFT:
# PR Summary

The current default inline prediction color doesn't provide sufficient contrast in terminals that don't use a pure black background (like VSCode terminal).  Fix is to use `dim white italic` escape sequence.  Just using `dim` would result in the color changing depending on the token currently being rendered and for the dark-grey color for parameters it makes it even darker.

Fix https://github.com/PowerShell/PSReadLine/issues/3477

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3493)

### DIFF
--- a/PSReadLine/Cmdlets.cs
+++ b/PSReadLine/Cmdlets.cs
@@ -95,10 +95,10 @@ namespace Microsoft.PowerShell
 
         // Find the most suitable color using https://stackoverflow.com/a/33206814
         // Default prediction color settings:
-        //  - use FG color 'dark black' for the inline-view suggestion text
+        //  - use FG color 'dim' for the inline-view suggestion text
         //  - use FG color 'yellow' for the list-view suggestion text
         //  - use BG color 'dark black' for the selected list-view suggestion text
-        public const string DefaultInlinePredictionColor       = "\x1b[38;5;238m";
+        public const string DefaultInlinePredictionColor       = "\x1b[2m";
         public const string DefaultListPredictionColor         = "\x1b[33m";
         public const string DefaultListPredictionSelectedColor = "\x1b[48;5;238m";
 
--- a/PSReadLine/Cmdlets.cs
+++ b/PSReadLine/Cmdlets.cs
@@ -95,10 +95,10 @@ namespace Microsoft.PowerShell
 
         // Find the most suitable color using https://stackoverflow.com/a/33206814
         // Default prediction color settings:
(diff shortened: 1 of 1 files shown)


daxian-dbw (approved):
LGTM. Thanks for submitting the fix!

PR #3543, opened 2023-01-19
De-duplicate prediction results with the history results

daxian-dbw:
# PR Summary

De-duplicate prediction results with the history results.
A prediction result could be the same as the one found in history. This happens sometimes to the Az predictor.
When used in list view, it's confusing for the user to see 2 same results in the list, both from history and a predictor. So, we de-duplicate the prediction results against the history results.

For now, we only remove the prediction results that are exactly the same (case-sensitive or case-insensitive decided by the user with the 'HistorySearchCaseSensitive' setting, which is `false` by default) as one of the history results.

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3543)

### DIFF
--- a/PSReadLine/Prediction.Views.cs
+++ b/PSReadLine/Prediction.Views.cs
@@ -144,12 +144,9 @@ namespace Microsoft.PowerShell
                         continue;
                     }
 
-                    if (results == null)
-                    {
-                        results = new List<SuggestionEntry>(capacity: count);
-                    }
-
                     _cacheHistorySet.Add(line);
+                    results ??= new List<SuggestionEntry>(capacity: count);
+
                     if (matchIndex == 0)
                     {
                         results.Add(new SuggestionEntry(line, matchIndex));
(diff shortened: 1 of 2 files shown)


StevenBucher98 (approved):
LGTM, awesome work Dongbo!

Verified previous duplications with Az Predictor are no longer there on this branch

PSReadLine/Prediction.Views.cs
```diff
@@ -435,19 +447,46 @@ private void AggregateSuggestions()
...
+                                    // those skipped ones. This is due to the limitation of the 'OnSuggestionDisplayed' interface method, which didn't
+                                    // assume any prediction results from a predictor could be filtered out at the initial design time. We will have to
+                                    // change the predictor interface to pass in accurate information, such as:
+                                    //   void OnSuggestionDisplayed(Guid predictorId, uint session, int countOrIndex, int[] skippedIndices)
+                                    //
+                                    // However, an interface change has huge impacts. At least, a newer version of PSReadLine will stop working on the
+                                    // existing PowerShell 7+ versions. For this particular issue, the chance that it could happen is low and the impact
+                                    // of the inaccurate feedback is also low, so we should delay this interface change until another highly-demanded
+                                    // change to the interface is required in future (e.g. changes related to supporting OpenAI models).
```
andyleejordan, line 488: Thanks for the detailed note 👍

andyleejordan (approved):
Looks great, thanks Dongbo!

daxian-dbw:
Thank you both for the reivew!

PR #3558, opened 2023-02-01
[WIP] Make PredictionListView adapt to the height of the window

### LINKED ISSUE #3557: Allow PredictionListView suggestion count to auto-adjust based on the height of the console
### Prerequisites

I love the `Set-PSReadLineOption -PredictionViewStyle ListView` option of PSReadLine nad rely on it everyday. One issue I run into a lot is when I am in a code editor (VS Code, Visual Studio, Rider, etc) and I have my terminal open, usually the terminal is at the bottom of my screen and shrunk down to only a few lines visible which results in this warning:

```
WARNING: The prediction 'ListView' is temporarily disabled because the current window size of the console is too small. To use the 'ListView', please make sure the 'WindowWidth' is not less than '54' and the 'WindowHeight' is not less than '15'.
```

image

While it is true that the window does not have very many lines visible, I still feel like there is enough room there to show at least a few items in the list view.

Instead of having a hard cap of 10 suggestions, I think the suggestion count should auto-adjust based on the size of the window.

Currently there is a `MinWindowHeight` of 15 lines. Anything less than this and the `PredictionListView` simply stops running.

Instead of having a hard cap of 15 lines visible, I think the suggestion count should simply decrease for smaller windows. There is currently a hard upper limit (`ListMaxCount`) of 10 suggestions. I could see there being a minimum suggestion count (`ListMinCount`) of maybe 3 suggestions to prevent there from being any weird behavior.

ChristopherHaws:
Closes #3557

# PR Summary

This PR changes the `PredictionListView` to allow for suggestions to show up on terminals that have a height less than 15 lines tall. Instead of having a hard limit of 15 lines (and 10 suggestions), this PR makes it so there is no longer a limit of 15 lines. Instead there is a min limit of 5 lines (3 suggestions + 2 lines reserved for the terminal prompt). As the terminal gets taller, more suggestions will be displayed, up to a max of 10 (which is the current limit).

Some improvements that could possibly be made are:
- Making the upper and lower bounds configurable instead of hard coded allowing users to change the default behavior.
- Add an option to have PSReadLine automatically switch to `PredictionInlineView` when the windows is smaller than `PredictionListView.MinWindowHeight` and switch back when the terminal window is increased in size

I haven't added tests yet since this is a draft PR. I wanted to get feedback before I write the tests to make sure this is something that might actually get merged :)

## PR Checklist

- [ ] Summarized changes
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3558)

### DIFF
--- a/PSReadLine/Prediction.Views.cs
+++ b/PSReadLine/Prediction.Views.cs
@@ -207,12 +207,14 @@ namespace Microsoft.PowerShell
         /// </summary>
         private class PredictionListView : PredictionViewBase
         {
+            internal const int MaxPromptLineCount = 2;
             internal const int ListMaxCount = 10;
+            internal const int ListMinCount = 3;
             internal const int ListMaxWidth = 100;
             internal const int SourceMaxWidth = 15;
 
             internal const int MinWindowWidth = 54;
-            internal const int MinWindowHeight = 15;
+            internal const int MinWindowHeight = MaxPromptLineCount + ListMinCount;
 
             private List<SuggestionEntry> _listItems;
             private int _listItemWidth;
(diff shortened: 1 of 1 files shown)


iSazonov:
@daxian-dbw What do you think about the proposal?

daxian-dbw:
@ChristopherHaws Thanks for submitting the PR!

I already have a work-in-progress branch that improves the list view with the following functionalities:
1. scrollable list view, so a smaller window can still show a good amount of results
2. adjustable list view height, to adapt to use cases like VSCode terminal and Azure Cloud Shell.

So, I will close this PR, but thanks for experimenting with this idea.

PR #3583, opened 2023-02-17
Improve the list view to be scrollable and auto-adjust the list view height

### LINKED ISSUE #3563: Auto-resize Predictor ListView when the height is too small
### Prerequisites

When my shell windows are too small to display all ten predictions in `ListView`, it should reduce the size to be able to display a smaller number of predictions. 

This is particularly applicable in CloudShell where the default window when you open CloudShell is too small for `ListView` and so fails silently when I try using `F2` and gives this error message when I try to set:
```
PS /home/steven> Set-PSReadLineOption -PredictionViewStyle ListView
WARNING: The prediction 'ListView' is temporarily disabled because the current window size of the console is too small. To use the 'ListView', please make sure the 'WindowWidth' is not less than '54' and the 'WindowHeight' is not less than '15'.
```

https://user-images.githubusercontent.com/20605456/216405289-d6449cb9-5b75-4306-95b1-53a204d9d7b3.mov

There are a few ways to improve this experience. 
- Incorporate similiar error message that `Set-PSReadLineOption` gives when the user tries F2
- Make `ListView` number of results resizable based on the window.
- Make `ListView` number of results resizable based on the window and incorporate scrollable functionality to enable users to see the other results given a smaller window.

cc @theJasonHelmick @daxian-dbw

daxian-dbw:
## PR Summary

Fix #3563
Improve the list view to be scrollable and auto-adjust the list view height.
- Now the list view is scrollable. It can display up to 50 items.
- The max list view height is still 10, but the list view height can auto-adjust based on the terminal height: the max height changes to 5 when the terminal height is <=20, and then changes to 3 when the terminal height is <= 10.
- A metadata line is added to show extra information for the list view.
- When the terminal size is too small to show the list view, it prints a warning message to inform the user about it.
- The key-bindings <kbd>PageUp</kbd> and <kbd>PageDown</kbd> can navigate within the list view by moving up/down a page.
- The key-bindings <kbd>Ctrl+PageUp</kbd> and <kbd>Ctrl+PageUp</kbd> can navigate within the list view by jumping to the previous/next source.

The scrollable list view:

scrollable

The auto-adjustable list view height:

auto-adjust

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3583)

### DIFF
--- a/PSReadLine/PSReadLineResources.resx
+++ b/PSReadLine/PSReadLineResources.resx
@@ -841,7 +841,7 @@ Or not saving history with:
     <value>The prediction 'ListView' is temporarily disabled because the current window size of the console is too small. To use the 'ListView', please make sure the 'WindowWidth' is not less than '{0}' and the 'WindowHeight' is not less than '{1}'.</value>
   </data>
   <data name="WindowSizeTooSmallWarning" xml:space="preserve">
-    <value>! term size too small to show the list view</value>
+    <value>! terminal size too small to show the list view</value>
   </data>
   <data name="NeedsUpdateHelp" xml:space="preserve">
     <value>No help content available. Please use Update-Help to download the latest help content.</value>
--- a/PSReadLine/PSReadLineResources.resx
+++ b/PSReadLine/PSReadLineResources.resx
@@ -840,6 +840,9 @@ Or not saving history with:
   <data name="WindowSizeTooSmallForListView" xml:space="preserve">
     <value>The prediction 'ListView' is temporarily disabled because the current window size of the console is too small. To use the 'ListView', please make sure the 'WindowWidth' is not less than '{0}' and the 'WindowHeight' is not less than '{1}'.</value>
   </data>
+  <data name="WindowSizeTooSmallWarning" xml:space="preserve">
+    <value>! term size too small to show the list view</value>
+  </data>
   <data name="NeedsUpdateHelp" xml:space="preserve">
     <value>No help content available. Please use Update-Help to download the latest help content.</value>
   </data>
--- a/PSReadLine/Prediction.Views.cs
+++ b/PSReadLine/Prediction.Views.cs
@@ -678,9 +678,9 @@ namespace Microsoft.PowerShell
             private void RenderWarningLine(StringBuilder buffer)
             {
                 // Add italic text effect to the highlight color.
-                string highlightColor = _singleton._options._listPredictionColor + "\x1b[3m";
+                string highlightStyle = _singleton._options._listPredictionColor + "\x1b[3m";
 
-                buffer.Append(highlightColor)
+                buffer.Append(highlightStyle)
                     .Append(PSReadLineResources.WindowSizeTooSmallWarning)
                     .Append(VTColorUtils.AnsiReset);
             }
@@ -708,12 +708,12 @@ namespace Microsoft.PowerShell
             private void RenderMetadataLine(StringBuilder buffer)
             {
                 // Add italic text effect to the highlight color.
-                string highlightColor = _singleton._options._listPredictionColor + "\x1b[3m";
-                string dimmedColor = PSConsoleReadLineOptions.DefaultInlinePredictionColor;
-                string activeColor = null;
+                string highlightStyle = _singleton._options._listPredictionColor + "\x1b[3m";
+                string dimmedStyle = PSConsoleReadLineOptions.DefaultInlinePredictionColor;
+                string activeStyle = null;
 
                 // Render the quick indicator.
-                buffer.Append(highlightColor)
+                buffer.Append(highlightStyle)
                     .Append('<')
                     .Append(_selectedIndex > -1 ? _selectedIndex + 1 : "-")
                     .Append('/')
@@ -759,7 +759,7 @@ namespace Microsoft.PowerShell
                 //    will be 55 (7+48), so we choose '60' as the minimal requirement for rendering 2 sources.
                 int maxSourceCount = _listViewWidth >= 80 ? 3 : 2;
                 int charPosition = buffer.Length;
-                int totalCountPartLength = buffer.Length - highlightColor.Length - VTColorUtils.AnsiReset.Length;
+                int totalCountPartLength = buffer.Length - highlightStyle.Length - VTColorUtils.AnsiReset.Length;
                 int additionalPartLength = 0;
 
                 int selected = -1;
@@ -793,7 +793,7 @@ namespace Microsoft.PowerShell
                 }
 
                 // Start the extra information about the sources -- add the opening arrow bracket.
-                AppendColor(buffer, dimmedColor, ref activeColor, out _).Append('<');
+                AppendColor(buffer, dimmedStyle, ref activeStyle, out _).Append('<');
                 additionalPartLength++;
 
                 // Add the prefix, continue to use dimmed color.
@@ -823,7 +823,7 @@ namespace Microsoft.PowerShell
                     SourceInfo info = _sources[index];
                     if (selected == index)
                     {
-                        AppendColor(buffer, highlightColor, ref activeColor, out nextCharPos)
+                        AppendColor(buffer, highlightStyle, ref activeStyle, out nextCharPos)
                             .Append(info.SourceName)
                             .Append('(')
                             .Append(_selectedIndex - info.PrevSourceEndIndex)
@@ -833,7 +833,7 @@ namespace Microsoft.PowerShell
                     }
                     else
                     {
-                        AppendColor(buffer, dimmedColor, ref activeColor, out nextCharPos)
+                        AppendColor(buffer, dimmedStyle, ref activeStyle, out nextCharPos)
                             .Append(info.SourceName)
                             .Append('(')
                             .Append(info.ItemCount)
@@ -847,14 +847,14 @@ namespace Microsoft.PowerShell
                 // Add the suffix.
                 if (startFrom + maxSourceCount < _sources.Count)
                 {
-                    AppendColor(buffer, dimmedColor, ref activeColor, out _)
+                    AppendColor(buffer, dimmedStyle, ref activeStyle, out _)
                         .Append(' ')
                         .Append(SuggestionEntry.Ellipsis);
                     additionalPartLength += 2;
                 }
 
                 // Add the closing arrow bracket.
-                AppendColor(buffer, dimmedColor, ref activeColor, out _)
+                AppendColor(buffer, dimmedStyle, ref activeStyle, out _)
                     .Append('>')
                     .Append(VTColorUtils.AnsiReset);
                 additionalPartLength++;
--- a/PSReadLine/Prediction.Views.cs
+++ b/PSReadLine/Prediction.Views.cs
@@ -207,18 +207,27 @@ namespace Microsoft.PowerShell
         /// </summary>
         private class PredictionListView : PredictionViewBase
         {
-            internal const int ListMaxCount = 10;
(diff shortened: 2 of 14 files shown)


PSReadLine/Prediction.Views.cs
```diff
@@ -273,6 +303,36 @@ internal PredictionListView(PSConsoleReadLine singleton)
             internal override bool HasPendingUpdate => _updatePending;
             internal override bool HasActiveSuggestion => _listItems != null;
 
+            /// <summary>
+            /// Calculate the max width and height of the list view based on the current terminal size.
+            /// </summary>
+            private (int maxWidth, int maxHeight, bool extraCheck) RefreshMaxViewSize()
```
SteveL-MSFT: `extraCheck` here is too ambiguous.  Looking at the usage below, would it be better to call it `heightCheck`?

PSReadLine/Prediction.Views.cs
```diff
@@ -501,16 +581,35 @@ private void AggregateSuggestions()
 
                 if (_listItems?.Count > 0)
                 {
-                    _listItemHeight = Math.Min(_listItems.Count, ListMaxCount);
+                    // Initialize the view window position here.
+                    _listViewTop = 0;
+                    _listViewEnd = Math.Min(_listItems.Count, _maxViewHeight);
+                    _listViewHeight = _listViewEnd - _listViewTop;
```
SteveL-MSFT, line 587: isn't `_listViewTop` always zero here?

PSReadLine/Prediction.Views.cs
```diff
@@ -520,36 +619,254 @@ internal override void RenderSuggestion(List<StringBuilder> consoleBufferLines,
...
+                }
+            }
+
+            /// <summary>
+            /// Generate the rendering text for the warning message.
+            /// </summary>
+            private void RenderWarningLine(StringBuilder buffer)
+            {
+                // Add italic text effect to the highlight color.
+                string highlightColor = _singleton._options._listPredictionColor + "\x1b[3m";
```
SteveL-MSFT: The use of italics should just be part of the configurable prediction highlight instead of always being added.  Some users may prefer bold, for example.  So this should be `highlightStyle` and not `highlightColor`

PSReadLine/Prediction.Views.cs
```diff
@@ -520,36 +619,254 @@ internal override void RenderSuggestion(List<StringBuilder> consoleBufferLines,
...
+                    else
+                    {
+                        AppendColor(buffer, dimmedColor, ref activeColor, out nextCharPos)
+                            .Append(info.SourceName)
+                            .Append('(')
+                            .Append(info.ItemCount)
+                            .Append(')');
+                    }
+
+                    // Need to take into account multi-cell characters when calculating length.
```
SteveL-MSFT, line 843: Have you tested with double byte chars and or emojis?

daxian-dbw: Yes, I did test with double byte characters:

image

But emojis are not guaranteed to work, because PSReadLine today doesn't handle surrogate pairs well. There is a label used to track all the emoji related issues.

PSReadLine/Prediction.Views.cs
daxian-dbw: About the new colors and styles, there are a few of them:
- un-highlighted texts in the metadata line
- highlighted text in the metadata line
- warning message

Currently,
- the un-highlighted text uses the default inline view color/style;
- the highlighted text uses the `ListPredictionColor` plus `italics`;
- the warning message uses the `ListPredictionColor` plus `italics`;

I'm not sure about what to expose for configuration yet. It doesn't feel needed to expose them all in my opinion. The following are my thoughts:
1. always apply `italics` to the text rendered in the metadata line, and use the `ListPredictionColor` as the user-configurable highlighting color/style.
2. maybe expose configuration for un-highlighted text color/style, but I'm not super sure if it's really needed.
3. warning message in this case will be uncommon for users to see because it will usually happen when the terminal height is very small (<5). So, exposing a configuration for it seems overkill to me.

I don't have a clear idea what/how to make configurable about the new color/styles at the moment, so I prefer to wait and collect some feedback about this, which hopefully will give us a clear idea. I will open an issue about this to track it.

PSReadLine/Prediction.Views.cs
daxian-dbw: It is. I prefer `_listViewHeight = _listViewEnd - _listViewTop` over `_listViewHeight = _listViewEnd` here because it shows how the height is calculated.

PSReadLine/Prediction.Views.cs
daxian-dbw: Changed the name to be `checkOnHeight`.

PSReadLine/Prediction.Views.cs
daxian-dbw: Issue opened: https://github.com/PowerShell/PSReadLine/issues/3584

PSReadLine/PSReadLineResources.resx
```diff
@@ -840,6 +840,9 @@ Or not saving history with:
   <data name="WindowSizeTooSmallForListView" xml:space="preserve">
     <value>The prediction 'ListView' is temporarily disabled because the current window size of the console is too small. To use the 'ListView', please make sure the 'WindowWidth' is not less than '{0}' and the 'WindowHeight' is not less than '{1}'.</value>
   </data>
+  <data name="WindowSizeTooSmallWarning" xml:space="preserve">
+    <value>! term size too small to show the list view</value>
```
SteveL-MSFT:
```suggestion
    <value>! terminal height too small to show the list view</value>
```

SteveL-MSFT (approved):

PSReadLine/PSReadLineResources.resx
daxian-dbw: This warning also applies when width is too small, which should be pretty rare to happen, but could happen. I will change `term` to `terminal`.

PSReadLine/Prediction.Views.cs
```diff
@@ -273,6 +303,36 @@ internal PredictionListView(PSConsoleReadLine singleton)
...
+            {
+                var console = _singleton._console;
+                int maxWidth = Math.Min(console.BufferWidth, ListViewMaxWidth);
+
+                (int maxHeight, bool moreCheck) = console.BufferHeight switch
+                {
+                    > ListViewMaxHeight * 2 => (ListViewMaxHeight, false),
+                    > ListViewMaxHeight => (ListViewMaxHeight / 2, false),
+                    _ => (ListViewMaxHeight / 3, true)
+                };
```
andyleejordan, line 319: Am I correct in parsing this as "10 if the console is taller than 20, else 5 if it's between 10 and 20, else 3"? What's the reason to not just fill the terminal e.g. `Math.Min(console.BufferHeight, ListViewMaxHeight)`, aesthetics?

PSReadLine/Prediction.Views.cs
```diff
@@ -323,7 +386,11 @@ internal override void GetSuggestion(string userInput)
 
                     if (UseHistory)
                     {
-                        _listItems = GetHistorySuggestions(userInput, ListMaxCount);
+                        _listItems = GetHistorySuggestions(userInput, HistoryMaxCount);
+                        if (_listItems?.Count > 0)
+                        {
+                            _sources = new List<SourceInfo>() { new SourceInfo(SuggestionEntry.HistorySource, _listItems.Count - 1, -1) };
```
andyleejordan, line 392: This is going to set `ItemCount = EndIndex - PrevSourceEndIndex;` in the initializer of `SourceInfo` for `_listItems.Count - 1 - -1`, which I think is what you're intending since the double negative of `PrevSourceEndIndex` cancels out the -1 applied to `Count`, and so we get back the original `_listItems.Count`. Just had to walk myself through it to catch that double-negative (thought at first we were subtracting 2 instead).

andyleejordan: Maybe add `prevSourceEndIndex:` to name the argument like you do below, just for clarification really.

```suggestion
                            _sources = new List<SourceInfo>() { new SourceInfo(SuggestionEntry.HistorySource, _listItems.Count - 1, prevSourceEndIndex: -1) };
```

PSReadLine/Prediction.Views.cs
```diff
@@ -471,22 +544,29 @@ private void AggregateSuggestions()
...
-                                    // change the predictor interface to pass in accurate information, such as:
-                                    //   void OnSuggestionDisplayed(Guid predictorId, uint session, int countOrIndex, int[] skippedIndices)
-                                    //
-                                    // However, an interface change has huge impacts. At least, a newer version of PSReadLine will stop working on the
-                                    // existing PowerShell 7+ versions. For this particular issue, the chance that it could happen is low and the impact
-                                    // of the inaccurate feedback is also low, so we should delay this interface change until another highly-demanded
-                                    // change to the interface is required in future (e.g. changes related to supporting OpenAI models).
-                                    _singleton._mockableMethods.OnSuggestionDisplayed(item.Id, item.Session.Value, count + skipCount);
+                                    int prevEndIndex = _sources.Count > 0 ? _sources[_sources.Count - 1].EndIndex : -1;
```
andyleejordan, line 549: I'm trying to make a committable suggestion but GitHub is not agreeing with me today, just add a comment:
```
// If we had at least one source, we take the end index of the last source in the list.
```

PSReadLine/Prediction.Views.cs
```diff
@@ -501,16 +581,35 @@ private void AggregateSuggestions()
...
                     Reset();
                 }
             }
 
+            /// <summary>
+            /// Generate the rendering text for the list view.
+            /// </summary>
             internal override void RenderSuggestion(List<StringBuilder> consoleBufferLines, ref int currentLogicalLine)
             {
+                if (_warnAboutSize || (_checkOnHeight && HeightIsTooSmall()))
```
andyleejordan, line 600: Ah that's where `checkOnHeight` ends up being used.

andyleejordan (commented):
LGTM!

PSReadLine/Prediction.Views.cs
daxian-dbw: We should avoid filling the terminal with the list view so as to leave some space for the user to see the context information. e.g. commands that ran previously and their results.

PSReadLine/Prediction.Views.cs
daxian-dbw: Will make it a named argument.

PSReadLine/Prediction.Views.cs
daxian-dbw: Will add this comment.

PSReadLine/Prediction.Views.cs
daxian-dbw: Yeah. The heigh may become too small after navigating to an item due to the item containing a long text that spans multiple lines.

daxian-dbw:
@andschwa Thanks for the review! I submitted https://github.com/PowerShell/PSReadLine/pull/3600 to address you feedback.

ChristopherHaws:
Any idea when this will be released (even to a prerelease version)? I just ran `powershell -noprofile -command "Update-Module PSReadLine -AllowPrerelease"` and PSReadLine is still on version 2.2.6 and doesnt have these changes.

Thanks! :)

PR #3630, opened 2023-03-21
Improve the default sensitive history scrubbing to allow safe property access

### LINKED ISSUE #3318: Allow saving command to history file if it reads a property with a sensitive name
### Prerequisites

If I run this command, then PSReadLine does not save it to the history file by default, because of the sensitive name `Password`:

```PowerShell
(Import-Clixml "${Env:HOME}\credential.clixml").GetNetworkCredential().Password | Set-Clipboard
```

However, the command itself does not contain any passwords. I would like PSReadLine to save this command to the history file so that I can more easily run similar commands in future PowerShell sessions.

The existing implementation of [[Microsoft.PowerShell.PSConsoleReadLine]::GetDefaultAddToHistoryOption](https://github.com/PowerShell/PSReadLine/blob/60a48d617723f4bc07e460dfcdb550662a3ef5ea/PSReadLine/History.cs#L555) considers `Password` a [sensitive name](https://github.com/PowerShell/PSReadLine/blob/60a48d617723f4bc07e460dfcdb550662a3ef5ea/PSReadLine/History.cs#L115-L118) and returns MemoryOnly, which prevents PSReadLine from saving the command to the history file. [innerAst](https://github.com/PowerShell/PSReadLine/blob/60a48d617723f4bc07e460dfcdb550662a3ef5ea/PSReadLine/History.cs#L596) is a StringConstantExpressionAst with StringConstantType : BareWord. Its parent is a MemberExpressionAst, the grandparent is a CommandExpressionAst, and the great grandparent is a PipelineAst.

I imagine the heuristic could be changed to allow a MemberExpressionAst in which the name of the member is sensitive, if it is not at the left side of an assignment operator like `$shield.Password = "12345"`. [The VariableExpressionAst case already checks IsOnLeftSideOfAnAssignment](https://github.com/PowerShell/PSReadLine/blob/60a48d617723f4bc07e460dfcdb550662a3ef5ea/PSReadLine/History.cs#L605-L606).

daxian-dbw:
# PR Summary

Fix #3318

Improve the default sensitive history scrubbing to allow safe property access.
When the sensitive string is part of a property access:
- If this member access operation is not part of an assignment, then we consider it safe;
- Otherwise, if the right-hand side is a pipeline or a variable, we also consider it safe.

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3630)

### DIFF
--- a/PSReadLine/History.cs
+++ b/PSReadLine/History.cs
@@ -525,6 +525,22 @@ namespace Microsoft.PowerShell
             return result;
         }
 
+        private static bool IsSafePropertyUsage(Ast member)
+        {
+            bool result = false;
+
+            if (member.Parent is MemberExpressionAst memberExpr)
+            {
+                // - If the property is NOT on the left side of an assignment, then it's safe.
+                // - Otherwise, if the right-hand side is a pipeline or a variable, then we consider it safe.
+                result = !IsOnLeftSideOfAnAssignment(memberExpr, out Ast rhs)
+                    || rhs is PipelineAst
+                    || (rhs is CommandExpressionAst cmdExpr && cmdExpr.Expression is VariableExpressionAst);
+            }
+
+            return result;
+        }
+
         private static ExpressionAst GetArgumentForParameter(CommandParameterAst param)
         {
             if (param.Argument is not null)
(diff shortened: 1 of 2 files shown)


andyleejordan (approved):
Great work!

PR #3639, opened 2023-03-27
Set the current location in `PredictionClient` when it's supported

daxian-dbw:
# PR Summary

Set the current location in `PredictionClient` when it's supported.
This is in accordance with https://github.com/PowerShell/PowerShell/pull/19414, which adds the property `CurrentLocation` to `PredictionClient`, so as to make it easy for a predictor implementation to know the current location.

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3639)

### DIFF
--- a/PSReadLine/Prediction.cs
+++ b/PSReadLine/Prediction.cs
@@ -4,8 +4,10 @@ Copyright (c) Microsoft Corporation.  All rights reserved.
 
 using System;
 using System.Collections.Generic;
+using System.Reflection;
 using System.Threading.Tasks;
 using System.Management.Automation;
+using System.Management.Automation.Runspaces;
 using System.Management.Automation.Language;
 using System.Management.Automation.Subsystem.Prediction;
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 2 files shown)


PSReadLine/Prediction.cs
```diff
@@ -16,8 +18,42 @@ namespace Microsoft.PowerShell
...
+        {
+            Version ver = typeof(PSObject).Assembly.GetName().Version;
+            if (ver.Major < 7 || ver.Minor < 4)
+            {
+                return;
+            }
+
+            Type pcType = typeof(PredictionClient);
+            // Property added in 7.4
+            s_pCurrentLocation = pcType.GetProperty("CurrentLocation");
```
andyleejordan, line 39: Just since it's reflection I want to make sure I understand this properly: it is not getting the _value_ of `PredictionClient.CurrentLocation` (it can't because there's no instance here) but it's basically getting a magical pointer to that property on that type which can be used on later instances to resolve its value?

PSReadLine/Prediction.cs
```diff
@@ -16,8 +18,42 @@ namespace Microsoft.PowerShell
...
+        /// compatibility issues.
+        /// </summary>
+        private static void UpdatePredictionClient(Runspace runspace, EngineIntrinsics engineIntrinsics)
+        {
+            // Set the current location if the 'CurrentLocation' property exists.
+            if (s_pCurrentLocation is not null)
+            {
+                // Set the current location if it's a local Runspace. Otherwise, set it to null.
+                object path = runspace.RunspaceIsRemote ? null : engineIntrinsics.SessionState.Path.CurrentLocation;
+                s_pCurrentLocation.SetValue(s_predictionClient, path);
```
andyleejordan, line 54: And so because reflection is weird, this is setting the property `CurrentLocation` on the instance `s_predictionClient` to the value of `path`, just like `s_predictionClient.CurrentLocation = path` would, but avoids the issue where that property doesn't exist in previous versions.

andyleejordan (approved):
LGTM as long my reflection understanding is right.

PSReadLine/Prediction.cs
daxian-dbw: Yes, that's right, but not for getting the value of that property. It's for setting the value of that property later.

PSReadLine/Prediction.cs
daxian-dbw: Yeah, that's correct.

andyleejordan: Cool, thanks!

SeeminglyScience (approved):
LGTM!

daxian-dbw:
Thank both of you for the review!

Issue #2698, opened 2021-07-22
Temporarily disabling command history


JVimes:
## Description of the new feature/enhancement

It would be nice if there were a concise way to disable/enable recording command history. This would be useful to keep sensitive data out of the history.

I know there's a need for this feature because [hundreds of](https://stackoverflow.com/questions/13257775/powershells-clear-history-doesnt-clear-history/36900056#36900056) StackOverflow users are looking for a way to remove sensitive data from the history.

Just setting HistorySaveStyle to "SaveNothing" doesn't satisfy this. The sensitive data gets saved again when turning history back on. There's no window of time whose commands don't eventually end up in the history.

### One idea

    Disable-PSReadlineHistory
    Enable-PSReadlineHistory

The user could sandwich the sensitive command between these calls.

### Another idea

Make it part of an existing PSReadLine command. As previously mentioned, disabling/enabling via HistorySaveStyle does not work (sensitive info still gets saved).

lzybkr:
The default `AddToHistoryHandler` has some heuristics to avoid saving credentials in your history, but you can easily use your own custom handler. Personally I use:

```PowerShell
Set-PSReadLineOption -AddToHistoryHandler {
        param([string]$line)
        return $line.Length -gt 3 -and $line[0] -ne ' ' -and $line[0] -ne ';'
    }
```

This lets me put a space at the beginning of the line or a semi-colon at the end to suppress adding to history. It also avoids adding short commands.

JVimes:
That's good to learn, @lzybkr, it could be handy.

It's just hard to sell my office on PowerShell when something like this requires plumbing rather than a real, readable command.

lzybkr:
Commands are easily implemented, or you could wrap `Set-PSReadLineOption` easily enough, but I don't think they would get much use.

Disabling history briefly should feel very interactive, and entering a command to disable, then re-enable feels clunky.

JVimes:
Whatever the implementation, [it will get a lot of use](https://stackoverflow.com/a/36900056/316760). Extending Set-PSReadLineOption would be a little less readable but I'd be happy enough with it.

lzybkr:
PSReadLine has always had the capability of disabling history (via my suggestion or `Set-PSReadLineOption`) so I don't really think something new will see a lot of adoption.

Also note that my suggestion on ignoring commands starting with a space is exactly what folks coming from bash or zsh are used to.

Wrappers might look like:

```PowerShell
[...]

PR #3641, opened 2023-04-04
Improve the sensitive history scrubbing to allow retrieving token from `az`, `gcloud`, and `kubectl`

### LINKED ISSUE #3633: DefaultAddToHistoryHandler filters too aggressively
### Prerequisites

So, per https://github.com/PowerShell/PSReadLine/issues/1459 I don't really think this list of things to filter is a good heuristic. I often do things like:

```
$env:token = command-which-outputs-token ; command-which-requires-token-to-function
```

This filter breaks any sort of idiom like that since depending on the context it might be $env:secret, $env:password, or whatever. I just disable this and I don't think it should be on by default.

Turn off this heuristic, it is very surprising behavior and doesn't really accomplish the intended goal.

daxian-dbw:
# PR Summary

Fix #3633
Improve the sensitive history scrubbing to allow retrieving token from `az`, `gcloud`, and `kubectl`.
The default sensitive history scrubbing was updated to handle retrieving token with [gcloud](https://cloud.google.com/sdk/gcloud/reference/auth/print-access-token), with Azure ([both az cli and Az PowerShell](https://learn.microsoft.com/en-us/azure/healthcare-apis/get-access-token?tabs=azure-cli)), and with `kubectl` (both [`get secret` and `describe secret`](https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kubectl/)).

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3641)

### DIFF
--- a/PSReadLine/History.cs
+++ b/PSReadLine/History.cs
@@ -128,7 +128,8 @@ namespace Microsoft.PowerShell
             "Set-SecretVaultDefault",
             "Test-SecretVault",
             "Unlock-SecretVault",
-            "Unregister-SecretVault"
+            "Unregister-SecretVault",
+            "Get-AzAccessToken",
         };
 
         private void ClearSavedCurrentLine()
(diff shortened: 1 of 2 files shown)


andyleejordan (approved):
I'm ok with this but wary of the precedent it sets for us to get this right not only for these commands, but for every possible ask in the future. Like it's a nice improvement...but I could see this growing huge if we don't have a plan to decide which commands we don't scrub (and do scrub). Perhaps we should have a more general way to allow users to add exceptions?

daxian-dbw:
> Perhaps we should have a more general way to allow users to add exceptions?

The user can already add exceptions, or customized scrubbing by `Set-PSReadLineOption -AddToHistoryHandler { <a script block> }` (see [doc on this parameter](https://learn.microsoft.com/en-us/powershell/module/psreadline/set-psreadlineoption?view=powershell-7.3#-addtohistoryhandler)). Here is an example that applies additional rules on top of the default: https://github.com/PowerShell/PSReadLine/issues/2698#issuecomment-1439942616. User can also choose to disable the scrubbing by `Set-PSReadLineOption -AddToHistoryHandler $null`.

The improvement made to the default scrubbing algorithm is for general scenarios that many users will run into. Otherwise, we will ask the user to apply the `AddToHistoryHandler` for their individual needs.

PR #3644, opened 2023-04-04
Fix to force refreshing suggestion in the inline view when plugin is in use

daxian-dbw:
# PR Summary

Fix to force refreshing suggestion in the inline view when plugin is in use, even if the current history suggestion can still apply to the new input.

Scenario:
When the prediction source is `HistoryAndPlugin`, we favor plugin results over history results in the inline view. For an command-line input, if the plugin doesn't return anything, the history result will be used.

Today, when an inline view prediction is displayed, if you keep typing and the existing suggestion can still apply to the resulted new input, PSReadLine will continue to use the existing suggestion text instead of refreshing the suggestion.
So, when the history suggestion is used for inline view, if the new input resulted from your typing can still fit in the existing history suggestion, the plugin won't be triggered, even though it now may return more relevant result.

This PR fixes the above issue:
- when the prediction source is `History` only, the behavior is not changed
- when the prediction source is `HistoryAndPlugin`
   - the existing suggestion is from a plugin, then the current behavior is not changed -- if the existing suggestion can still apply to the new input, PSReadLine continues to use it instead of refreshing suggestion.
   - the existing suggestion is from history, then new input will force refreshing the suggestion, unless the new input is resulted by the user accepting the next word of the existing history suggestion.

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3644)

### DIFF
--- a/PSReadLine/Prediction.Views.cs
+++ b/PSReadLine/Prediction.Views.cs
@@ -1082,9 +1082,24 @@ namespace Microsoft.PowerShell
                 {
                     _inputText = userInput;
 
-                    if (_suggestionText == null || _suggestionText.Length <= userInput.Length ||
-                        _lastInputText.Length > userInput.Length ||
-                        !_suggestionText.StartsWith(userInput, _singleton._options.HistoryStringComparison))
+                    bool needToRefresh = _suggestionText == null
+                        || _suggestionText.Length <= userInput.Length
+                        || _lastInputText.Length > userInput.Length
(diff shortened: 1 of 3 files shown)


andyleejordan (approved):
LGTM! Thanks for the thorough commentary and tests, made it much easier to follow and double-check.

PR #3667, opened 2023-05-02
Support tooltip rendering in the prediction list view

### LINKED ISSUE #3638: Support rendering the tooltip in PSReadLine list view
### Prerequisites

Today, the `PredictiveSuggestion` type has the `Tooltip` property, but PSReadLine doesn't use it in its rendering.
For completeness, it should render the tooltip, and ideally the tooltips should be displayed in a popup terminal UI window when the user moves the focus to an item in the list, in the form similar to this: https://github.com/waf/CSharpRepl#c-repl

_No response_

daxian-dbw:
# PR Summary

Support tooltip rendering in the prediction list view. Fix #3638.

- By default, the list view can use up to 4 lines for tooltip rendering:

   tooltip1

- For tooltip that spans more than 4 lines, we allow the user to see the full tooltip in the alternate screen buffer with the keybinding <kbd>F4</kbd>

   tooltip2

- When terminal size shrinks, the maximum lines can be used for tooltip also decreases -- it uses up to 2 lines for tooltip when the max list height is reduced to 5.

   tooltip3

- When the terminal size is even smaller and the max list view height becomes 3, it uses up to 1 line for tooltip.

   tooltip4

- You can change the tooltip color by `Set-PsreadLineOption -Colors @{ ListPredictionTooltip = "green" }`, for example.

   tooltip5

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [x] Doc Issue filed: 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3667)

### DIFF
--- a/PSReadLine/Prediction.Views.cs
+++ b/PSReadLine/Prediction.Views.cs
@@ -215,6 +215,7 @@ namespace Microsoft.PowerShell
             // List view constants.
             internal const int ListViewMaxHeight = 10;
             internal const int ListViewMaxWidth = 100;
+            internal const int TooltipMaxHeight = 4;
             internal const int SourceMaxWidth = 15;
 
             // Minimal window size.
@@ -236,6 +237,11 @@ namespace Microsoft.PowerShell
             private int _maxViewHeight;
             // The actual height of the list view that is currently rendered.
             private int _listViewHeight;
+            // The max number of physical lines for rendering tooltip.
+            private int _maxTooltipHeight;
+            // The actual number of lines used for rendering the tooltip.
+            private int _tooltipHeight;
+
             // The actual width of the list view that is currently rendered.
             private int _listViewWidth;
             // An index pointing to the item that is shown in the first slot of the list view.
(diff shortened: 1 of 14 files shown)


StevenBucher98:
Love the UI! Is there a way to disable the tooltip?

daxian-dbw:
Same as disabling tooltip in menu completion today: `Set-PSReadLineOption -showToolTips:$false`

StevenBucher98 (approved):
Can't give a proper code review but the design and UX look great to me!

PSReadLine/Prediction.Views.cs
```diff
@@ -865,6 +917,146 @@ static StringBuilder AppendColor(StringBuilder buffer, string colorToUse, ref st
...
             }
 
+            /// <summary>
+            /// Generate the rendering text for the tooltip.
+            /// </summary>
+            private int RenderTooltip(string tooltip, List<StringBuilder> consoleBufferLines, ref int currentLogicalLine)
+            {
+                const int LengthOfLeadingPart = 6;
+                const string IndicatorSymbol = ">>";
+                const string MsgForViewAll = "(<F4> to view all)";
```
andyleejordan, line 927: Just curious if we do this elsewhere. I suspect that key binding is customizable by the user, but this won't update. Is that the case?

test/ListViewTooltipTest.cs
andyleejordan, line 1: Nice test.

andyleejordan (approved):
Looks good to me!

PSReadLine/Prediction.Views.cs
daxian-dbw: Right, the <F4> key could already be bound by a user and this message currently won't update in that case.
We don't do this elsewhere -- first time advertising a key-binding in the rendering. Will collect the feedback and see how to best deal with it.

daxian-dbw: Opened #3668 to track this.

Issue #3745, opened 2023-07-15
Add "ConsoleBouncer" option, to terminate stragglers (to fix ctrl+c)


jazzdelightsme:
## Description of the new feature/enhancement

Ctrl+c cancellation does not work really well on Windows. **Each** process attached to a console receives a ctrl+c signal (as opposed to just the active shell), and *sometimes,* when a shell has launched some large tree of child processes (imagine a build system, for example), some processes do not exit (perhaps due to races between process creation and console attachment), leaving multiple processes all concurrently trying to consume console input, which Does Not Work Well(TM). It's usually not too bad when `cmd.exe` is your shell, because you can just keep mashing on <kbd>ctrl</kbd>+<kbd>c</kbd> and usually get back to a usable state. But it's considerably worse in PowerShell, because PSReadLine temporarily disables ctrl+c signals when waiting at the prompt, which can lead to the console being completely unrecoverable.

In the [ConsoleBouncer](https://github.com/jazzdelightsme/ConsoleBouncer) project, I sought to solve the problem with a PowerShell module. When loaded, the module installs a ctrl+c handler, and takes inventory of processes currently attached to the console; these are the "allowed processes". When a ctrl+c signal is received, all processes not in the "allowed list" (if any) are terminated.

This approach has some weaknesses. For one, it assumes that <kbd>ctrl</kbd>+<kbd>c</kbd> means "kill everything and get back to the shell", but that assumption is not always true. The Windows console debugger processes `kd.exe` and `cdb.exe`, for example, handle <kbd>ctrl</kbd>+<kbd>c</kbd> (interpreting it as a signal to break into the target). So ConsoleBouncer is perhaps mostly-good-enough for most people who need it, probably, but is definitely too kludgy for “everyone”. ***But*** PSReadLine could implement something similar that would actually solve the problem more directly.

The defining characteristic of a borked shell is that in its mind, it is just sitting at the prompt, waiting for user input, where at the same time, there are other processes attached to the console, also consuming input (leading to wild and unpredictable behavior, as different keystrokes go to different processes). The ConsoleBouncer module works by receiving a ctrl+c signal, and it does not know what PSReadLine in the shell (or anybody else) is up to. It assumes that ctrl+c means “everybody out!” (And it’s that assumption that is “mostly good enough”, but not really completely true.)

[...]

PR #3764, opened 2023-07-30
Add TerminateStragglers option (Windows)

jazzdelightsme:
# PR Summary

Addresses #3745

There is a lot more detail about the problem this PR solves in the code comments and the linked Issue. In short, this PR:

* Adds a new "TerminateStragglers" option.
* When enabled, existing console-attached PIDs are gathered into an "allow" list (on Windows only).
* When the shell informs us that it is time to wait for input, right before we disable ctrl+c signals, (if enabled) we ensure that the shell will not be broken by other console-attached children that may also be attempting to read input.

How tested: added printfs and manually triggered the problem scenario, to verify that the code works as expected.

There are a few open questions; I will leave some comments in the PR (for example: if someone on non-Windows enables the option, should we write a warning, or just smile and accept it?).

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: https://github.com/MicrosoftDocs/PowerShell-Docs/issues/10346

 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/PowerShell/PSReadLine/pull/3764&drop=dogfoodAlpha

### DIFF
--- a/PSReadLine/Options.cs
+++ b/PSReadLine/Options.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Management.Automation;
 using System.Reflection;
+using System.Runtime.InteropServices;
 using System.Threading;
 using Microsoft.PowerShell.PSReadLine;
 
@@ -165,6 +166,14 @@ namespace Microsoft.PowerShell
                     {
                         Options.SetColor(property, e.Value);
                     }
+                }
+            }
+            if (options._terminateStragglers.HasValue)
+            {
+                Options.TerminateStragglers = options.TerminateStragglers;
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    PlatformWindows.SetTerminateStragglers(Options.TerminateStragglers);
                 }
             }
         }
--- a/PSReadLine/Options.cs
+++ b/PSReadLine/Options.cs
@@ -168,12 +168,12 @@ namespace Microsoft.PowerShell
                     }
                 }
             }
-            if (options._terminateStragglers.HasValue)
+            if (options._terminateOrphanedConsoleApps.HasValue)
             {
-                Options.TerminateStragglers = options.TerminateStragglers;
                 if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                 {
-                    PlatformWindows.SetTerminateStragglers(Options.TerminateStragglers);
+                    Options.TerminateOrphanedConsoleApps = options.TerminateOrphanedConsoleApps;
+                    PlatformWindows.SetTerminateOrphanedConsoleApps(Options.TerminateOrphanedConsoleApps);
                 }
             }
         }
--- a/PSReadLine/Options.cs
+++ b/PSReadLine/Options.cs
@@ -178,7 +178,11 @@ namespace Microsoft.PowerShell
                 }
                 else
                 {
-                    throw new PlatformNotSupportedException(string.Format(CultureInfo.CurrentUICulture, PSReadLineResources.OptionNotSupportedOnNonWindows, nameof(Options.TerminateOrphanedConsoleApps)));
+                    throw new PlatformNotSupportedException(
+                        string.Format(
+                            CultureInfo.CurrentUICulture,
+                            PSReadLineResources.OptionNotSupportedOnNonWindows,
+                            nameof(Options.TerminateOrphanedConsoleApps)));
                 }
             }
         }
--- a/PSReadLine/Options.cs
+++ b/PSReadLine/Options.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
 using System.Management.Automation;
 using System.Reflection;
 using System.Runtime.InteropServices;
@@ -174,6 +175,10 @@ namespace Microsoft.PowerShell
                 {
                     Options.TerminateOrphanedConsoleApps = options.TerminateOrphanedConsoleApps;
                     PlatformWindows.SetTerminateOrphanedConsoleApps(Options.TerminateOrphanedConsoleApps);
+                }
+                else
+                {
+                    throw new PlatformNotSupportedException(string.Format(CultureInfo.CurrentUICulture, PSReadLineResources.OptionNotSupportedOnNonWindows, nameof(Options.TerminateOrphanedConsoleApps)));
                 }
             }
         }
--- a/PSReadLine/PlatformWindows.cs
+++ b/PSReadLine/PlatformWindows.cs
@@ -210,6 +210,7 @@ static class PlatformWindows
                 // gone, so we'll get rid of them:
                 TerminateStragglers();
             }
+
             SetOurInputMode();
         }
     }
@@ -628,7 +629,7 @@ static class PlatformWindows
             numPids = native_GetConsoleProcessList(pids, (uint) size);
         }
 
-        if ((0 == numPids) || (numPids > size))
+        if (0 == numPids || numPids > size)
         {
             return null; // no TerminateOrphanedConsoleApps for you, sorry
         }
@@ -683,27 +684,27 @@ static class PlatformWindows
             int HrInit();
 
             [PreserveSig]
-            int AddTab( IntPtr hwnd );
+            int AddTab(IntPtr hwnd);
 
             [PreserveSig]
-            int DeleteTab( IntPtr hwnd );
+            int DeleteTab(IntPtr hwnd);
 
             [PreserveSig]
-            int ActivateTab( IntPtr hwnd );
+            int ActivateTab(IntPtr hwnd);
 
             [PreserveSig]
-            int SetActiveAlt( IntPtr hwnd );
+            int SetActiveAlt(IntPtr hwnd);
 
             // ITaskbarList2
             [PreserveSig]
-            int MarkFullscreenWindow( IntPtr hwnd, [MarshalAs(UnmanagedType.Bool)] bool fFullscreen );
+            int MarkFullscreenWindow(IntPtr hwnd, [MarshalAs(UnmanagedType.Bool)] bool fFullscreen);
 
             // ITaskbarList3
             [PreserveSig]
-            int SetProgressValue( IntPtr hwnd, UInt64 ullCompleted, UInt64 ullTotal );
+            int SetProgressValue(IntPtr hwnd, UInt64 ullCompleted, UInt64 ullTotal);
 
             [PreserveSig]
-            int SetProgressState( IntPtr hwnd, TaskbarStates state );
+            int SetProgressState(IntPtr hwnd, TaskbarStates state);
 
             // N.B. for copy/pasters: we've left out the rest of the ITaskbarList3 methods...
         }
@@ -748,7 +749,7 @@ static class PlatformWindows
                 {
                     proc = Process.GetProcessById((int) pid);
                 }
-                catch( ArgumentException )
+                catch (ArgumentException)
                 {
                     // Ignore it: process could be gone, or something else that we
                     // likely can't do anything about it.
@@ -765,8 +766,6 @@ static class PlatformWindows
 
     [DllImport("kernel32.dll")]
     internal static extern ulong GetTickCount64();
-
-    private const int DefaultGraceMillis = 1000;
 
     private static int MillisLeftUntilDeadline(ulong deadline)
     {
@@ -785,8 +784,8 @@ static class PlatformWindows
         return (int) diff;
     }
 
+    private const int DefaultGraceMillis = 1000;
     private const int MaxRounds = 2;
-    private const bool ClearProgress = true;
 
     //
     //                         TerminateOrphanedConsoleApps
@@ -901,8 +900,8 @@ static class PlatformWindows
         int round = 0;
         int killAttempts = 0;
 
-        while ((round++ < MaxRounds) &&
-               (GatherStragglers(procsToTerminate) > 0))
+        while (round++ < MaxRounds &&
+               GatherStragglers(procsToTerminate) > 0)
         {
             // We'll give them up to GracePeriodMillis for them to exit on their
             // own, in case they actually did receive the original ctrl+c, and are
@@ -933,27 +932,24 @@ static class PlatformWindows
 
         // In forcible termination scenarios, if there was a child updating the terminal's
         // progress state, it may be left stuck that way... we can clear that out.
-        if (ClearProgress)
+        uint consoleMode = GetConsoleOutputMode();
+        if (ItLooksLikeWeAreInTerminal())
         {
-            uint consoleMode = GetConsoleOutputMode();
-            if (ItLooksLikeWeAreInTerminal())
+            // We can use the [semi-]standard OSC sequence:
+            // https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC
+            if (0 != (consoleMode & (uint) ENABLE_VIRTUAL_TERMINAL_PROCESSING))
             {
-                // We can use the [semi-]standard OSC sequence:
-                // https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC
-                if (0 != (consoleMode & (uint) ENABLE_VIRTUAL_TERMINAL_PROCESSING))
-                {
-                    // Use "bell" if we actually tried to whack anything.
-                    char final = (killAttempts > 0) ? '\a' : '\x001b';
-                    Console.Write("\x001b]9;4;0;0" + final);
-                }
+                // Use "bell" if we actually tried to whack anything.
+                string final = (killAttempts > 0) ? "\a" : "\x001b\\";
+                Console.Write("\x001b]9;4;0;0" + final);
             }
-            else
+        }
+        else
+        {
+            IntPtr hwnd = GetConsoleWindow();
+            if (hwnd != IntPtr.Zero)
             {
-                IntPtr hwnd = GetConsoleWindow();
-                if (hwnd != IntPtr.Zero)
-                {
-                    int ret = TaskbarProgress.SetProgressState(hwnd, TaskbarStates.NoProgress);
-                }
+                int ret = TaskbarProgress.SetProgressState(hwnd, TaskbarStates.NoProgress);
             }
         }
     }
--- a/PSReadLine/PlatformWindows.cs
+++ b/PSReadLine/PlatformWindows.cs
@@ -760,12 +760,10 @@ static class PlatformWindows
         }
     }
 
-    private static readonly Lazy<uint> _myPid = new Lazy<uint>(() =>
+    private static readonly Lazy<uint> _myPid = new(() =>
     {
-        using (var me = Process.GetCurrentProcess())
-        {
-            return (uint)me.Id;
-        }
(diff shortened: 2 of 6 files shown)


PSReadLine/Options.cs
```diff
@@ -167,6 +168,14 @@ private void SetOptionsInternal(SetPSReadLineOption options)
                     }
                 }
             }
+            if (options._terminateStragglers.HasValue)
+            {
+                Options.TerminateStragglers = options.TerminateStragglers;
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    PlatformWindows.SetTerminateStragglers(Options.TerminateStragglers);
+                }
```
jazzdelightsme, line 186: >} [](http://example.com/codeflow?start=16&length=1)

If not on Windows, do we want to notify the user that this doesn't actually do anything? #Pending

PSReadLine/PlatformWindows.cs
```diff
@@ -614,4 +611,350 @@ public override void BlankRestOfLine()
...
+                }
+            }
+        }
+        return procsToTerminate.Count;
+    }
+
+    [DllImport("kernel32.dll")]
+    internal static extern ulong GetTickCount64();
+
+    private const int DefaultGraceMillis = 1000;
```
jazzdelightsme: >DefaultGraceMillis = 1000; [](http://example.com/codeflow?start=22&length=26)

We *could* make this configurable... but I don't think it's necessary; and the downside is that it's more cognitive load for the user ("what is this?") #Pending

PSReadLine/PlatformWindows.cs
```diff
@@ -614,4 +611,350 @@ public override void BlankRestOfLine()
...
+        {
+            // Should not ever actually happen...
+            diff = DefaultGraceMillis;
+        }
+
+        return (int) diff;
+    }
+
+    private const int MaxRounds = 2;
+    private const bool ClearProgress = true;
```
jazzdelightsme: >private const bool ClearProgress = true; [](http://example.com/codeflow?start=4&length=40)

Similarly, would we want this to be configurable? #Pending

PSReadLine/Options.cs
jazzdelightsme: (actually, the Options-setting line should be inside the `if` block, too, and that might be enough)

daxian-dbw: We should throw an `NotSupportedException` on non-Windows.
Can you please add an error message to `PSReadLineResources.resx` and a new property for that message in `PSReadLineResources.Designer.cs`?

PSReadLine/PlatformWindows.cs
```diff
@@ -201,6 +202,14 @@ internal static void Init(ref ICharMap charMap)
...
             }
 
+            // Is the TerminateOrphanedConsoleApps feature enabled?
+            if (_allowedPids != null)
+            {
+                // We are about to disable Ctrl+C signals... so if there are still any
+                // console-attached children, the shell will be broken until they are
+                // gone, so we'll get rid of them:
+                TerminateStragglers();
+            }
```
daxian-dbw, line 212: Minor style suggestion -- add an empty line after the closing bracket.
```suggestion
            }

```

PSReadLine/PlatformWindows.cs
```diff
@@ -614,4 +611,350 @@ public override void BlankRestOfLine()
...
+    // If the TerminateOrphanedConsoleApps option is enabled, this is the list of PIDs
+    // that are allowed to stay attached to the console (effectively the current process
+    // plus ancestors).
+    private static uint[] _allowedPids;
+
+    internal static void SetTerminateOrphanedConsoleApps(bool enabled)
+    {
+        if (enabled)
+        {
+            _allowedPids = GetConsoleProcessList();
```
daxian-dbw, line 681: Why not use `HashSet<uint>` for `_allowedPids`?

PSReadLine/PlatformWindows.cs
```diff
@@ -614,4 +611,350 @@ public override void BlankRestOfLine()
...
+        [Guid("ea1afb91-9e28-4b86-90e9-9e9f8a5eefaf")]
+        [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
+        private interface ITaskbarList3
+        {
+            // ITaskbarList
+            [PreserveSig]
+            int HrInit();
+
+            [PreserveSig]
+            int AddTab( IntPtr hwnd );
```
daxian-dbw: Minor style thing. Can you please change the following members too?
```suggestion
            int AddTab(IntPtr hwnd);
```

PSReadLine/PlatformWindows.cs
```diff
@@ -614,4 +611,350 @@ public override void BlankRestOfLine()
...
+        foreach (var pid in currentPids)
+        {
+            if (!_allowedPids.Contains(pid))
+            {
+                Process proc = null;
+                try
+                {
+                    proc = Process.GetProcessById((int) pid);
+                }
+                catch( ArgumentException )
```
daxian-dbw:
```suggestion
                catch(ArgumentException)
```

PSReadLine/PlatformWindows.cs
daxian-dbw: Agree to hard code it, at least to start with.

PSReadLine/PlatformWindows.cs
daxian-dbw: There seems no good reason to not cleanup the progress bar or indicator, so I think we should make this the default behavior, not configurable.
I think this can be removed.

PSReadLine/PlatformWindows.cs
```diff
@@ -614,4 +611,350 @@ public override void BlankRestOfLine()
...
+        return (int) diff;
+    }
+
+    private const int MaxRounds = 2;
+    private const bool ClearProgress = true;
+
+    //
+    //                         TerminateOrphanedConsoleApps
+    //
+    // This feature works around a bad interaction on Windows between:
```
daxian-dbw, line 853: Thank you for the default comments about this feature!

PSReadLine/PlatformWindows.cs
```diff
@@ -614,4 +611,350 @@ public override void BlankRestOfLine()
...
+        // because in my specific scenario the process that was spawning processes got
+        // taken out with the original ctrl+c signal.
+        //
+        // If it takes more than a few rounds of cleanup, we may be in some kind of
+        // pathological situation, and we'll bow out.
+        int round = 0;
+        int killAttempts = 0;
+
+        while ((round++ < MaxRounds) &&
+               (GatherStragglers(procsToTerminate) > 0))
```
daxian-dbw:
```suggestion
        while (round++ < MaxRounds &&
               GatherStragglers(procsToTerminate) > 0)
```

PSReadLine/PlatformWindows.cs
```diff
@@ -614,4 +611,350 @@ public override void BlankRestOfLine()
...
+                process.Dispose();
+            }
+        } // end retry loop
+
+        // In forcible termination scenarios, if there was a child updating the terminal's
+        // progress state, it may be left stuck that way... we can clear that out.
+        if (ClearProgress)
+        {
+            uint consoleMode = GetConsoleOutputMode();
+            if (ItLooksLikeWeAreInTerminal())
```
daxian-dbw: Maybe it's enough to just check whether "ENABLE_VIRTUAL_TERMINAL_PROCESSING" is turned on? Could console host or another console emulator support the progress indicator too?

PSReadLine/PlatformWindows.cs
```diff
@@ -614,4 +611,350 @@ public override void BlankRestOfLine()
...
+                    char final = (killAttempts > 0) ? '\a' : '\x001b';
+                    Console.Write("\x001b]9;4;0;0" + final);
+                }
+            }
+            else
+            {
+                IntPtr hwnd = GetConsoleWindow();
+                if (hwnd != IntPtr.Zero)
+                {
+                    int ret = TaskbarProgress.SetProgressState(hwnd, TaskbarStates.NoProgress);
```
daxian-dbw: Two questions here:
1. Did you verify if this code path works in PowerShell 7.2 and 7.3? I vaguely remember .NET doesn't support COM interop as in .NET Framework, so wonder if this works.
2. It looks you assume a straggler process won't do the task bar progress when it runs in WT (or when it's using VT processing??), but could it be possible that a process does both the progress indicator and the task bar progress?

PSReadLine/PlatformWindows.cs
```diff
@@ -614,4 +611,350 @@ public override void BlankRestOfLine()
...
+        {
+            uint consoleMode = GetConsoleOutputMode();
+            if (ItLooksLikeWeAreInTerminal())
+            {
+                // We can use the [semi-]standard OSC sequence:
+                // https://conemu.github.io/en/AnsiEscapeCodes.html#ConEmu_specific_OSC
+                if (0 != (consoleMode & (uint) ENABLE_VIRTUAL_TERMINAL_PROCESSING))
+                {
+                    // Use "bell" if we actually tried to whack anything.
+                    char final = (killAttempts > 0) ? '\a' : '\x001b';
```
daxian-dbw: According to https://conemu.github.io/en/AnsiEscapeCodes.html#OSC_Operating_system_commands:
> **Note.** These codes may ends with ‘ESC\’ (two symbols - ESC and BackSlash) or ‘BELL’ (symbol with code \x07, same as ‘^a’ in *nix). For simplifying, endings in the following table marked as ‘ST’.

So, it seems you should add `"\x1b\\"` at the end instead of just a `ESC` symbol, no?

PSReadLine/PlatformWindows.cs
```diff
@@ -614,4 +611,350 @@ public override void BlankRestOfLine()
...
+        uint numPids = native_GetConsoleProcessList(pids, (uint) size);
+
+        if (numPids > size)
+        {
+            size = (int) numPids + 10; // a bit extra, since we may be racing attaches.
+            pids = new uint[size];
+            numPids = native_GetConsoleProcessList(pids, (uint) size);
+        }
+
+        if ((0 == numPids) || (numPids > size))
```
daxian-dbw:
```suggestion
        if (0 == numPids || numPids > size)
```

PSReadLine/PlatformWindows.cs
```diff
@@ -614,4 +611,350 @@ public override void BlankRestOfLine()
...
+    internal enum TaskbarStates
+    {
+        NoProgress    = 0,
+        Indeterminate = 0x1,
+        Normal        = 0x2,
+        Error         = 0x4,
+        Paused        = 0x8,
+    }
+
+    internal static class TaskbarProgress
```
daxian-dbw, line 706: Do you have an example screenshot of this task bar progress? Is it a winform thing?

PSReadLine/PlatformWindows.cs
daxian-dbw: Can you please move this const value to be together with `MaxRounds`?

PSReadLine/PlatformWindows.cs
```diff
@@ -614,4 +611,350 @@ public override void BlankRestOfLine()
...
+    // situation with cmd.exe as your shell, you can just mash on ctrl+c for a while and
+    // kill all the stragglers manually; but if you have PSReadLine loaded, your shell is
+    // borked, and you are stuck. You CAN recover, IF you can track down and kill all the
+    // straggler processes manually.
+    //
+    // So when enabled, this feature does that for you: it kills all those straggler
+    // processes, right before we disable ctrl+c signals and wait for user input, ensuring
+    // that the user has a usable shell.
+    //
+    // Note that GUI processes do not attach to the console, so if you have launched
```
daxian-dbw, line 887: I need to build the changes and see what it does when PowerShell is working with child process with stdin redirected. Haven't got the chance to do so, but will reply back if I find any issue with that.

daxian-dbw (commented):
@jazzdelightsme Sorry for the late review. It looks good overall. Left some comments.

PSReadLine/PlatformWindows.cs
jazzdelightsme: It is extraordinarily unlikely to have more than a small handful of processes in the allow list (I would bet P99.99 is < 8). It's simpler and faster to just scan a list that small.

PSReadLine/PlatformWindows.cs
jazzdelightsme: Maybe you figured it out when you got to where this is used below, but it's the same thing as what you can drive with the "ESC ] 9 ; 4" VT sequence (but it works even where that sequence isn't supported).

I can figure out how to produce a .gif if that doesn't clear it up.

PSReadLine/PlatformWindows.cs
jazzdelightsme: Yes, other terminals *might* support that sequence... but unfortunately there is no way to query which sequences are supported (there are constant discussions on the Terminal repo about this, it seems like--it's a hard, unsolved problem). So if we are not in Terminal, I think the safest thing to do is just talk to the shell directly.

PSReadLine/PlatformWindows.cs
jazzdelightsme: Huh... I tested that, but I agree it's better to follow the doc.

PSReadLine/PlatformWindows.cs
jazzdelightsme: 1: I tried 7.2.13, 7.3.6, and 7.4.0-preview.4, in both Terminal and conhost. I added a `Console.WriteLine` to print out the `ret` value, and it was always 0 (and progress was cleared), so it seems to work. 🤷

2: both the VT sequence and the COM API are two different paths to the same thing; so we do VT if we believe it will work, else we fall back to COM. Make sense?

PSReadLine/PlatformWindows.cs
daxian-dbw: > but it's the same thing as what you can drive with the "ESC ] 9 ; 4" VT sequence

Got it. Can you please add a comment in code to explain they are the same thing?

> I added a Console.WriteLine to print out the ret value, and it was always 0 (and progress was cleared), so it seems to work.

No need to produce a GIF. I was curious because I didn't know they are the same thing. Now that you have tested it and confirmed it works, we are good.

PSReadLine/PlatformWindows.cs
daxian-dbw: I played with the changes and found an issue:

> $p = Start-Process pwsh -ArgumentList '-c Write-Host start $pid; sleep -seconds 30; Write-Host stop' -NoNewWindow -passThru

After the option is enabled, when starting a console application like above, the process will be terminated immediately even though it doesn't really consume console input.

I don't think this is the right behavior. In this case, the new process shouldn't be considered as an "orphaned process" from the user's viewpoint.

Maybe we should check if a console-attached process is actually an orphan by verifying if its parent process id still points to an alive process that was created earlier.

daxian-dbw: I played with the changes and found an issue:

> $p = Start-Process pwsh -ArgumentList '-c Write-Host start $pid; sleep -seconds 30; Write-Host stop' -NoNewWindow -passThru

After the option is enabled, when starting a console application like above, the process will be terminated immediately even though it doesn't really consume console input.

I don't think this is the right behavior. In this case, the new process shouldn't be considered as an "orphaned process" from the user's viewpoint.

Maybe we should check if a console-attached process is actually an orphan by verifying if its parent process id still points to an alive process that was created earlier.

daxian-dbw (changes requested):

PSReadLine/PlatformWindows.cs
jazzdelightsme: Actually, this is *exactly* the sort of process that we want to kill. Although you came up with a *mostly* benign example (it only spews a few pieces of "junk" into the middle of your interactive session), this process is very much capable of hosing your shell (more than just writing output while you are at the prompt).

Here is a modified example, which demonstrates that it is connected to the console's input (even without PSReadLine loaded, it will have weird behavior):

`$p = Start-Process pwsh -ArgumentList '-NoProfile -c Write-Host start $pid; write-host "input redirected? $([Console]::IsInputRedirected)" ; write-host "output redirected? $([Console]::IsOutputRedirected)" ; Write-Host "give me some input: " ; $userInput = [Console]::ReadLine() ; Write-Host "input was: $userInput" ;  sleep -seconds 30; Write-Host stop' -NoNewWindow -passThru`

Maybe you were thinking "well, it *could* read input, but if it *doesn't*, then it should be allowed". Unfortunately there is (of course) no way to tell if an arbitrary process is *going* to read input or not. But the fact that it *can* (while we are also waiting for input) qualifies it for termination.

There are some mitigating factors here:

1. (The biggest, IMO): it seems pretty unlikely that someone would do something like this directly from the command line. If they do it in a script, it is much more likely that the script is doing something sensible with the child process, and not returning to the prompt "early" (before the child process has exited), in which case everything would be fine.
2. This new option is off by default, so it definitely won't start causing a bunch of people to randomly have their processes killed.

Also, I think this aligns well with documented guidance for how/when to use `Start-Process` ([https://github.com/MicrosoftDocs/PowerShell-Docs/issues/6239](https://github.com/MicrosoftDocs/PowerShell-Docs/issues/6239)).

daxian-dbw: > Maybe you were thinking "well, it could read input, but if it doesn't, then it should be allowed". Unfortunately there is (of course) no way to tell if an arbitrary process is going to read input or not. But the fact that it can (while we are also waiting for input) qualifies it for termination.

I think the criteria should be whether a console-attached process is an orphan.

It's OK if a user explicitly starts a process from PowerShell that also consumes console input -- they are allowed to shoot their own foot. But we should not prevent a user from doing so to just run some background work.

The issue we want to address is to terminate the orphan processes spawned by a build process that was terminated, so we should make sure the changes doesn't go beyond that and break a legit scenario.
- for the target issue we want to address, it's desired for the user to not know a thing about PSReadLine killing those orphan processes, because they should have been gone when the user pressed `ctrl+c` to terminate the _build_ parent process.
- for the side effect that cause `start-process` to not work as expected, it will result in lots of troubles for the user to diagnose why the process was killed.

jazzdelightsme: Okay, my personal feeling is "I wouldn't like that"... but I understand that it could be difficult for someone to figure out why such a process was killed; and given that it should not be a problem for the scenario I am trying to solve, I am willing to compromise :D so I added the check.

daxian-dbw: Thank you. I will do another review soon (first thing tomorrow morning if cannot be earlier :))

daxian-dbw (approved):
LGTM. Thanks!

PSReadLine/PlatformWindows.cs
daxian-dbw: Immediate children may have children processes that accidentally derives the standard input (which I agree is a bad thing to do), so ideally we should check the parent id of the console-attached process to make sure it still points to an alive process that was created earlier.

But I think we can just check for immediate children for simplicity, and wait for feedback.

jazzdelightsme:
Thank you @daxian-dbw! QQ: I am having trouble finding a "roadmap" Discussion/Issue. I'm wondering what the shipping vehicle for this change will be; can you point me at the roadmap? TIA!

daxian-dbw:
We don't have a published roadmap but we plan to get the 2.3 GA out along with PS 7.4 GA.

PR #3786, opened 2023-08-20
Windows keyboard layout handling: get the current layout from the parent terminal process

### LINKED ISSUE #1393: Ctrl-C doesn't work if PowerShell was started with Russian keyboard layout

ForNeVeR:
# PR Summary

This is my another attempt on #1393, with a different approach this time.

Fixes #1393 in my environment _somewhat_: we'll use the proper current keyboard layout from the terminal process, which means the `Ctrl+C` will work in English layout and won't work in Cyrillic. But that's a step forward, since before it was not working at all if you started the process in non-English layout.

The terminal process lookup is _a heuristic (since there's no official way to determine which process provides a terminal for you)_ and won't work in 100% of cases (because in some convoluted cases, the process/terminal/console ownership structure may be different). But it should be a good and stable improvement for most of the PowerShell use cases.

After we've got the parent terminal process (which I currently define as _a parent process in the process tree that has its main window visible_), we ask the OS of its current keyboard layout, and then pass it to `ToUnicodeEx` (since `ToUnicode` is hard-coded to always cache the wrong keyboard layout — the one we've called it the first time with).

Open issues: this only works well in Windows Terminal, IntelliJ and VSCode, but doesn't work in conhost. Perhaps let's leave it as it is for now?

## PR Checklist

- [x] Summarized changes
- [x] ~~Make sure you've added one or more new tests~~
  - Impossible, unfortunately. This kind of behavior is very hard to test.
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 
    - **I'm not sure, folks, please help me to figure it out**

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/PowerShell/PSReadLine/pull/3786)


ForNeVeR:
Friendly ping. Folks, could you please help with the review here?

daxian-dbw:
@jazzdelightsme, if you have time, can you please help review this PR? Thanks!

lzybkr:
It's been a very long time since I last looked at this problem, but the code and approach seem sound to me.

It'd be nice to see it fixed for conhost since folks still use it, but knowing the problem is fixed in Windows Terminal might get some folks to switch.

ForNeVeR:
I am open to ideas on how to fix it for conhost.

I've tried googling that but haven't found any working solutions on how to properly extract the active keyboard layout from conhost.

All we need is a way to get the active layout for a `conhost` process. If we can get that, then everything else is just a matter of determining whether we are attached to a conhost or not (and I believe this is possible to determine, even if a little tricky and will require another heuristic).

Just for the record, I use this version of PSReadline daily for the last several releases in VSCode, IntelliJ and Windows Terminal, and the functionality works reliably for me: it never crashed and never misdetected the terminal host.

PSReadLine/WindowsKeyboardLayoutUtil.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// </para>
+        /// </remarks>
+        public static IntPtr? GetConsoleKeyboardLayout()
+        {
+            // Define a limit not get stuck in case processed form a loop (possible in case pid reuse).
+            const int iterationLimit = 20;
+
+            var pbi = new PROCESS_BASIC_INFORMATION();
+            var process = Process.GetCurrentProcess();
+            for (var i = 0; i < iterationLimit; ++i)
```
daxian-dbw: style change:
```suggestion
            for (int i = 0; i < iterationLimit; i++)
```

PSReadLine/WindowsKeyboardLayoutUtil.cs
```diff
@@ -0,0 +1,98 @@
...
+        ///     Currently, we check up to 20 parent processes to see if their main window (as determined by the
+        ///     <see cref="Process.MainWindowHandle"/>) is visible.
+        /// </para>
+        /// <para>
+        ///     If this method returns <c>null</c>, it means it was unable to find the parent terminal process, and so
+        ///     you have to call the <see cref="GetConsoleKeyboardLayoutFallback"/>, which is known to not work properly
+        ///     in certain cases, as documented by https://github.com/PowerShell/PSReadLine/issues/1393
+        /// </para>
+        /// </remarks>
+        public static IntPtr? GetConsoleKeyboardLayout()
```
daxian-dbw: Please move the method `GetConsoleKeyboardLayout` to the end of the file `PlatformWindows.cs`. Consider renaming it to `GetKeyboardLayout` and you can call it by `PlatformWindows.GetKeyboardLayout()`.

PSReadLine/WindowsKeyboardLayoutUtil.cs
```diff
@@ -0,0 +1,98 @@
...
+        public static IntPtr GetConsoleKeyboardLayoutFallback()
+        {
+            return GetKeyboardLayout(0);
+        }
+
+        [DllImport("User32.dll", SetLastError = true)]
+        private static extern IntPtr GetKeyboardLayout(uint idThread);
+
+        [DllImport("Ntdll.dll")]
+        static extern int NtQueryInformationProcess(
```
daxian-dbw: This is already defined in `PlatformWindows`

PSReadLine/WindowsKeyboardLayoutUtil.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        [DllImport("user32.dll")]
+        [return: MarshalAs(UnmanagedType.Bool)]
+        static extern bool IsWindowVisible(IntPtr hWnd);
+
+        [DllImport("user32.dll", SetLastError = true)]
+        static extern uint GetWindowThreadProcessId(IntPtr hwnd, out IntPtr proccess);
+
+        [StructLayout(LayoutKind.Sequential)]
+        private struct PROCESS_BASIC_INFORMATION
```
daxian-dbw: Same here. This struct is already defined in `PlatformWindows`

PSReadLine/WindowsKeyboardLayoutUtil.cs
```diff
@@ -0,0 +1,98 @@
...
+            var process = Process.GetCurrentProcess();
+            for (var i = 0; i < iterationLimit; ++i)
+            {
+                var isVisible = IsWindowVisible(process.MainWindowHandle);
+                if (!isVisible)
+                {
+                    // Main process window is invisible. This is not (likely) a terminal process.
+                    var status = NtQueryInformationProcess(process.Handle, 0, ref pbi, Marshal.SizeOf(pbi), out var _);
+                    if (status != 0 || pbi.InheritedFromUniqueProcessId == IntPtr.Zero)
+                        break;
```
daxian-dbw: `PlatformWindows.GetParentPid` does this functionality. Please check if you can directly use it.

PSReadLine/WindowsKeyboardLayoutUtil.cs
```diff
@@ -0,0 +1,98 @@
...
+
+                    try
+                    {
+                        process = Process.GetProcessById(pbi.InheritedFromUniqueProcessId.ToInt32());
+                    }
+                    catch (Exception)
+                    {
+                        // No access to the process, or the process is already dead. Either way, we cannot determine its
+                        // keyboard layout.
+                        return null;
```
daxian-dbw: How about change to `break;` there, and see my comment below about the `return null` line.

PSReadLine/WindowsKeyboardLayoutUtil.cs
```diff
@@ -0,0 +1,98 @@
...
+
+                    continue;
+                }
+
+                var tid = GetWindowThreadProcessId(process.MainWindowHandle, out _);
+                if (tid == 0) return null;
+                return GetKeyboardLayout(tid);
+            }
+
+            return null;
```
daxian-dbw: Instead of this, how about just return `GetKeyboardLayout(0)` here?
```suggestion
            // comment like "fall back to the default keyboard when we failed to find the parent terminal process"
            return GetKeyboardLayout(0);
```

PSReadLine/WindowsKeyboardLayoutUtil.cs
```diff
@@ -0,0 +1,98 @@
...
+                        // keyboard layout.
+                        return null;
+                    }
+
+                    continue;
+                }
+
+                var tid = GetWindowThreadProcessId(process.MainWindowHandle, out _);
+                if (tid == 0) return null;
+                return GetKeyboardLayout(tid);
```
daxian-dbw: Style change:
```suggestion
                if (tid > 0)
                {
                    return GetKeyboardLayout(tid);
                }
```

jazzdelightsme:
I know some experts in this area; I will look into it...

PSReadLine/WindowsKeyboardLayoutUtil.cs
```diff
@@ -0,0 +1,98 @@
...
+        ///     may be higher in the process tree in case PowerShell is a child of some other console process.
+        /// </para>
+        /// <para>
+        ///     Currently, we check up to 20 parent processes to see if their main window (as determined by the
+        ///     <see cref="Process.MainWindowHandle"/>) is visible.
+        /// </para>
+        /// <para>
+        ///     If this method returns <c>null</c>, it means it was unable to find the parent terminal process, and so
+        ///     you have to call the <see cref="GetConsoleKeyboardLayoutFallback"/>, which is known to not work properly
+        ///     in certain cases, as documented by https://github.com/PowerShell/PSReadLine/issues/1393
```
daxian-dbw: How about directly doing the fallback within this method when unable to get the active keyboard layout?

daxian-dbw (commented):
Have you tried using [GetForegroundWindow](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getforegroundwindow)?

I changed the `GetConsoleKeyboardLayout` method to the following and it seems working fine for me to retrieve the current active keyboard layout, even when pwsh is opened in Conhost from cmd.exe.

```c#
public static IntPtr? GetConsoleKeyboardLayout()
{
    IntPtr handle = GetForegroundWindow();
    if (handle != IntPtr.Zero)
    {
        uint tid = GetWindowThreadProcessId(handle, out _);
        if (tid > 0)
        {
            IntPtr layout = GetKeyboardLayout(tid);
            return layout;
        }
    }

    return null;
}
```

ForNeVeR:
Thanks for your useful feedback! Will take a look tonight.

ForNeVeR:
@daxian-dbw, thanks for your fantastic feedback. I've applied all the suggestions (the most significant change being the migration to `GetForegroundWindow`), and everything I was able to check still works well!

Notably, I no longer can reproduce the issue with `conpty` on the latest stable build of Windows 11 (22H2): that one seems to always work (even in non-English keyboard layout), so I guess that's fine by me. (Other terminal apps, including Windows Terminal, still require the workaround from this PR.)

PSReadLine/PlatformWindows.cs
```diff
@@ -1015,4 +1015,34 @@ private static void TerminateStragglers()
...
+        {
+            uint tid = GetWindowThreadProcessId(foregroundWindow, out _);
+            if (tid != 0)
+            {
+                return GetKeyboardLayout(tid);
+            }
+        }
+
+        // Fall back to the default keyboard when we failed to find the parent terminal process.
+        return GetKeyboardLayout(0);
```
daxian-dbw: I just took a look at #3078 and realized that we actually circled back ...

There are 2 concerns expressed in that PR (https://github.com/PowerShell/PSReadLine/pull/3078#discussion_r761562034):
- there could be a race condition -- a window popping up by a background task causes `GetForegroundWindow()` to return the wrong handle.
- calling `GetForegroundWindow()` and `GetWindowThreadProcessId` on every keystroke that produces a `ConsoleKeyInfo key` with a control character of `key.KeyChar` is unnecessary overhead, so caching makes sense.

I think we can solve them both by combining the use of `GetForegroundWindow()` and looking up the process chain:
1. I have verified that when running in conhost, the process id returned by `GetWindowThreadProcessId(GetForegroundWindow(), out uint pid)` points to the console application that was first started with the conhost window. Namely, if you start with `cmd -> pwsh`, `pid` points to the `cmd.exe` process; if you start with `pwsh` directly in conhost, then `pid` points to the `pwsh` process.

1. When running in Windows Terminal and VSCode, the `pid` return from above call points to `WindowsTerminal` and `Code` respectively.

1. Note that, all the process `pid` points to in the above scenarios is either the current `pwsh` process or a parent process of the current `pwsh` process. You should be able to find it when walking up the process chain.
   - So, to solve the "_race condition_" concern, we should call `GetWindowThreadProcessId(GetForegroundWindow(), out uint pid)` and then walking up the process chain to verify if we can find `pid` in the chain:
      1. if we can find it, then we are good and should cache the returned thread id to use for the rest of session;
      1. if we cannot find it, then it was a race condition and we simply try again (call `GetWindowThreadProcessId(GetForegroundWindow, ..)` and verify it) until we are good.

   - To solve the "_caching thread id_" concern, all the above "call and verify" operations should be done in the one-time initializer of PSReadLine. Once we find the correct thread id, we will cache it and use it for the rest of the session.

@ForNeVeR and @lzybkr Please let me know what you think.

daxian-dbw (changes requested):

PSReadLine/PlatformWindows.cs
ForNeVeR: Ok, this suggestion makes sense. I'll try to piece it all together.

ForNeVeR:
Thanks for your useful feedback again! I believe I've addressed the suggestions. Please take another look!

PSReadLine/PlatformWindows.cs
```diff
@@ -1015,4 +1015,125 @@ private static void TerminateStragglers()
...
+    ///     that at the moment we are called, most likely a current foreground window owner will be the terminal window
+    ///     owner as well. We verify that by going up the process tree and if we find the foreground process owner among
+    ///     the current process' parent hierarchy, then we consider we found it.
+    /// </para>
+    /// <para>
+    ///     The terminal process is not always the direct parent of the current process, but may be higher in the
+    ///     process tree in case PowerShell is a child of some other console process.
+    /// </para>
+    /// <para>
+    ///     The result is cached for performance reasons. To avoid thread id reuse, we keep the thread handle as well.
```
daxian-dbw: According to the `GetThreadId` [doc](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getthreadid#remarks):
> Until a thread terminates, its thread identifier uniquely identifies it on the system.

So, it sounds the thread identifier will not be reused unless it's terminated, in which case the thread handle would be invalid too. So, why not just cache the thread id?

PSReadLine/PlatformWindows.cs
```diff
@@ -1015,4 +1015,125 @@ private static void TerminateStragglers()
...
+                // Define a limit not get stuck in case processed form a loop (possible in case pid reuse).
+                const int iterationLimit = 20;
+
+                PROCESS_BASIC_INFORMATION pbi = new();
+                Process process = Process.GetCurrentProcess();
+                int activeProcessId = process.Id;
+                for (int i = 0; i < iterationLimit; ++i)
+                {
+                    int status = NtQueryInformationProcess(process.Handle, 0, out pbi, Marshal.SizeOf(pbi), out _);
+                    activeProcessId = pbi.InheritedFromUniqueProcessId.ToInt32();
```
daxian-dbw: We already have the method `GetParentPid(Process)` (in this same file). Please directly use that helper method.

PSReadLine/PlatformWindows.cs
```diff
@@ -1015,4 +1015,125 @@ private static void TerminateStragglers()
...
+            {
+                uint tid = GetWindowThreadProcessId(foregroundWindow, out var terminalOwnerPid);
+                if (tid == 0)
+                {
+                    return null;
+                }
+
+                // We've got the process, but is it a correct process? Verify using the process parent hierarchy.
+                // Define a limit not get stuck in case processed form a loop (possible in case pid reuse).
+                const int iterationLimit = 20;
```
daxian-dbw: I have verified that when starting `pwsh` directly from `RunTask` with conhost, the owner pid is the pid of the pwsh process, so the check should start from the current process.

PSReadLine/PlatformWindows.cs
```diff
@@ -1015,4 +1015,125 @@ private static void TerminateStragglers()
...
+
+                    try
+                    {
+                        process = Process.GetProcessById(activeProcessId);
+                    }
+                    catch (Exception)
+                    {
+                        // No access to the process, or the process is already dead. Either way, we cannot determine its
+                        // keyboard layout.
+                        return null;
```
daxian-dbw: When this happens, a parent of the currently process was terminated, and thus we will never be able to verify/confirm that the foreground window is actually the one used by the current process. We can still cache the finding in this case (tid = 0), so the layout will always be retrieve by `GetKeyboardLayout(0)` essentially.

daxian-dbw: Also, I think we should retrieve and cache the `tid` in the `OneTimeInit` method. `OneTimeInit` gets called very early (at the pwsh.exe process start in the most common scenario), so it's less likely for a parent process to be terminated at that point.

PSReadLine/PlatformWindows.cs
ForNeVeR: > So, it sounds the thread identifier will not be reused unless it's terminated

This is correct.

> in which case the thread handle would be invalid too

I _assume_ this should not be true. A handle should be "valid" (i.e. available for use, or at least to check if its still alive) until it is closed.

Though I found no documentation on this, unfortunately.

ForNeVeR: I guess we can add a check for `WaitForSingleObject(threadHandle, 0)` to figure out if a thread was invalidated.

This would still be open for a race condition, but only for a very little amount of time.

I wonder if we care for this now, actually.

In the worst case, if a thread owning the terminal for our process has been terminated, and we are still trying to process some key presses, we may try to access a wrong thread (if its id was reused already).

My vote is to still try to add as much protection as reasonably possible. But if you think we could just ignore this, then well we can.

daxian-dbw: According to https://learn.microsoft.com/en-us/windows/win32/procthread/creating-windows-in-threads:
> The thread that creates the window owns the window and its associated message queue.

And according to https://learn.microsoft.com/en-us/windows/win32/procthread/terminating-a-thread:
> Terminating a thread has the following results:
>  * Any resources owned by the thread, such as windows and hooks, are freed.

So, in our case here, if the thread that created the terminal window gets terminated, then the terminal window itself is gone. Therefore, I think we can safely assume that the thread identified by the returned `tid` is alive as long as PowerShell is still alive.

Given the above, we should just cache the `tid` once we've verified it's the right one.

PSReadLine/PlatformWindows.cs
ForNeVeR: Ah, this is probably why it wasn't properly working for me in the earlier version of this PR!

PSReadLine/PlatformWindows.cs
ForNeVeR: I am still a little concerned, but alright, let's give this approach a try.

daxian-dbw:
@ForNeVeR I spent 2 days playing with code and decided that we should move back to your `MainWindowHandle` approach for two reasons:
1. `GetForegroundWindow` doesn't work for conhost either. The returned window handle is the same as the `GetConsoleWindow` call -- it's the window created for the console application (_owner is `cmd` or `pwsh` for example_), not the terminal window (_owner is conhost I guess_) that can really reflect keyboard layout changes. I cannot see the layout change by using the window handle returned from `GetForegroundWindow` when pwsh is running under conhost.

1.  For the scenarios where `GetForegroundWindow` returns the correct window handle, walking up the process tree is also needed in order to verify if it's the correct one. It turns out the `MainWindowHandle` approach works for those scenarios too, and is more reliable because we do have the risk of race condition when using `GetForegroundWindow`.

So, I changed the code back and did some polishing. Now, about the `GetTerminalOwnerThreadId` method:
- It should continue to work well in Windows Terminal (_with profile_), IntelliJ and VSCode;
- It doesn't work when PowerShell runs in conhost, or when it gets started from Start Menu (_or <kbd>win+r</kbd> "Run"_) with Windows Terminal as the default terminal application (_without profile_).
- For the scenarios where it doesn't work, I make sure it returns 0, so that the behavior is unchanged.

It's not perfect, but I'm afraid there is no perfect solution that can work in all scenarios.
I'm willing to take this improvement to unblock the Russian keyboard layout users. Can you please take a look at my changes and verify again that it still works for you in Windows Terminal (_with profile_), IntelliJ and VSCode? Please let me know if you have any concerns about my changes. Thanks!

ForNeVeR:
@daxian-dbw, thank you so much for your help.

I can confirm the following observations:
- the resulting solution works well in 
  - Windows Terminal (when explicitly started)
  - Visual Studio terminal (aka "Developer PowerShell")
  - IntelliJ
  - VSCode

It does not work for:
- conhost
- Windows Terminal when used as default terminal app
  - this one is curious, I wasn't aware of the alternate process structure when starting a process that way

Generally, for all of my use cases, it works. There's of course some room for improvement, but I believe it is significantly better than without the patch, at least.

So, I still vote for merging it, even in the current non-ideal state.

Perhaps we could introduce a better fix by introducing keyboard layout options? For me it would be even better if it just forced the English layout, not looking at the current one I use when processing the shortcuts. But that's a different topic, IMO, and even if we introduce such an option, having a heuristic that makes it work by default in more cases is good anyway.

iSazonov:
As far as I remember this problem is only with PSRL, yes? I may have missed something, but is it only for Ctrl-C? If so, is it a consequence of setting the ENABLE_PROCESSED_INPUT flag? What if we remove this flag and set the handler to handle Ctrl+C?

ForNeVeR:
> As far as I remember this problem is only with PSRL, yes?

Yep, correct.

> I may have missed something, but is it only for Ctrl-C?

No, it is not. Other shortcuts such as `Ctrl+R` also stop working.

> If so, is it a consequence of setting the ENABLE_PROCESSED_INPUT flag?

I've removed the `ENABLE_PROCESSED_INPUT` flag from the PSReadLine sources (so that it doesn't get passed to `SetConsoleInputMode`), and it doesn't seem to affect the behavior.

iSazonov:
It seems I misread ENABLE_PROCESSED_INPUT flag description. I see PSRL clear the flag and system send Ctrl+C to input buffer. So if we set the flag the system will process Ctrl+C and call out callback if we set it. Perhaps this could fix Ctrl+C at least.

daxian-dbw:
@ForNeVeR Thanks for double verifying all the scenarios. I agree this is the desired improvement even though not perfect. For scenarios where it doesn't work, the old behavior is kept unchanged. I will merge this PR.

@iSazonov It's by design for PSReadLine to clear that flag. When editing, PSReadLine wants to capture <kbd>ctrl+c</kbd> and handles it in its own way. The issue here is that `ToUnicode` treats `Ctrl+c` as dead key when the Russian keyboard layout is in use. So, another way to see it is: if we can detect dead key in a better way without depending on `ToUnicode`, then that would the perfect solution to this issue. For now, we are still depending on `ToUnicodeEx`.

daxian-dbw (approved):

PR #3810, opened 2023-09-21
Handle large history file properly by reading lines in the streaming way

### LINKED ISSUE #537: Limit size of ConsoleHost_history.txt file
Recently, accessing command history via the up-arrow has been quite slow (3-5 seconds to respond), and today, when I opened PowerShell, the console was non-responsive and after about 3 mintues would throw the error below:

```
Windows PowerShell
Copyright (C) 2016 Microsoft Corporation. All rights reserved.

PS C:\Users\me\Documents>
Oops, something went wrong.  Please report this bug with the details below.
Report on GitHub: https://github.com/lzybkr/PSReadLine/issues/new
Last 0 Keys:


Exception:
System.OutOfMemoryException: Array dimensions exceeded supported range.
   at System.Collections.Generic.List`1.set_Capacity(Int32 value)
   at System.Collections.Generic.List`1.EnsureCapacity(Int32 min)
   at System.Collections.Generic.List`1.Add(T item)
   at System.IO.File.InternalReadAllLines(String path, Encoding encoding)
   at Microsoft.PowerShell.PSConsoleReadLine.<ReadHistoryFile>b__67_0()
   at Microsoft.PowerShell.PSConsoleReadLine.WithHistoryFileMutexDo(Int32 timeout, Action action)
   at Microsoft.PowerShell.PSConsoleReadLine.DelayedOneTimeInitialize()
   at Microsoft.PowerShell.PSConsoleReadLine.Initialize(Runspace runspace, EngineIntrinsics engineIntrinsics)
```

With little digging I found that the console history file found here:

```
C:\Users\me\AppData\Roaming\Microsoft\Windows\PowerShell\PSReadline\ConsoleHost_history.txt
```
had grown to 13MB and contained about 3 months of commands, and deleting the file fixed both issues.

How can the size of this file be moderated?

Environment data
----------------
PS version: 5.1.14393.1198
PSReadline version: 1.2
os: 10.0.14393.0 (rs1_release.160715-1616)
PS file version: 10.0.14393.0 (rs1_release.160715-1616)

### LINKED ISSUE #1360: In VS Code getting this System.OutOfMemoryException: Array dimensions exceeded supported range.
Oops, something went wrong.  Please report this bug with the details below.
Report on GitHub: https://github.com/lzybkr/PSReadLine/issues/new
Last 0 Keys:

Exception:
System.OutOfMemoryException: Array dimensions exceeded supported range.
   at System.Collections.Generic.List`1.set_Capacity(Int32 value)
   at System.Collections.Generic.List`1.EnsureCapacity(Int32 min)
   at System.Collections.Generic.List`1.Add(T item)
   at System.IO.File.InternalReadAllLines(String path, Encoding encoding)
   at Microsoft.PowerShell.PSConsoleReadLine.<ReadHistoryFile>b__67_0()
   at Microsoft.PowerShell.PSConsoleReadLine.WithHistoryFileMutexDo(Int32 timeout, Action action)
   at Microsoft.PowerShell.PSConsoleReadLine.DelayedOneTimeInitialize()
   at Microsoft.PowerShell.PSConsoleReadLine.Initialize(Runspace runspace, EngineIntrinsics engineIntrinsics)

### LINKED ISSUE #3771: PSReadLine cannot handle super large history file properly
### Prerequisites

### Exception report

```console
System out of memory  and it consume lots of memory when i see in task manager.
```

### Screenshot

image

### Environment data

```console
PS Version: 5.1.14393.5582
PS HostName: ConsoleHost
PSReadLine Version: 1.2
PSReadLine EditMode: Windows
OS: 10.0.14393.0 (rs1_release.160715-1616)
BufferWidth: 139
BufferHeight: 3000
```

### Steps to reproduce

Start powershell in visual code or windows it always crashed 

### Expected behavior

always crashed

### Actual behavior

Start powershell in visual code or windows it always crashed

daxian-dbw:
# PR Summary

Fix #3771, Fix #1360, Fix #537

Handle large history file properly by reading lines in the streaming way.

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/PowerShell/PSReadLine/pull/3810)

### DIFF
--- a/PSReadLine/History.cs
+++ b/PSReadLine/History.cs
@@ -497,7 +497,7 @@ namespace Microsoft.PowerShell
                     string line;
                     while ((line = sr.ReadLine()) is not null)
                     {
-                        if (!line.EndsWith('`'))
+                        if (!line.EndsWith("`", StringComparison.Ordinal))
                         {
                             // A complete history record is guaranteed to start from the next line.
                             break;
--- a/PSReadLine/History.cs
+++ b/PSReadLine/History.cs
@@ -470,14 +470,14 @@ namespace Microsoft.PowerShell
                 const long offset_05mb = 524288;
 
                 // 1mb content contains more than 34,000 history lines for a typical usage, which should be
-                // more than enough to cover 25,000 history records (a history record could be a multi-line
+                // more than enough to cover 20,000 history records (a history record could be a multi-line
                 // command). Similarly, 0.5mb content should be enough to cover 10,000 history records.
                 // We optimize the file reading when the history count falls in those ranges. If the history
                 // count is even larger, which should be very rare, we just read all lines.
                 long offset = historyCount switch
                 {
                     <= 10000 => offset_05mb,
-                    <= 25000 => offset_1mb,
+                    <= 20000 => offset_1mb,
                     _ => 0,
                 };
 
@@ -486,22 +486,22 @@ namespace Microsoft.PowerShell
 
                 if (offset > 0 && fs.Length > offset)
                 {
-                    // When the file size is larger than 1mb, we only read the last 1mb content from the end.
-                    int? b1 = null, b2 = null;
+                    // When the file size is larger than the offset, we only read that amount of content from the end.
                     fs.Seek(-offset, SeekOrigin.End);
 
-                    // After seeking, the current position may point at the middle of a history record, or even
-                    // a byte within a unicode char. So, we need to find the start of the next history record.
-                    while ((b2 = fs.ReadByte()) is not -1)
+                    // After seeking, the current position may point at the middle of a history record, or even at a
+                    // byte within a UTF-8 character (history file is saved with UTF-8 encoding). So, let's ignore the
+                    // first line read from that position.
+                    sr.ReadLine();
+
+                    string line;
+                    while ((line = sr.ReadLine()) is not null)
                     {
-                        // Read bytes until we find the first newline ('\n' == 0xA) that is not right after a backtick ('`' == 0x60).
-                        // It means a separate full history record will start from the next byte.
-                        if (b2 is 0xA && b1.HasValue && b1 is not 0x60)
+                        if (!line.EndsWith('`'))
                         {
+                            // A complete history record is guaranteed to start from the next line.
                             break;
                         }
-
-                        b1 = b2;
                     }
                 }
 
--- a/PSReadLine/History.cs
+++ b/PSReadLine/History.cs
@@ -457,11 +457,60 @@ namespace Microsoft.PowerShell
             {
                 WithHistoryFileMutexDo(1000, () =>
                 {
-                    var historyLines = File.ReadAllLines(Options.HistorySavePath);
+                    var historyLines = ReadHistoryLinesImpl(Options.HistorySavePath, Options.MaximumHistoryCount);
                     UpdateHistoryFromFile(historyLines, fromDifferentSession: false, fromInitialRead: true);
                     var fileInfo = new FileInfo(Options.HistorySavePath);
                     _historyFileLastSavedSize = fileInfo.Length;
                 });
+            }
+
+            static IEnumerable<string> ReadHistoryLinesImpl(string path, int historyCount)
+            {
+                const long offset_1mb = 1048576;
+                const long offset_05mb = 524288;
+
+                // 1mb content contains more than 34,000 history lines for a typical usage, which should be
+                // more than enough to cover 25,000 history records (a history record could be a multi-line
+                // command). Similarly, 0.5mb content should be enough to cover 10,000 history records.
+                // We optimize the file reading when the history count falls in those ranges. If the history
+                // count is even larger, which should be very rare, we just read all lines.
+                long offset = historyCount switch
+                {
+                    <= 10000 => offset_05mb,
+                    <= 25000 => offset_1mb,
+                    _ => 0,
+                };
+
+                using var fs = new FileStream(path, FileMode.Open);
+                using var sr = new StreamReader(fs);
+
+                if (offset > 0 && fs.Length > offset)
+                {
+                    // When the file size is larger than 1mb, we only read the last 1mb content from the end.
+                    int? b1 = null, b2 = null;
+                    fs.Seek(-offset, SeekOrigin.End);
+
+                    // After seeking, the current position may point at the middle of a history record, or even
+                    // a byte within a unicode char. So, we need to find the start of the next history record.
+                    while ((b2 = fs.ReadByte()) is not -1)
+                    {
+                        // Read bytes until we find the first newline ('\n' == 0xA) that is not right after a backtick ('`' == 0x60).
+                        // It means a separate full history record will start from the next byte.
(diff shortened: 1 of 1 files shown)


PSReadLine/History.cs
```diff
@@ -457,12 +457,61 @@ private void ReadHistoryFile()
...
+                    int? b1 = null, b2 = null;
+                    fs.Seek(-offset, SeekOrigin.End);
+
+                    // After seeking, the current position may point at the middle of a history record, or even
+                    // a byte within a unicode char. So, we need to find the start of the next history record.
+                    while ((b2 = fs.ReadByte()) is not -1)
+                    {
+                        // Read bytes until we find the first newline ('\n' == 0xA) that is not right after a backtick ('`' == 0x60).
+                        // It means a separate full history record will start from the next byte.
+                        if (b2 is 0xA && b1.HasValue && b1 is not 0x60)
```
iSazonov: Eh, it is not good practice to read unicode chars by bytes. Thus I could save the file in Utf16.

Since we want address an edge case - in most cases this file is in kilobytes or megabytes but not gigabytes - 
I suggest reading line by line, but if the file is really larger than 1.5GB ignore the first lines until the remainder of the file is around 500Mb.

daxian-dbw: > it is not good practice to read unicode chars by bytes. Thus I could save the file in Utf16.

That's a good point. I will rework on this.

PSReadLine/History.cs
```diff
@@ -457,12 +457,61 @@ private void ReadHistoryFile()
...
+
+                if (offset > 0 && fs.Length > offset)
+                {
+                    // When the file size is larger than the offset, we only read that amount of content from the end.
+                    fs.Seek(-offset, SeekOrigin.End);
+
+                    // After seeking, the current position may point at the middle of a history record, or even at a
+                    // byte within a UTF-8 character (history file is saved with UTF-8 encoding). So, let's ignore the
+                    // first line read from that position.
+                    sr.ReadLine();
```
daxian-dbw, line 495: @iSazonov Regarding you comment "I could save the file in Utf16", PSReadLine always saves the history file in UTF-8 encoding. Both [File.CreateText](https://learn.microsoft.com/dotnet/api/system.io.file.createtext?view=netframework-4.8) and [File.AppendText](https://learn.microsoft.com/dotnet/api/system.io.file.appendtext?view=netframework-4.8) writes UTF-8 encoding text to the file, so if a user saves the file in Unicode, then it may be broken to read later after PSReadLine writes something to the file. So, I will assume it's UTF-8 encoding.

Can you please review again? Thanks!

PSReadLine/History.cs
daxian-dbw: > I suggest reading line by line, but if the file is really larger than 1.5GB ignore the first lines until the remainder of the file is around 500Mb.

I'd still like to only read partial content from the file, for 2 reasons:
1. reduce GC and unneeded operation as much as possible, also help get the prompt to user faster when the history file is large.
2. there is no good way to find out the reminder of the file when reading line by line, because the position of the underlying file stream is not going to point at the next byte to be read due to the block reading and caching of the stream reader.

PSReadLine/History.cs
iSazonov: If you see no problem that the history file may be corrupted in very rare cases this code looks good.

---

For reflection only, I have the following questions and doubts.

1. The user can actually save the file in a different encoding.
2. Only .Net Core at some point started using Utf8 by default. Windows PowerShell can write in Utf16 by default.
3. Although StreamReader detects the encoding automatically, the current code seems to ignore this as it skips the beginning of the file.
4. Why do we store the beginning of a large file when now we don't use it anyway and the user doesn't even know about it? Probably we can find a way to trim it, although it's not easy to think of.

PSReadLine/History.cs
```diff
@@ -457,12 +457,61 @@ private void ReadHistoryFile()
...
+
+                    string line;
+                    while ((line = sr.ReadLine()) is not null)
+                    {
+                        if (!line.EndsWith("`", StringComparison.Ordinal))
```
iSazonov, line 500:
```suggestion
                        if (!line.EndsWith('`', StringComparison.Ordinal))
```

daxian-dbw: The overload `String.EndsWith(char)` is only available in .NET, not in .NET Framework.

PSReadLine/History.cs
daxian-dbw: > The user can actually save the file in a different encoding.

If the user manually changes encoding, to like UTF-16 LE, the history file will be corrupted once PSReadLine writes updates to the file, because PSReadLine always writes text in UTF-8. I verified this using both Windows PowerShell and PowerShell 7+.

See the code below (Both [File.CreateText](https://learn.microsoft.com/dotnet/api/system.io.file.createtext?view=netframework-4.8) and [File.AppendText](https://learn.microsoft.com/dotnet/api/system.io.file.appendtext?view=netframework-4.8) writes UTF-8 encoding text to the file, on both .NET and .NET Framework):

https://github.com/PowerShell/PSReadLine/blob/d045b508f41e2c3fe9763768a5adfc58a5439907/PSReadLine/History.cs#L366-L379

> Only .Net Core at some point started using Utf8 by default. Windows PowerShell can write in Utf16 by default.

Again, [File.CreateText](https://learn.microsoft.com/dotnet/api/system.io.file.createtext?view=netframework-4.8) and [File.AppendText](https://learn.microsoft.com/dotnet/api/system.io.file.appendtext?view=netframework-4.8) writes UTF-8 encoding text to the file, on both .NET and .NET Framework

> Although StreamReader detects the encoding automatically, the current code seems to ignore this as it skips the beginning of the file.

I tried it out, and it turns out that `StreamReader` detects encoding when creating the instance.

> Why do we store the beginning of a large file when now we don't use it anyway and the user doesn't even know about it? Probably we can find a way to trim it, although it's not easy to think of.

I think you missed something in the code. If the user decides to set max history count to more than 20,000, we will read all content from the file. So, the history is still accessible to the user as long as they want.

daxian-dbw:
@iSazonov Can you please approve the PR if all your concerns are resolved? I need an approval to merge the PR.

iSazonov (approved):

iSazonov:
> @iSazonov Can you please approve the PR if all your concerns are resolved? I need an approval to merge the PR.

Approved with one note but I haven't full permissions :-)

PSReadLine/History.cs
```diff
@@ -457,12 +457,61 @@ private void ReadHistoryFile()
...
+                        {
+                            // A complete history record is guaranteed to start from the next line.
+                            break;
```
iSazonov, line 503: Should we return `line` here if there wasn't  an apostrophe or we agree to lose one line for simplicity.?

daxian-dbw: This line could be the last line of a multi-line history record, so we cannot use it.

PR #3791, opened 2023-08-23
[Vi-Mode] Supports di' and di" commands (quoted text objects)

springcomp:
# PR Summary

Fixes #3790.

## PR Checklist

- [x] PR has a meaningful title
    - Use the present tense and imperative mood when describing your changes
- [x] Summarized changes
- [x] Make sure you've added one or more new tests
- [x] Make sure you've tested these changes in terminals that PowerShell is commonly used in (i.e. conhost.exe, Windows Terminal, Visual Studio Code Integrated Terminal, etc.)
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - [x] Documentation needed at [PowerShell-Docs](https://github.com/MicrosoftDocs/PowerShell-Docs)
        - [ ] Doc Issue filed: 

As #2059 has been readily accepted, we have not had the time to discuss the documentation.
This PR, like #2059, wire keybindings to private methods. Please, let me know how to best document this behaviour.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/PowerShell/PSReadLine/pull/3791)

PR #2059, opened 2020-12-18
[vi-mode] Supports 'diw' text-object command.

springcomp:
# PR Summary

Fixes #2057.

This PR is a first attempt at supporting text-objects in vi-mode. It includes a reasonably complete implementation of the <kbd>d</kbd><kbd>i</kbd><kbd>w</kbd> command.

Although I have worked hard to make this PR _production ready_, it must still be considered WIP. I would really appreciate feedback about the implementation.

If this is deemed interesting enough, it could be the first step towards a much complete set of pull requests as alluded to in #584.

## PR Checklist

- [x] PR has a meaningful title
    - Use the present tense and imperative mood when describing your changes
- [x] Summarized changes
- [x] Make sure you've added one or more new tests
- [x] Make sure you've tested these changes in terminals that PowerShell is commonly used in (i.e. conhost.exe, Windows Terminal, Visual Studio Code Integrated Terminal, etc.)
- **User-facing changes**
    - [x] Documentation needed at [PowerShell-Docs](https://github.com/MicrosoftDocs/PowerShell-Docs)
        - [ ] Doc Issue filed: _pending feedback_

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/PowerShell/PSReadLine/pull/2059)

Issue #3735, opened 2023-07-07
ViMode: Yanking until given character


Konstantin-Glukhov:
### Prerequisites

- [X] Write a descriptive title.

### Description of the new feature/enhancement

Yanking until given character for -ViMode Command

### Proposed technical implementation details (optional)

<y,t>            ViYankToBeforeChar                    Yanks until given character forward.
<y,T>            ViYankToBeforeCharBackward    Yanks until given character backward.
<y,f>            ViYankToChar                              Yanks inclusively until given character forward.
<y,F>            ViYankToCharBackward              Yanks inclusively until given character backward.

matteocoder:
If one wanted to implement this functionality using only methods from PSReadline, how would you programmatically access the register which is used in command mode to store the yanked text? I mean in the same way one can access the kill ring using only the psconsolereadline class.

Konstantin-Glukhov:
I am not sure. How are y0 and y$ implemented? Shouldn't be the same? First find a position of the searched character and then yank from current position to the found position?

matteocoder:
@Konstantin-Glukhov Thanks for the reply. My problem in particular is that I cannot find, looking at the documentation, a psreadline method like
```ps1
[Microsoft.PowerShell.PSConsoleReadLine]::ViPlaceThisStringInTheViClipboard(<string>)
```
Implementing the motions seems already possible with the available methods, but it seems it is not possible to directly access the place where text is yanked. I'm happy to be proven wrong, though.

Konstantin-Glukhov:
Did you find how <y,$> is implemented?

matteocoder:
Not yet. I know nothing about C#, so it's kind of hard for me.

EDIT: the text is pasted to a private `_viRegister`, so it seems like it's not accessible from outside.

PR #3853, opened 2023-10-31
Refactoring Key Chords Handling

springcomp:
# PR Summary

This PR refactors the handling of key chords and consolidates the behaviour from Emacs and VI modes into a single consistent function.

This PR addresses freedback made on #3791 following #2059 which has been merged a bit too optimistically.

# PR Content

In order to achieve a consistent behaviour, the _Chord Dispatch Table_ structures are refactored using a `KeyHandlerOrChordDispatchTable` class that can hold either a regular `KeyHandler` or a nested `ChordDispatchTable`. This means that chord dispatch tables are organized in a tree structure and looking up a key handler is a simple matter of walking down the tree structure.

This allows to greatly simplify the handling of VI-mode three key-chord commands, such as <kbd>d</kbd>, <kbd>g</kbd>, <kbd>g</kbd> (`<d,g,g>` _delete relative lines_) or <kbd>d</kbd>, <kbd>i</kbd>, <kbd>w</kbd> (`<d,i,w>` _delete inner word_)

This allows to remove all the references to a _secondary_ dispatch table when handling key chords in favor of a more consistent algorithm that walks through the chord dispatch table tree structure and identifies whether the next node is a leaf (_i.e_ a `KeyHandler`) or another nested node (_i.e_ an inner-level chord dispatch table).

**Note** in #3791 was discussed the fact that we may want to limit key chords to a maximum of three keys. This PR does not enforce any limit. Maybe such a limit could be hardcoded in the `PSConsoleReadLine.SetKeyHandler()` CmdLet implementation.

This PR consolidates handling of key chords in the `ProcessOneKey` function – which could be renamed `ProcessKeyChord`. Most of the current implementations were using recursive code – albeit, indirectly by calling either the `Chord` or the `ViChord` method. In contrast, this PR _almost_ eliminates recursive code in favor of walking the tree structure. However, handling of VI-mode digit-arguments is a bit complex and I was not able to solve this without resorting to recursive code. This make the `ProcessOneKey` function a bit hard to understand. I have tried adding as much comments as I could think of.

The unit-tests sometimes randomly fail on my machine, but that was already the case with the current code in the `master` branch before this PR. Sometimes, unit-tests fail as part of running the whole test suite. When running one failing test again it then succeeds. Finally, most tests around history fail on my machine but – again – that is already the case with the current code in the `master` branch 🤔. This has mostly been addressed in [a separate pull request](/PowerShell/PSReadLine/pull/4686).

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
        - [ ] Doc Issue filed: 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/PowerShell/PSReadLine/pull/3853)

### DIFF
--- a/PSReadLine/TextObjects.Vi.cs
+++ b/PSReadLine/TextObjects.Vi.cs
@@ -1,98 +1,10 @@
 using System;
-using System.Collections.Generic;
 
 namespace Microsoft.PowerShell
 {
     public partial class PSConsoleReadLine
     {
-        internal enum TextObjectOperation
-        {
-            None,
-            Change,
-            Delete,
-        }
-
-        internal enum TextObjectSpan
-        {
-            None,
-            Around,
-            Inner,
-        }
-
-        private TextObjectOperation _textObjectOperation = TextObjectOperation.None;
-        private TextObjectSpan _textObjectSpan = TextObjectSpan.None;
-
-        private readonly Dictionary<TextObjectOperation, Dictionary<TextObjectSpan, KeyHandler>> _textObjectHandlers = new()
-        {
-            [TextObjectOperation.Delete] = new() { [TextObjectSpan.Inner] = MakeKeyHandler(ViDeleteInnerWord, "ViDeleteInnerWord") },
-        };
-
-        private void ViChordDeleteTextObject(ConsoleKeyInfo? key = null, object arg = null)
-        {
-            _textObjectOperation = TextObjectOperation.Delete;
-            ViChordTextObject(key, arg);
-        }
-
-        private void ViChordTextObject(ConsoleKeyInfo? key = null, object arg = null)
-        {
-            if (!key.HasValue)
-            {
-                ResetTextObjectState();
-                throw new ArgumentNullException(nameof(key));
-            }
-
-            _textObjectSpan = GetRequestedTextObjectSpan(key.Value);
-
-            // Handle text object
-            var textObjectKey = ReadKey();
-            if (_viChordTextObjectsTable.TryGetValue(textObjectKey, out _))
-            {
-                _singleton.ProcessOneKey(textObjectKey, _viChordTextObjectsTable, ignoreIfNoAction: true, arg: arg);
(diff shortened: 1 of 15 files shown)


test/OptionsTest.cs
```diff
@@ -85,15 +85,17 @@ public void GetKeyHandlers()
...
 
             TestSetup(KeyMode.Emacs);
             
             handlers = PSConsoleReadLine.GetKeyHandlers(Chord: new string[] { "ctrl+x" });
-            Assert.NotEmpty(handlers);
-            foreach (var handler in handlers)
-            {
-                Assert.Equal("Ctrl+x", handler.Key);
-            }
+            Assert.Empty(handlers);
```
springcomp, line 98: As far as I can see, this is the only behaviour change. Previously to this PR, a partial chord would return `ChordFirstKey` as its handler.

This PR eliminates this indirection and does not return any handler for partial chords. Only handlers at the leaf are returned.

PSReadLine/TextObjects.Vi.cs
```diff
@@ -1,98 +1,13 @@
...
-
-        private static void ResetTextObjectState()
-        {
-            _singleton._textObjectOperation = TextObjectOperation.None;
-            _singleton._textObjectSpan = TextObjectSpan.None;
-        }
-
-        private static void ViDeleteInnerWord(ConsoleKeyInfo? key = null, object arg = null)
+        /// <summary>
+        /// Delete the content inside and including the current word
```
springcomp, line 8: The `ViDeleteInnerWord` method is now wired directly form the chord dispatch table and has been made public so that it can be remapped by the user.

All the complex wiring to ultimately call this function on `<d,i,w>` has been eliminated away, in favor of simply wiring this method as a handler in the appropriate location in the chord dispatch table.

springcomp:
This pull request will make #3735 easy to address.

PR #3897, opened 2023-12-11
Stop trying to de-duplicate completion results

### LINKED ISSUE #3896: MenuComplete filtering duplicates when it shouldn't be
### Prerequisites

### Exception report

```console
N/A
```

### Screenshot

Test files:
explorer

<BR>How the test function behaves (correctly) in ISE when using CTRL+SPACE:
ISE

<BR>How it behaves (correctly) in VSCode:
VSCode

<BR></BR>How it behaves in the console (with PSReadLine).  The duplicates have been eliminated when they shouldn't be; although the ListItemText is identical for the duplicate CompletionMatches, the objects represented by each CompletionResult are unique.
pwsh Console

### Environment data

```console
PS Version: 7.4.0
PS HostName: ConsoleHost
PSReadLine Version: 2.3.4
PSReadLine EditMode: Windows
OS: 10.0.19041.1 (WinBuild.160101.0800)
BufferWidth: 210
BufferHeight: 9000
```

### Steps to reproduce

This is the test function I used.

Auto-complete (using either `TAB` or `CTRL+SPACE`) should only show files (and not folders) that start with "file" + a number between 2 and 9 (regardless of what comes after the number).

```
<# Get-ChildItem C:\test results:

file3.txt  <--- This is a directory
file1 - Copy.txt
file1.txt
file2 - Copy.txt
file2.txt
file4 - Copy.txt
file4.txt
[file].txt
#>

$gciParams = @{
    Path = Join-Path -Path 'C:\Test' -ChildPath '*'
    Filter = 'file*.*'
    File = $true
    Force = $true
}
<# (gci @gciParams).Name results:

file1 - Copy.txt
file1.txt
file2 - Copy.txt
file2.txt
file4 - Copy.txt
file4.txt
#>

$pattern = '.*[2-9].*'
$gciCommand = {Get-ChildItem @gciParams | Where-Object {$_.Name -match $pattern}}
<# (Invoke-Command $gciCommand).Name results:

file2 - Copy.txt
file2.txt
file4 - Copy.txt
file4.txt
#>

function Test-CompletionResult
{
    [CmdletBinding()]
    param (
        [Parameter(Mandatory, Position = 0, ParameterSetName = 'Num', ValueFromPipeline, ValueFromPipelineByPropertyName)]
        [ArgumentCompleter({
            param ($Command, $Parameter, $WordToComplete, $CommandAst, $FakeBoundParams)
            $objList = @()
            (Invoke-Command -ScriptBlock $gciCommand).foreach({
                $filePath = $_.FullName
                $fileNum  = ($_.Name | Select-String -Pattern '\d').Matches.Value
                $objList += [pscustomobject] @{
                    Path = $filePath
                    Num  = $fileNum
                }
            })
            ($objList | Where-Object {$_.Num -like "*$WordToComplete*"}).foreach({
                [System.Management.Automation.CompletionResult]::new($_.Num, $_.Num, 'ProviderItem', $_.Path)
            })
        })]        
        [int]$FileNum
    )
    process {Write-Output $FileNum}
}
```

### Expected behavior

- MenuComplete items shouldn't be filtered out (or recognized as "duplicates") when the ListItemText is the same yet the CompletionResults are different for each MenuComplete item (as the CompletionResult tooltip(s) in my example illustrate)
- Expected behavior should mirror how ISE and VSCode present CompletionResult items

### Actual behavior

The "duplicate" MenuComplete items are filtered out.

daxian-dbw:
# PR Summary

Fix #3896
Given that both ISE and VSCode editor completion doesn't de-duplicate completion results, we probably should stop trying to de-duplicate completion results in PSReadLine to be consistent.

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/PowerShell/PSReadLine/pull/3897)

### DIFF
--- a/PSReadLine/Completion.cs
+++ b/PSReadLine/Completion.cs
@@ -298,31 +298,6 @@ namespace Microsoft.PowerShell
                     var length = _tabCompletions.ReplacementLength;
                     if (start < 0 || start > _singleton._buffer.Length) return null;
                     if (length < 0 || length > (_singleton._buffer.Length - start)) return null;
-
-                    if (_tabCompletions.CompletionMatches.Count > 1)
-                    {
-                        // Filter out apparent duplicates -- the 'ListItemText' is exactly the same.
-                        var hashSet = new HashSet<string>();
-                        var matches = _tabCompletions.CompletionMatches;
-                        List<int> indices = null;
-
-                        for (int i = 0; i < matches.Count; i++)
-                        {
-                            if (!hashSet.Add(matches[i].ListItemText))
-                            {
-                                indices ??= new List<int>();
-                                indices.Add(i);
-                            }
-                        }
-
-                        if (indices is not null)
-                        {
-                            for (int i = indices.Count - 1; i >= 0; i--)
-                            {
-                                matches.RemoveAt(indices[i]);
-                            }
-                        }
-                    }
                 }
                 catch (Exception)
                 {


blaisemGH:
I noticed on the latest master branch that I am ending up with a lot of duplicates in a particular scenario. Specifically, when I try to tab complete a function from a module that has not been imported yet.

An easy reproducible example is to type `Get-PSre` and hit tab. Assuming you have not imported PowerShellGet, you should end up with multiple entries for `Get-PSRepository`. In a couple of my modules that export more functions, I can end up with dozens of entries as each exported function appears 4-6 times.

I am not sure if it's related to this particular commit or not, but it seems suspicious. On 2.3.5 I don't have this bug.

Should I raise a separate issue on this topic?

daxian-dbw:
@blaisemGH Can you check if those duplicate results are also present when using intellisense in VSCode editor?

blaisemGH:
> @blaisemGH Can you check if those duplicate results are also present when using intellisense in VSCode editor?

Yes, they are also there:

image

blaisemGH:
> @blaisemGH Can you check if those duplicate results are also present when using intellisense in VSCode editor?

@daxian-dbw Yes, the duplicates are also there in the latest version:

image

PowerShellGet has not been imported, so its exported functions seem to be duplicated.

The extension terminal is running PS7.4.2

blaisemGH:
@daxian-dbw 

Actually I just tested it again, and I seem to have made a mistake with my earlier statement that it began with 2.4.0. The issue existed already in PSReadLine 2.3.5:

image

However, it seems to have not existed in PSReadLine 2.3.4:

image

~~But the completion.cs file is the same across these versions, so I guess it's not related to this commit.~~ Nevermind, it is this. For some reason the module 2.3.5 download from PSGallery is not the same as the Tag v2.3.5 in github—the latter shows that completion.cs still has the dedup logic, even though it doesn't (see next comment).

blaisemGH:
@daxian-dbw Ok I just cloned the github tag v2.3.5, and tried building it with and without the changes in this commit.  The duplicates are removed with the deduplication logic; with this commit's changes (removing the deduplication logic), the duplicates are present.

So I guess the code removed in this commit was covering up the duplicates.

MartinGC94:
I have a fix for this in the tab completion code here: https://github.com/PowerShell/PowerShell/pull/21113 though that obviously won't help older PS versions.  
IMO the deduplication in PSReadLine was annoying because it was based on the list item text so if I for example tried to complete `[File<Ctrl+Space>` I would end up with just `[System.Net.WebRequestMethods+File]` when I was actually looking for `[System.IO.File]` so if you must bring it back then please use the actual completion text this time.

blaisemGH:
Oh nice. I was digging through the tabexpansion2 source too, but glad to know there's already a PR on it. Also on my machine, it does seem to be caused by multiple available module versions and each being prepended with the module name.

I didn't know the deduplication code in PSReadLine was causing issues, but at any rate, even without that I can understand why this is the wrong place to resolve an upstream error.

Thanks for the feedback here.

daxian-dbw:
Yeah, the deduplication was problematic and the reason to remove it is to get consistency as in VSCode editor (not terminal) and ISE. @MartinGC94 Thanks for the PR! I will make sure review it (and maybe merge it :)) on Monday.

MartinGC94:
@daxian-dbw Any idea on when PSReadLine 2.4.2 will be released properly (and included in PS7)? Today I tried completing `Get-MpComputerStatus` and got `ConfigDefender\Get-MpComputerStatus` and thought I had made a mistake in my previous PR but it turned out to be this deduplication logic that had hidden the other result from me: `Defender\Get-MpComputerStatus`.
Side note: Weird choice by the defender team to create 2 modules with the same commands that seem to function identical. Why couldn't they just have updated the old one?

PR #4462, opened 2025-02-07
Fix handling of surrogate pairs in length calculation

### LINKED ISSUE #3857: Using an emoji as the prompt text parse error replacement causes glitchy text to remain in the text buffer
### Prerequisites

### Exception report

```console
N/A
```

### Screenshot

glitchy buffer
glitchy buffer2

### Environment data

```console
PS Version: 7.4.0-rc.1
PS HostName: ConsoleHost (Windows Terminal)
PSReadLine Version: 2.3.4
PSReadLine EditMode: Windows
OS: 10.0.22621.1 (WinBuild.160101.0800)
BufferWidth: 78
BufferHeight: 43
```

### Steps to reproduce

Prerequisite: use the following `profile.ps1`
```ps1
if (([console]::outputEncoding).EncodingName -ne 'Unicode (UTF-8)') {
    #[Console]::InputEncoding = [Console]::OutputEncoding = $OutputEncoding = [System.Text.Utf8Encoding]::new()
}
function prompt {
    "👉 "
}
Set-PSReadLineOption -PromptText '👉 ','❌ '
Set-PSReadLineOption -predictionsource none
```

1. Write text containing a character that will cause PSReadline to register a parsing error, such as `hello world(`
2. Press `Esc` to delete the whole line

### Expected behavior

The text buffer should be completely cleared.

### Actual behavior

The characters which caused the parse error remain stuck on the buffer, and cannot be accessed by moving the cursor to their position. Additionally, when the parse error happens, the error prompt shifts slightly to the right.

totkeks:
# PR Summary

Adds new conditions to handle surrogate pairs when calculating the buffer length of a string.

Fixes https://github.com/PowerShell/PSReadLine/issues/3857.

## PR Checklist

- [X] PR has a meaningful title
- [X] Summarized changes
- [X] Make sure you've tested these changes in terminals that PowerShell is commonly used in (i.e. conhost.exe, Windows Terminal, Visual Studio Code Integrated Terminal, etc.)
- **User-facing changes**
    - [X] Not Applicable
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/PowerShell/PSReadLine/pull/4462)


totkeks:

springcomp:
I think this pull request needs a unit-test to demonstrate the requirements, but a couple unit tests would be nice.

andyleejordan:
Oh I'm so sorry, we should've gotten to this sooner. Looks like the correct fix the identified bug to me.

andyleejordan:
Hey @totkeks I merged the main branch in (it's sadly still called "master") and added some Claude generated unit tests which looked correct to me. Can you take a look and confirm the assertions are covering the bugfix?

PSReadLine/Render.Helper.cs
```diff
@@ -57,7 +57,13 @@ internal static int LengthInBufferCells(string str, int start, int end)
             for (var i = start; i < end; i++)
             {
                 var c = str[i];
-                if (c == 0x1b && (i+1) < end && str[i+1] == '[')
+                if (char.IsHighSurrogate(c) && (i + 1) < end && char.IsLowSurrogate(str[i + 1]))
```
andyleejordan: @totkeks Would [IsSurrogatePair(c, c+1)](https://learn.microsoft.com/en-us/dotnet/api/system.char.issurrogatepair?view=net-10.0#system-char-issurrogatepair(system-char-system-char)) work better here rather than checking `IsHigh(c) && IsLow(c+1)`?

PSReadLine/Render.Helper.cs
```diff
@@ -77,7 +83,13 @@ internal static int LengthInBufferCells(StringBuilder sb, int start, int end)
             for (var i = start; i < end; i++)
             {
                 var c = sb[i];
-                if (c == 0x1b && (i + 1) < end && sb[i + 1] == '[')
+                if (char.IsHighSurrogate(c) && (i + 1) < end && char.IsLowSurrogate(sb[i + 1]))
```
andyleejordan: Ditto.

PSReadLine/Render.Helper.cs
andyleejordan: Yup, slightly more readable. I changed it.

andyleejordan (approved):
Looks good to me. @daxian-dbw?

PSReadLine/Render.Helper.cs
```diff
@@ -57,7 +57,13 @@ internal static int LengthInBufferCells(string str, int start, int end)
             for (var i = start; i < end; i++)
             {
                 var c = str[i];
-                if (c == 0x1b && (i+1) < end && str[i+1] == '[')
+                if ((i + 1) < end && char.IsSurrogatePair(c, str[i + 1]))
+                {
+                    sum++;
+                    i++; // Skip the low surrogate
+                    continue;
+                }
```
daxian-dbw, line 65: This doesn't look right. It effectively counts every surrogate pair as 1 cell in terminal display, but that's not the case for almost all surrogate pairs. For example, 😀 and 👉 are both 2-cell width.

I intended to rewrite the length calculation using the lib [Wcwidth](https://www.nuget.org/packages/Wcwidth/4.0.1) and rewrite the cursor movement code using `System.Globalization.StringInfo`, but just didn't get the time to invest into this area ...

daxian-dbw: The root cause lies in `LengthInBufferCells`, but for the reported issue, it's because `LengthInBufferCells` returns a smaller number than the actual buffer cells taken by the '❌' emoji.

The '❌' emoji is the Unicode point `U+274C`. It's a single 16-byte Unicode character, not a surrogate pair. However, it takes 2 buffer cells (1 char -> 2 cells in this case). So, for a prompt like '❌ ', the buffer length should be 3, but `LengthInBufferCells` returns 2 incorrectly.

daxian-dbw (changes requested):

daxian-dbw:
I will close this PR as it's not the right fix. Please see the discussion in the original PR for more information.

Issue #2504, opened 2021-05-28
Accessibility mode for PSReadLine


SteveL-MSFT:
Related: https://github.com/PowerShell/PSReadLine/issues/859

The re-draw that PSReadLine does is only one aspect that affects accessibility (specifically screen readers that scrape the screen buffer).  I would be interested in what users would like to have in an "Accessibility mode" for PSReadLine beyond the buffer re-draw issue.

daxian-dbw:
I think we need:
- Ask the users who are asking for this support: what features in PSReadLine they would like to have when working with a screen reader.
- Then we go through the asked features and figure out whether each of them makes sense from the user perspective, and whether it's feasible from the implementation perspective (without depending on erasing/re-rendering).

christopherpross:
Hey guys,
I am a screenreader user (NVDA) and would be happy if there would be support for screenreaders.
It would be nice, if the Completition feature in menu style works and the history browsing.

RareBird15:
As a totally blind programmer learning PowerShell and coming from a BASH background, I believe making PSReadLine accessible would significantly enhance the command-line experience for screen reader users. Here are the key features I'd like to see made accessible:

1. Syntax error notifications: Audible or screen reader-friendly notifications of syntax errors as we type.

2. Customizable keybindings: Ability to set and use custom keyboard shortcuts that work well with screen readers.

3. BASH-style completion and history search: Accessible ways to navigate and select from completion options and search command history (e.g., CTRL-R functionality).

4. Undo/redo functionality: Accessible commands to undo and redo actions with clear feedback.

5. Automatic saving and cross-session history: Ability to access and navigate through command history from previous sessions.

6. Autocompletion and command autosuggestion: Screen reader-friendly presentation of autocomplete options and command suggestions.

7. Predictive IntelliSense: Accessible way to view and select from predictive suggestions.

8. Enhanced copy/cut/paste functionality: Improved clipboard operations with clear feedback for screen reader users.

9. Parameter and argument viewing: Accessible method to explore available parameters and arguments for commands.

10. Syntax highlighting: While visual, having an audible or tactile way to convey syntax information would be valuable.

Implementing these features in an accessible manner would greatly improve the PowerShell experience for visually impaired users, bringing it closer to the rich command-line environments we've experienced in other systems like Linux with BASH.

I'm happy to provide more detailed feedback on specific implementations or to test accessibility features as they're developed.

Orinks:
[...]

PR #4854, opened 2025-07-19
Add screen reader support

andyleejordan:
A new render implementation for use under screen readers that avoids redrawing the terminal buffer so as to avoid unnecessary and confusing output.

The differential rendering relies on calculating the common prefix of the `buffer` and `previousRender` strings. Nearly all necessary changes are consolidated in the new rendering function.

Features known not to be supported:

* Colors: as this necessitates redrawing to insert color sequences after the input is received and the AST parsed.
* Inline predictions: as this by definition changes the suffix and thus requires endless redrawing.
* List view predictions: since the render implementation never calls into the prediction engine, this is not available either.
* Menu completion: well, it "works" since it's not disabled and does its own rendering, but no effort has been made to improve `DrawMenu()`, so it's not accessible (and I'm not sure it could be given our current constraints).

Features known to be partially supported:

* Text selection: mark and select commands work as intended, but provide no visual indication.
* Multiple lines: as in newlines work fine, but there is no continuation prompt.
* Visually wrapped lines: editing above a wrapped line redraws all subsequent lines and hence is noisy.
* Status prompt based commands: what-is-key, digit-argument, and most notably, forward/backward incremental history search all render a "status prompt" on the line below the user's input buffer. This _is_ supported; however, it can be noisy since it necessarily has to render the whole buffer when the input buffer changes, including the status prompt (and search text). But what it's reading is almost always going to be relevant.

Everything else should generally work, even Vi mode, and the tests pass. That said, this isn't perfect, and moreover the approach specifically doesn't attempt to enable things from the ground up. There may be features that are available but turn out not to be accessible (like `MenuComplete`) and I believe they should be left as-is.

Specifically tested with NVDA on Windows and VoiceOver on macOS within VS Code's integrated terminal, with shell integration loaded, and Code's screen reader optimizations enabled.

Is a start to resolving #2504.


PSReadLine/Accessibility.cs
```diff
@@ -10,14 +11,73 @@ internal class Accessibility
     {
         internal static bool IsScreenReaderActive()
         {
-            bool returnValue = false;
-
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                return IsAnyWindowsScreenReaderEnabled();           // Check for macOS VoiceOver
```
daxian-dbw: The existing style in this repo always uses curly braces for if blocks, even for the single-statement if blocks. Can you please update the code to adhere to the existing style? Thanks

PSReadLine/Accessibility.cs
```diff
@@ -10,14 +11,73 @@ internal class Accessibility
...
+            // The official way to check for a screen reader on Windows is
+            // SystemParametersInfo(SPI_GETSCREENREADER, ...) but it doesn't
+            // detect the in-box Windows Narrator and is otherwise known to be
+            // problematic.
+            //
+            // Unfortunately, the alternative method used by Electron and
+            // Chromium, where the relevant screen reader libraries (modules)
+            // are checked for does not work in the context of PowerShell
+            // because it relies on those applications injecting themselves into
+            // the app. Which they do not because it's not a windowed app, so
```
daxian-dbw: > Which they do not because it's not a windowed app

Do you mean "PowerShell" is not a windowed app here? It'd be great if you can reword the comments here a little to make it clearer.

PSReadLine/Accessibility.cs
```diff
@@ -10,14 +11,73 @@ internal class Accessibility
...
+            // we're stuck using the known-to-be-buggy way.
+            bool spiScreenReader = false;
+            PlatformWindows.SystemParametersInfo(PlatformWindows.SPI_GETSCREENREADER, 0, ref spiScreenReader, 0);
+            if (spiScreenReader)
+                return true;
+
+            // At least we can correctly check for Windows Narrator using the
+            // NarratorRunning mutex (which is mostly not broken with
+            // PSReadLine, as it were).
```
daxian-dbw: > which is mostly not broken with PSReadLine, as it were

I'm not super sure what you mean by this :) Can you please reword it a bit?

PSReadLine/Accessibility.cs
```diff
@@ -10,14 +11,73 @@ internal class Accessibility
...
+                    Arguments = "read com.apple.universalaccess voiceOverOnOffKey",
+                    UseShellExecute = false,
+                    RedirectStandardOutput = true,
+                    RedirectStandardError = true,
+                    CreateNoWindow = true
+                };
+
+                using Process process = Process.Start(startInfo);
+                process.WaitForExit(250);
+                if (process.ExitCode == 0)
```
daxian-dbw: Nitpick, to avoid an `InvalidOperationException` in case the process doesn't finish within 250ms.
```suggestion
                if (process.HasExited && process.ExitCode == 0)

```

PSReadLine/BasicEditing.cs
```diff
@@ -138,7 +141,8 @@ private static void ForwardDeleteImpl(int endPosition, Action<ConsoleKeyInfo?, o
                         !InViEditMode()));
 
                 buffer.Remove(current, length);
-                _singleton.Render();
+                // Moves cursor to current (backwards probably) and then deletes length
+                _singleton.SafeRender($"\x1b[{length}P", current);
```
daxian-dbw: `ESC [ <n> P` only shifts characters from the current line, not the characters of the next line. So, this may not work for editing multi-line text.

daxian-dbw (commented):
Review in progress. Need to stop here today and want to share my comments so far.

PSReadLine/Accessibility.cs
andyleejordan: It doesn't though, I checked 🫠

PSReadLine/Accessibility.cs
andyleejordan: Correct.

PSReadLine/Accessibility.cs
andyleejordan: Let's add a formatting tool if there's a consistent style we're wishing to apply to this repo. Prettier is a great tool.

PSReadLine/Accessibility.cs
andyleejordan: Windows Narrator is mostly not broken with PSReadLine. Not in the way that NVDA and VoiceOver are.

andyleejordan: It seems to be programmed to ignore duplicate output, the whole re-rendering issue doesn't present itself under Narrator. Which made figuring out "what's wrong" (given no reproduction steps) a tad difficult.

PSReadLine/BasicEditing.cs
andyleejordan: Yeah I should note in the PR description, this has not been tested with multi-line input at all.

PSReadLine/Accessibility.cs
daxian-dbw: > It doesn't though, I checked

My bad :) About this style, it was mixed originally in the repo. After it was transferred to PowerShell org, I followed the style used in PowerShell code ([`csharp_prefer_braces = true:silent`](https://github.com/PowerShell/PowerShell/blob/master/.editorconfig#L134) -- prefer to include braces even for single-statement blocks).

It was only me working in the repo before, so formatting was not a problem :) Totally agree formatting tools will help in this case.

PSReadLine/BasicEditing.cs
```diff
@@ -335,6 +344,7 @@ private bool AcceptLineImpl(bool validate)
 
             if (renderNeeded)
             {
+                // TODO: Evaluate this under a screen reader
                 ForceRender();
```
daxian-dbw: render is needed when there is a status bar showing when user press `enter`, e.g. it's in the `back-i-search` when user press `enter`.

PSReadLine/Cmdlets.cs
```diff
@@ -533,6 +540,8 @@ public object ListPredictionTooltipColor
 
         public bool TerminateOrphanedConsoleApps { get; set; }
 
+        public bool ScreenReader { get; set; }
```
daxian-dbw: The property name doesn't indicate the Boolean type. How about renamed it to be `ScreenReaderIsActive`?

andyleejordan: That seems awkward as a CLI option, you don't usually type `-FooIsActive` to enable something. What about `-ScreenReaderMode`?

PSReadLine/Accessibility.cs
daxian-dbw: Minor change based on our conversation.
```suggestion
            // the app. Which they do not because PowerShell not a windowed app, so
```

PSReadLine/Accessibility.cs
daxian-dbw: Minor update based on the conversation.
```suggestion
            // NarratorRunning mutex. Windows Narrator is mostly not broken with
            // PSReadLine, not in the way that NVDA and VoiceOver are.

```

PSReadLine/Cmdlets.cs
daxian-dbw: The property name in `PSConsoleReadLineOptions` doesn't have to be the same as the parameter in the cmdlet `SetPSReadLineOption`. The former will be the one that is referenced a lot in the code base, so it would be better to make it more readable, such as `UseScreenReader` or `ScreenReaderEnabled`. The parameter name can be made `-EnableScreenReadeMode` or similar.

Should we use a different set of default key bindings when a Screen Reader is turned on? If so, then it may be better to add a new EditMode `ScreenReaderMode`, while defines its own set of default bindings.

PSReadLine/Render.cs
```diff
@@ -218,11 +218,11 @@ private void RenderWithPredictionQueryPaused()
             Render();
         }
 
-        private void Render()
+        private void Render(bool force = false)
         {
             // If there are a bunch of keys queued up, skip rendering if we've rendered very recently.
             long elapsedMs = _lastRenderTime.ElapsedMilliseconds;
-            if (_queuedKeys.Count > 10 && elapsedMs < 50)
+            if (!force && _queuedKeys.Count > 10 && elapsedMs < 50)
```
daxian-dbw: It's possible to have more checks to run when `force is false`, so how about having an outer check on `force` like this:
```c#
if (!force)
{
    // If there are a bunch of keys queued up, skip rendering if we've rendered very recently.
    if (_queuedKeys.Count > 10 && elapsedMs < 50)
    {
        ...
    }

    // If we've rendered very recently, skip the terminal window resizing check as it's unlikely
    // ...
    if (elapsedMs < 50)
    {
        ...
    }
}
```

PSReadLine/Render.cs
```diff
@@ -242,12 +242,128 @@ private void Render()
...
+            }
+
+            // If we had to wrap to render everything, update _initialY
+            var endPoint = ConvertOffsetToPoint(currentBuffer.Length);
+            int physicalLine = endPoint.Y - _initialY;
+            if (_initialY + physicalLine > bufferHeight)
+            {
+                // We had to scroll to render everything, update _initialY.
+                _initialY = bufferHeight - physicalLine;
+            }
```
daxian-dbw: Y is in the range [0, buffer_height), so the check should be `>= bufferHeight`
We should calculate how many physical lines were scrolled up, and move the initial Y up the same number of lines.
```suggestion
            if (endPoint.Y >= bufferHeight)
            {
                // We had to scroll to render everything, update _initialY.
                int scolledLines = endPoint.Y - bufferHeight + endPoint.X is 0 ? 0 : 1;
                _initialY -= scolledLines;
            }
```

daxian-dbw: There is one edge case that I can think of: `endPoint.X == 0 && endPoint.Y == bufferHeight`. In such a case, there is no scrolling yet -- the written-out text exactly fits to the last physical line, and thus when calculating the cursor position, we get (0, bufferHeight). In this case, we shouldn't update `_initialY`.

PSReadLine/Cmdlets.cs
andyleejordan: Let me know which two names you'd like it to be.

I think let's not do a new `EditMode` but simply, for the very few that it seems to affect, disable the function by returning early if in screen reader mode. Otherwise we have to copy/paste the default bindings, and then we'd have different defaults for Windows/non-Windows, or we'd have to add new logic for unbinding by function name instead of key chord.

PSReadLine/Render.cs
andyleejordan: Good catch.

PSReadLine/Render.cs
andyleejordan: Hm, I'd want to test this change a bunch. As is, I'm relying on the existing logic being proven. If we change it here, do we want to change it in `ReallyRender` too?

PSReadLine/Cmdlets.cs
```diff
@@ -153,6 +153,7 @@ public class PSConsoleReadLineOptions
 
         /// <summary>
         /// The predictive suggestion feature is disabled by default.
+        /// TODO: This no longer appears to be true.
         /// </summary>
         public const PredictionSource DefaultPredictionSource = PredictionSource.None;
```
daxian-dbw: Good catch. This constant is no longer used. It can be removed now.

PSReadLine/Render.cs
```diff
@@ -1201,17 +1353,23 @@ internal Point EndOfBufferPosition()
             return ConvertOffsetToPoint(_buffer.Length);
         }
 
-        internal Point ConvertOffsetToPoint(int offset)
+        internal Point ConvertOffsetToPoint(int offset, StringBuilder buffer = null)
         {
+            // This lets us re-use the logic in the screen reader rendering implementation
+            // where the status line is added to the buffer without modifying the locabl state.
```
andyleejordan:
```suggestion
            // where the status line is added to the buffer without modifying the local state.
```

PSReadLine/Render.cs
```diff
@@ -261,7 +413,7 @@ private void ForceRender()
             // and minimize writing more than necessary on the next render.)
 
             var renderLines = new RenderedLineData[logicalLineCount];
-            var renderData = new RenderData {lines = renderLines};
+            var renderData = new RenderData { lines = renderLines };
```
andyleejordan, line 425: Something in VS Code is still sometimes formatting things.

PSReadLine/Render.cs
```diff
@@ -218,36 +218,188 @@ private void RenderWithPredictionQueryPaused()
...
+
+            // Make cursor invisible while we're rendering.
+            _console.CursorVisible = false;
+
+            // Calculate what to render and where to start the rendering.
+            int commonPrefixLength = FindCommonPrefixLength(previousBuffer, currentBuffer);
+
+            // If we're scrolling through history we always want to re-render.
+            // Writing only the diff in this scenario is a weird UX.
+            bool renderDiff = commonPrefixLength > 0 && _anyHistoryCommandCount == 0;
```
andyleejordan: This is slightly updated behavior from the last review. Essentially, say your n-1 command was `git status` and n-2 was `git commit`, the original behavior would cause just `status` to be replaced by `commit` so you hit up-arrow to scroll through history...and you don't get the history item read, just the part that changed. I believe this would not be the desired behavior, so this additional check prevents differential rendering when going through the history.

PSReadLine/Render.cs
```diff
@@ -218,36 +218,188 @@ private void RenderWithPredictionQueryPaused()
...
+            {
+                // No-op, such as when selecting text or otherwise re-entering.
+            }
+            else if (previousBuffer.Length == 0)
+            {
+                // Previous buffer was empty so we just render the current buffer.
+                _console.SetCursorPosition(_initialX, _initialY);
+                _console.Write(currentBuffer);
+            }
+            else if (renderDiff)
```
andyleejordan: Once I had tests in place I felt a lot more comfortable updating the logic here to flow better.

PSReadLine/Render.cs
```diff
@@ -1259,7 +1417,10 @@ private int ConvertLineAndColumnToOffset(Point point)
             int y = _initialY;
 
             int bufferWidth = _console.BufferWidth;
-            var continuationPromptLength = LengthInBufferCells(Options.ContinuationPrompt);
+            var continuationPromptLength = Options.ScreenReaderModeEnabled
+                ? 0
+                : LengthInBufferCells(Options.ContinuationPrompt);
+
```
andyleejordan, line 1431: `ConvertLineAndColumnToOffset` is not used in the screen reader renderer, so it doesn't need to acommodate the modied buffer (where we may have appended the status line prompt). But it _is_ used in some movement functions, so this part needs to stay.

test/ListPredictionTest.cs
```diff
@@ -1355,8 +1373,8 @@ public void List_PluginSource_Acceptance()
                         TokenClassification.Command, "ec",
                         NextLine,
                         TokenClassification.ListPrediction, "<-/3>",
-                        TokenClassification.None, new string(' ', listWidth - 42), // 42 is the length of '<-/3>' plus '<TestPredictor(2) LongNamePredic�(1)>'.
-                        dimmedColors, "<TestPredictor(2) LongNamePredic�(1)>",
+                        TokenClassification.None, new string(' ', listWidth - 42), // 42 is the length of '<-/3>' plus '<TestPredictor(2) LongNamePredic…(1)>'.
```
andyleejordan, line 1376: Yeah, so this is funny. I'm not sure what Unicode character was used in the original file, but due to some encoding issues it seems to need to be updated. Otherwise the tests fail after re-saving this file.

test/MockConsole.cs
```diff
@@ -300,6 +300,17 @@ public virtual void BlankRestOfLine()
                 buffer[writePos + i].ForegroundColor = ForegroundColor;
             }
         }
+        
+        public virtual void BlankRestOfBuffer()
```
andyleejordan, line 304: The mocked version of the `0J` escape sequence used in the screen reader renderer.

test/OptionsTest.cs
```diff
@@ -24,7 +26,7 @@ public void ContinuationPrompt()
                         TokenClassification.None, '}')),
                 _.Ctrl_c,
                 InputAcceptedNow
-                ));
+                ));
```
andyleejordan: Oops, how did trim trailing whitespace not fix this?

test/UnitTestReadLine.cs
```diff
@@ -668,4 +683,14 @@ public fr_FR_Windows(ConsoleFixture fixture, ITestOutputHelper output)
...
     }
+    
+    public class ScreenReader : Test.ReadLine, IClassFixture<ConsoleFixture>
+    {
+        public ScreenReader(ConsoleFixture fixture, ITestOutputHelper output)
+            : base(fixture, output, "en-US", "windows")
+        {
+        }
+        
+        internal override bool ScreenReaderModeEnabled => true;
```
andyleejordan, line 694: The PSReadLine `Option` is specifically not set here in the construtor because the way the tests work, it needs to be set in `TestSetup`, hence this new method (which isn't a field because it's C# and needs to be virtual to be overridden, I guess).

tools/helper.psm1
```diff
@@ -210,7 +210,7 @@ function Start-TestRun
 
     function RunXunitTestsInNewProcess ([string] $Layout, [string] $OperatingSystem)
     {
-        $filter = "FullyQualifiedName~Test.{0}_{1}" -f ($Layout -replace '-','_'), $OperatingSystem
+        $filter = "(FullyQualifiedName~Test.{0}_{1})|(FullyQualifiedName~Test.ScreenReader)" -f ($Layout -replace '-','_'), $OperatingSystem
```
andyleejordan: Silly required work around.

test/KillYankTest.cs
```diff
@@ -672,8 +672,12 @@ public void SelectCommandArgument_CLIArgs()
         [SkippableFact]
         public void SelectCommandArgument_HereStringArgs()
         {
+            Skip.If(ScreenReaderModeEnabled, "We're still investigating exactly why this test fails in screen reader mode.");
```
andyleejordan, line 675: This was the only test I couldn't figure out, and decided not to block on it.

andyleejordan:
When this is merged, please rebase or merge, not squash 🙇

PSReadLine/Render.cs
```diff
@@ -218,36 +218,188 @@ private void RenderWithPredictionQueryPaused()
             Render();
         }
 
-        private void Render()
+        private void Render(bool force = false)
         {
             // If there are a bunch of keys queued up, skip rendering if we've rendered very recently.
             long elapsedMs = _lastRenderTime.ElapsedMilliseconds;
```
daxian-dbw: How about also moving this line into `if (!force)` block?

PSReadLine/Render.cs
daxian-dbw: Small optimization suggestion -- these 2 lines can be moved to the `else` block where the calculated values are used.
```c#
if (currentBuffer == previousBuffer)
{
    ...
}
else if (previousBuffer.Length == 0)
{
    ...
}
else
{
    // Calculate what to render and where to start the rendering.
    int commonPrefixLength = FindCommonPrefixLength(previousBuffer, currentBuffer);

    // If we're scrolling through history we always want to re-render.
    // Writing only the diff in this scenario is a weird UX.
    bool renderDiff = commonPrefixLength > 0 && _anyHistoryCommandCount == 0;

    if (renderDiff)
    {
        ...
    }
    else
    {
        ...
    }
}
```

PSReadLine/Render.cs
```diff
@@ -218,36 +218,188 @@ private void RenderWithPredictionQueryPaused()
...
+            bool renderDiff = commonPrefixLength > 0 && _anyHistoryCommandCount == 0;
+
+            if (currentBuffer == previousBuffer)
+            {
+                // No-op, such as when selecting text or otherwise re-entering.
+            }
+            else if (previousBuffer.Length == 0)
+            {
+                // Previous buffer was empty so we just render the current buffer.
+                _console.SetCursorPosition(_initialX, _initialY);
```
daxian-dbw: If the previous line is empty, do we need to move the cursor?

PSReadLine/Render.cs
```diff
@@ -218,36 +218,188 @@ private void RenderWithPredictionQueryPaused()
...
+            _previousRender = renderData;
+
+            // Calculate the coord to place the cursor for the next input.
+            var point = ConvertOffsetToPoint(_current, buffer);
+
+            if (point.Y == bufferHeight)
+            {
+                // The cursor top exceeds the buffer height and it hasn't already wrapped,
+                // (because we have exactly filled the line) so we need to scroll up the buffer by 1 line.
+                if (point.X == 0 && !currentBuffer.EndsWith("\n"))
```
daxian-dbw: >we have exactly filled the line

I think `currentBuffer.EndsWith("\n")` would always be `false` in this case. We may just call `_console.Write("\n")` without this check..

PSReadLine/Render.cs
```diff
@@ -218,36 +218,188 @@ private void RenderWithPredictionQueryPaused()
...
+                if (commonPrefixLength != previousBuffer.Length)
+                {
+                    // The buffers share a common prefix but the previous buffer has additional content.
+                    // Move cursor to where the difference starts and clear so we can rewrite.
+                    var diffPoint = ConvertOffsetToPoint(commonPrefixLength, buffer);
+                    _console.SetCursorPosition(diffPoint.X, diffPoint.Y);
+                    _console.Write("\x1b[0J");
+                } // Otherwise the previous buffer is a complete prefix and we just write.
+
+                var diffData = currentBuffer.Substring(commonPrefixLength);
```
daxian-dbw: Write directly from the current cursor position when `commonPrefixLength == previousBuffer.Length` may not work in some cases. One example:
- Previous line is `abcdef`, cursor is at the letter `d`
- Press `ctrl+v` to paste `defghi`

In this case, the new buffer is `abcdefghidef`, and `commonPrefixLength == previousBuffer.Length`, but the cursor is still at the first letter `d` when code flow reaches here. Then writting out the new diff `ghidef` will result in `abcghidef` displayed in terminal.

PSReadLine/Render.cs
```diff
@@ -218,36 +218,188 @@ private void RenderWithPredictionQueryPaused()
...
+                _initialY -= scrolledLines;
+            }
+
+            // Preserve the current render data.
+            var renderData = new RenderData
+            {
+                lines = new RenderedLineData[] { new(currentBuffer, isFirstLogicalLine: true) },
+                errorPrompt = (_parseErrors != null && _parseErrors.Length > 0) // Not yet used.
+            };
+            _previousRender = renderData;
```
daxian-dbw: Can we move this down to right before the line 395 `_previousRender.UpdateConsoleInfo(...)`?
The operations around `_previousRender` should be grouped together.

test/MockConsole.cs
```diff
@@ -300,6 +300,17 @@ public virtual void BlankRestOfLine()
...
+        {
+            var writePos = CursorTop * BufferWidth + CursorLeft;
+            for (int i = 0; i < buffer.Length - writePos; i++)
```
daxian-dbw: This new method can be simplified like the following:
```
for (; writePos < buffer.Length; writePos++)
{
    buffer[writePos].UnicodeChar = ' ';
    ...
}
```

test/MockConsole.cs
```diff
@@ -520,6 +533,17 @@ public override void BlankRestOfLine()
             }
         }
 
+        public override void BlankRestOfBuffer()
+        {
+            var writePos = (_offset + CursorTop) * BufferWidth + CursorLeft;
+            for (int i = 0; i < buffer.Length - writePos; i++)
```
daxian-dbw: Same here :)

tools/helper.psm1
daxian-dbw: This mixes the regular tests with the screen reader tests. It's better to separate them out so we have clear view of how many tests get run when screen reader is on and off.

I suggest the following changes in this function:
```suggestion
        $filter = if ($Layout) {
            "FullyQualifiedName~Test.{0}_{1}" -f ($Layout -replace '-','_'), $OperatingSystem
        } else {
            ## Today, tests for screen-reader mode only run on Windows with the 'en-US' layout.
            "FullyQualifiedName~Test.ScreenReader"
        }
```

Then make the following corresponding changes:
- in the `if ($env:APPVEYOR -or $env:TF_BUILD)` block, add `Write-Log "Testing en-US ..."` at the beginning;
- at the end of `if ($IsWindowsEnv)` block, add the following
   ```
   Write-Log "Testing screen reader mode ..."
   RunXunitTestsInNewProcess
   ```

daxian-dbw:
> When this is merged, please rebase or merge, not squash 🙇

@andyleejordan Why not squash? The commit history is not that clean (has changes reverted) and also this repo always do squash merge (like in PowerShell repo).

PSReadLine/Render.cs
andyleejordan: Ah, yes! Thank you. It had been bugging me that I was always calling `FindCommonPrefixLength` and it took the last cleanup commit to make this possible, but I didn't see it.

PSReadLine/Render.cs
andyleejordan: Probably not, I imagine it's always correct after rendering the prompt. Would want to test though.

PSReadLine/Render.cs
andyleejordan: Hm this is tricky. There was a reason for this. I think it was the edge case that you've filled the line and then for the last character _happened_ to insert a newline. The cursor then moved down to (0, +1) and we needed to not repeat inserting a newline.

PSReadLine/Render.cs
andyleejordan: Oh I see what you mean. Thinking about how to check for this in a way that re-renders as little as possible...I think the correct condition here is actually `commonPrefixLength == current`? As in, "from where we're about to write new data, can we just append?" The bug being that it was the checking the common scenario of typing characters one-by-one, not having manually moved the cursor; but in that case, `current == previousBuffer.Length` so it'll operate the same.

andyleejordan:
> > When this is merged, please rebase or merge, not squash 🙇
> 
> @andyleejordan Why not squash? The commit history is not that clean (has changes reverted) and also this repo always do squash merge (like in PowerShell repo).

Because one single commit is going to be way too big, and the commits _are_ clean, where reverts/changes logically follow new information. (Such as testing the DLL-loaded approach for screen readers.)

PSReadLine/Render.cs
andyleejordan: In fact, this is the exact change we figured out debugging together in 14fc559ff9be75031f0bc190b70ef3a7c6ee41c8

andyleejordan: Tested this: yes that edge case still requires this check.

daxian-dbw: Got it. If there is a newline at the end of input, the scrolling up should've happened. That scenario should be handled in the `if (endPoint.Y >= bufferHeight)` block at line 355, but it's not.

I think we need to take care of that scenario there, by adding `!currentBuffer.EndsWith("\n")` to the `if (endPoint.X == 0)` check at line 360. Note that this edge case could happen with `endPoint.Y > bufferHeight` at line 355.

```
358    // We had to scroll to render everything, update _initialY.
359    int offset = 1; // Base case to handle zero-indexing.
360    if (endPoint.X == 0 && !currentBuffer.EndsWith("\n"))
361    {
362        // The line hasn't actually wrapped yet because we have exactly filled the line.
363        offset -= 1;
364    }
```

Then here, we can just call `_console.Write("\n")` without the check.

PSReadLine/Render.cs
```diff
@@ -218,36 +218,188 @@ private void RenderWithPredictionQueryPaused()
...
+                _console.SetCursorPosition(_initialX, _initialY);
+                _console.Write("\x1b[0J");
+                _console.Write(currentBuffer);
+            }
+
+            // If we had to wrap to render everything, update _initialY
+            var endPoint = ConvertOffsetToPoint(currentBuffer.Length, buffer);
+            if (endPoint.Y >= bufferHeight)
+            {
+
```
daxian-dbw: Remove extra line
```suggestion
            {
```

PSReadLine/Render.cs
```diff
@@ -218,36 +218,188 @@ private void RenderWithPredictionQueryPaused()
...
+                // We had to scroll to render everything, update _initialY.
+                int offset = 1; // Base case to handle zero-indexing.
+                if (endPoint.X == 0)
```
daxian-dbw: As is discussed in the comment thread below: https://github.com/PowerShell/PSReadLine/pull/4854#discussion_r2294304909
```suggestion
                if (endPoint.X == 0 && !currentBuffer.EndsWith("\n"))
```

PSReadLine/Render.cs
daxian-dbw: When execution reaches here, `_current` is already changed -- it no longer reflects where the cursor is at. This uncommon scenario is tricky, and it potentially could affect the `commonPrefixLength != previousBuffer.Length` case.

I suggest we leave it aside for now. The current logic works well for the most common editing scenarios. Also, after integrating with VSCode terminal, it has another layer of accessibility feature to make it works as expected to end users.

We just need to be aware of this potential problem. I will open an issue after we merged this PR.

PSReadLine/Render.Helper.cs
```diff
@@ -96,6 +96,7 @@ internal static int LengthInBufferCells(char c)
             if (c < 256)
             {
                 // We render ^C for Ctrl+C, so return 2 for control characters
+                // TODO: Do we care about this under a screen reader?
```
daxian-dbw: I guess `^C` won't be read out? But it still takes up 2 cells even when screen reader is on.

andyleejordan: I checked, and the place where `^C` is specifically rendered as "^" plus "C" is `CancelLine` which is also used in screen reader mode (verified it shows up as a red ^C) so yes this part of `LengthInBufferCells` is still correct. TODO answered 😄

PSReadLine/Render.cs
andyleejordan: We really just need to get what PRs are open in, and then setup automatic code formatters so we're not wasting time with formatting manually. Especially in VS Code where Code Lens make it look like there are more or less lines than there actually are.

PSReadLine/Render.cs
andyleejordan: It's where it is because `ReallyRender` specifically saves `_previousRender` before calculating the offsets, and this was written to be a simplified version of that function. Do you want it moved in both functions?

test/MockConsole.cs
andyleejordan: I think keeping it to the same pattern as `BlankRestOfLine()` et. al. makes more sense, but sure.

tools/helper.psm1
andyleejordan: Honestly I have some work-in-progress to remove _all_ this layout logic so I took the simple approach here, but I can do this for now until that's in.

PSReadLine/Render.cs
andyleejordan: I'll leave a note about the known bug.

PSReadLine/Render.cs
daxian-dbw: It looks safe to move `_previousRender = renderData;` in `ReallyRender` too.

test/MockConsole.cs
daxian-dbw: The `BlankRestOfLine()` is different -- it compares with `BufferWidth - CursorLeft` in the loop, but this method compares with `buffer.Length - writePos;`. So, the old pattern becomes inefficient.

daxian-dbw:
@andyleejordan can you please simply push new changes instead of force pushing all commits? It makes it hard for me to continue the view as I cannot view the diffs between a commit that I have reviewed and the subsequent new changes.

> Because one single commit is going to be way too big, and the commits are clean, where reverts/changes logically follow new information. (Such as testing the DLL-loaded approach for screen readers.)

If the cost of keeping the commit history clean is to always force push all commits, then I'm afraid I disagree this is the way to go :) Since the PR already keeps all information, I think a squash merge should be fine.

tools/helper.psm1
daxian-dbw: Even if the layout logic is removed, I think we will still want to separate the 2 test runs -- one for regular testing and one for screen reader mode testing. In this way, it's easy to view how many tests are run for those 2 scenarios respectively, what tests are skipped in each of them, and etc.

daxian-dbw: > in the `if ($env:APPVEYOR -or $env:TF_BUILD)` block, add `Write-Log "Testing en-US ..."` at the beginning;

Can you also make this change please? It adds "Testing en-US ..." similarly to the highlighted caption in the screenshot below.

It would be even better if you can prepend a new line to both those 2 messages, so as to make them easier to be noticed.

image

andyleejordan: I'm not sure I follow, this is just CI (that's pending replacement). Running them as separate instances honestly makes them even harder to follow in that log. With the way I did this using a separate fixture this is how they look:
Screenshot 2025-08-26 at 11 37 10 AM


We're spending a lot of time in this PR on a system we've already identified as non-compliant and needing to be updated in another.

Like you certainly _could_ go update `RunXunitTestsInNewProcess` to have a log statement like you're desiring (instead of there sometimes being a log statement before it, and sometimes not, because old legacy code), but it'll get deleted as soon as we're done with this.

daxian-dbw: What I'm trying to say is, when running tests in CI or locally from command line, the test suites for `<keyboard_layout>-Windows` and `ScreenReader` should run separately, so that instead of a combined results output like in the following screenshot, we can view the specific result for each. It makes it easy to view the detailed numbers/status about each test run scenario.

image

----

> We're spending a lot of time in this PR on a system we've already identified as non-compliant and needing to be updated in another.

I'm happy to see the CI and build get modernized. But before that happens, we are still making changes based on what we have today. Since we need to run `RunXunitTestsInNewProcess` twice now, having a caption like `"Testing en-US ..."` and `Testing screen reader mode ...` helps to identify what is what when looking at the CI output.

andyleejordan:
@daxian-dbw in order to facilitate getting this PR in with your requirement to squash PRs, this is now dependent on the separate but necessary PRs:

* https://github.com/PowerShell/PSReadLine/pull/4909
* https://github.com/PowerShell/PSReadLine/pull/4910
* https://github.com/PowerShell/PSReadLine/pull/4911

I finished up and tested your last requests above, and will rebase on master per this project's documentation after those dependent PRs are merged. Leaving this as a draft so it doesn't accidentally get merged before then. Thanks for your help!

tools/helper.psm1
```diff
@@ -210,7 +210,14 @@ function Start-TestRun
 
     function RunXunitTestsInNewProcess ([string] $Layout, [string] $OperatingSystem)
     {
-        $filter = "FullyQualifiedName~Test.{0}_{1}" -f ($Layout -replace '-','_'), $OperatingSystem
+        $filter = if ($Layout) {
+            Write-Log "Testing $Layout on $OperatingSystem...`n"
```
daxian-dbw, line 214: This cleaner. Thanks!

PSReadLine/Render.cs
daxian-dbw: Issue opened to track this: https://github.com/PowerShell/PSReadLine/issues/4912

daxian-dbw (approved):
LGTM.

@andyleejordan Regarding PR reviews:
1. It's most efficient to only push new changes and squash in the end, because for each iteration
    - the author can just focus on addressing feedback.
    - the reviewers can just focus on the new diffs.
2. Commit history is preserved in the PR, along with all discussions.
3. This is how we do PR reviews in this repo, in PowerShell repo, and most repos out there I believe.

However, since you've been insisting on it so much, I will merge this PR with a merge node as you expect.

andyleejordan:
Thanks Dongbo!

PR #4862, opened 2025-07-22
Fix that ViFindBrace doesn't search for brace when current char is not a brace

sharpchen:
# PR Summary

Vi searches right pair of first left brace, or first right brace when current char is not a brace. This pr added this functionality.

After the update, `ViGotoBrace` should work like this(of course this will also affect other methods such as `ViDeleteBrace`):
foo

## PR Checklist

- [x] Summarized changes
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [ ] Doc Issue filed: 

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/PowerShell/PSReadLine/pull/4862)

### DIFF
--- a/PSReadLine/Movement.vi.cs
+++ b/PSReadLine/Movement.vi.cs
@@ -254,7 +254,28 @@ namespace Microsoft.PowerShell
                 case ')':
                     return ViFindBackward(i, '(', withoutPassing: ')');
                 default:
-                    return i;
+                    ReadOnlySpan<char> parenthese = stackalloc char[] { '{', '}', '(', ')', '[', ']' };
+                    int nextParen = i;
+                    // find next of any kind of paren
+                    for (; nextParen < _buffer.Length; nextParen++)
+                        for (int idx = 0; idx < parenthese.Length; idx++)
+                            if (parenthese[idx] == _buffer[nextParen]) goto Outer;
+
+                Outer:
+                    int match = _buffer[nextParen] switch
+                    {
+                        // if next is opening, find forward
+                        '{' => ViFindForward(nextParen, '}', withoutPassing: '{'),
+                        '[' => ViFindForward(nextParen, ']', withoutPassing: '['),
+                        '(' => ViFindForward(nextParen, ')', withoutPassing: '('),
+                        // if next is closing, find backward
+                        '}' => ViFindBackward(nextParen, '{', withoutPassing: '}'),
+                        ']' => ViFindBackward(nextParen, '[', withoutPassing: ']'),
+                        ')' => ViFindBackward(nextParen, '(', withoutPassing: ')'),
+                        _ => nextParen
+                    };
+
+                    return match == nextParen ? i : match;
             }
         }
 
--- a/PSReadLine/Movement.vi.cs
+++ b/PSReadLine/Movement.vi.cs
@@ -254,12 +254,16 @@ namespace Microsoft.PowerShell
                 case ')':
                     return ViFindBackward(i, '(', withoutPassing: ')');
                 default:
-                    ReadOnlySpan<char> parenthese = stackalloc char[] { '{', '}', '(', ')', '[', ']' };
+                    ReadOnlySpan<char> parentheses = stackalloc char[] { '{', '}', '(', ')', '[', ']' };
                     int nextParen = i;
                     // find next of any kind of paren
                     for (; nextParen < _buffer.Length; nextParen++)
-                        for (int idx = 0; idx < parenthese.Length; idx++)
-                            if (parenthese[idx] == _buffer[nextParen]) goto Outer;
+                        for (int idx = 0; idx < parentheses.Length; idx++)
+                            if (parentheses[idx] == _buffer[nextParen]) goto Outer;
+
+                    // if not found, nextParen could exceed the range
+                    if (nextParen >= _buffer.Length)
+                        return i;
 
                 Outer:
                     int match = _buffer[nextParen] switch
--- a/test/MovementTest.VI.cs
+++ b/test/MovementTest.VI.cs
@@ -424,6 +424,11 @@ namespace Test
         [SkippableFact]
         public void ViGotoBrace()
         {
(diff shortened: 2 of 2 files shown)


ghost:
S

andyleejordan:
You are right that's how it works. I'm hoping we can do two things: verify the order of precedence in vanilla Vi (I was playing around trying to determine it and couldn't make heads or tails of it), and add some unit tests. Thanks! Love the fix though.

sharpchen:
> verify the order of precedence in vanilla Vi

Not sure what it means? If you mean the standard behavior, [`:h %`](https://vimhelp.org/motion.txt.html#%25) documented the expected behavior in vim:

> Find the next item in this line after or under the cursor and jump to its match. 

And [`:h cpo-%`](https://vimhelp.org/options.txt.html#cpo-%25) described how the Vi behavior differs the way of Vim enhanced:
> Parens inside single and double quotes are also counted, causing a string that contains a paren to disturb the matching.  For example, in a line like "if (strcmp("foo(", s))" the first paren does not match the last one. 

It indeed requires some tests. I did find at least one inconsistent behavior.

sharpchen:
Hi @andyleejordan. I've rewrote the implementation, I think it's now good to go.
But I fail to run the test using `./build.ps1 -Test`, it seems `test/MovementTest.VI.cs` is ignored? Tests pass even on false conditions.

PSReadLine/Movement.vi.cs
```diff
@@ -254,7 +254,28 @@ private int ViFindBrace(int i)
                 case ')':
                     return ViFindBackward(i, '(', withoutPassing: ')');
                 default:
-                    return i;
+                    ReadOnlySpan<char> parenthese = stackalloc char[] { '{', '}', '(', ')', '[', ']' };
+                    int nextParen = i;
+                    // find next of any kind of paren
+                    for (; nextParen < _buffer.Length; nextParen++)
+                        for (int idx = 0; idx < parenthese.Length; idx++)
+                            if (parenthese[idx] == _buffer[nextParen]) goto Outer;
```
andyleejordan: The rare great use of `goto`

andyleejordan:
I'm not sure but if this build is like others (looking at PSScriptAnalyzer right now), `-Test` won't _rebuild_ so you have to run `./build.ps1` after any change and then follow it with `./build.ps1 -Test`.

AppVeyor is at least running tests, looks like `Test.en_US_Windows.ViGotoBrace` and `Test.ScreenReader.ViGotoBrace` need to be checked and updated.

sharpchen:
Looks like tests are only available on Windows.

> Shouldn't call ReadKey when there are no more keys

I honestly have no idea why test's failing. There's is a input but claims no key.

andyleejordan:
> Looks like tests are only available on Windows.

I forgot about that but yes this is unfortunately true.

andyleejordan:
@sharpchen I will get this built on Windows and see if I can't fix the test so we can get this in!

andyleejordan:
Claude Opus 4.6 was actualy very adapt at fixing this error (it helps that the test was failing for the same reason as other similar tests, which then already had a fix it applied).

andyleejordan (approved):
LGTM let's get it in @daxian-dbw

PSReadLine/Movement.vi.cs
```diff
@@ -254,7 +254,28 @@ private int ViFindBrace(int i)
...
+
+                Outer:
```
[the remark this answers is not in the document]
daxian-dbw: > In the default case, if no brace/paren/bracket is found from the current position to the end of the buffer, `nextParen` will end up equal to `_buffer.Length`, and `_buffer[nextParen]` will throw `IndexOutOfRangeException`

This is a real problem. I prefer to not using the `Outer` label here. For the `goto` statement, we should avoid using it unless it's really needed.

PSReadLine/Movement.vi.cs
```diff
@@ -254,7 +254,28 @@ private int ViFindBrace(int i)
                 case ')':
                     return ViFindBackward(i, '(', withoutPassing: ')');
                 default:
-                    return i;
+                    ReadOnlySpan<char> parenthese = stackalloc char[] { '{', '}', '(', ')', '[', ']' };
+                    int nextParen = i;
+                    // find next of any kind of paren
+                    for (; nextParen < _buffer.Length; nextParen++)
+                        for (int idx = 0; idx < parenthese.Length; idx++)
```
daxian-dbw: We can find the next parenthesis by the code below, instead of using nested loop here:
```c#
ReadOnlySpan<char> parens = stackalloc char[] { '{', '}', '(', ')', '[', ']' };

// Find next parenthesis of any type
int idx = _buffer.AsSpan(i).IndexOfAny(parens);
if (idx < 0)
{
    // No parenthesis, so nothing to match
    return i;
}

int nextParen = i + rel;
int match = _buffer[nextParen] switch
{
    ...
}

return match == nextParen ? i : match;
```

daxian-dbw (changes requested):

PSReadLine/Movement.vi.cs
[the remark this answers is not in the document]
daxian-dbw: We can avoid the nested loop all together.

PSReadLine/Movement.vi.cs
sharpchen: There's no `AsSpan` extension for `StringBuilder` though, did I miss something?
If that's true I think `goto` is succinct choice here to avoid extra boolean flag, we just need one more guard for the implicit increment on exit.

daxian-dbw: Right, I incorrectly thought `_buffer` was a string.
Then I'm fine keeping the `goto`. You just need to add an early return before `int match = _buffer[nextParen] switch`.

daxian-dbw (approved):

andyleejordan:
Thanks @sharpchen and @daxian-dbw!

rashworld-max (approved):

PR #4909, opened 2025-08-27
Fix ellipsis / continuation character encoding issue in tests

andyleejordan:
This character was always meant to be an ellipsis. I'm unsure exactly why Visual Studio interprets it so, but VS Code and GitHub do not. This commit replaces it with the actual Unicode character which ensures the tests continue to pass when these files are edited and saved in VS Code.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/PowerShell/PSReadLine/pull/4909)


daxian-dbw:
This PR was closed by GitHub, because, I guess, the title starts with 'Fix'.

PR #4910, opened 2025-08-27
Fix `IsMutexPresent()` to avoid incorrect use of `GetLastWin32Error()`

andyleejordan:
Since we're not actually making a system call.

### DIFF
--- a/PSReadLine/PlatformWindows.cs
+++ b/PSReadLine/PlatformWindows.cs
@@ -8,6 +8,7 @@ using System.ComponentModel;
 using System.Diagnostics;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Threading;
 using Microsoft.PowerShell;
 using Microsoft.PowerShell.Internal;
 using Microsoft.Win32.SafeHandles;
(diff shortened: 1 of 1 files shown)


PSReadLine/PlatformWindows.cs
```diff
@@ -79,6 +79,21 @@ IntPtr templateFileWin32Handle
...
     internal static extern IntPtr GetStdHandle(uint handleId);
 
+    internal const int ERROR_ALREADY_EXISTS = 0xB7;
+
+    internal static bool IsMutexPresent(string name)
+    {
+        try
+        {
+            using var mutex = new System.Threading.Mutex(false, name);
+            return Marshal.GetLastWin32Error() == ERROR_ALREADY_EXISTS;
```
andyleejordan: @SeeminglyScience astutely pointed out this is bad, as we're not exactly making a system call when gettingn the mutex. It should be `TryOpenExisting`.

andyleejordan: And fixed. Thanks!

daxian-dbw: Sorry, I just saw this after merging https://github.com/PowerShell/PSReadLine/pull/4854. Please submit a new PR for the fix.

andyleejordan: I already updated this one 😄

daxian-dbw (approved):

Issue #5101, opened 2026-03-08
BackwardKillLine triggering an error


Snowy117:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the [latest released version](https://www.powershellgallery.com/packages/PSReadLine)
- [x] Search the existing issues, especially the pinned issues.

### Exception report

```console
Oops, something went wrong...
Please report this bug with ALL the details below, including both the 'Environment' and 'Exception' sections.
Please report on GitHub: https://github.com/PowerShell/PSReadLine/issues/new?template=Bug_Report.yaml
Thank you!

### Environment
PSReadLine: 2.4.5
PowerShell: 7.5.4
OS: Microsoft Windows 10.0.26200
BufferWidth: 120
BufferHeight: 32

Last 6 Keys:

 Ctrl+v Enter
 UpArrow LeftArrow UpArrow Ctrl+u

### Exception

System.ArgumentOutOfRangeException: Index and length must refer to a location within the string. (Parameter 'length')
   at System.Text.StringBuilder.ToString(Int32 startIndex, Int32 length)
   at Microsoft.PowerShell.PSConsoleReadLine.Kill(Int32 start, Int32 length, Boolean prepend)
   at Microsoft.PowerShell.PSConsoleReadLine.BackwardKillLine(Nullable`1 key, Object arg)
   at Microsoft.PowerShell.PSConsoleReadLine.ProcessOneKey(PSKeyInfo key, Dictionary`2 dispatchTable, Boolean ignoreIfNoAction, Object arg)
   at Microsoft.PowerShell.PSConsoleReadLine.InputLoop()
   at Microsoft.PowerShell.PSConsoleReadLine.ReadLine(Runspace runspace, EngineIntrinsics engineIntrinsics, CancellationToken cancellationToken, Nullable`1 lastRunStatus)
```

### Screenshot

<img width="1710" height="846" alt="Image" src="https://github.com/user-attachments/assets/d57ae1e0-36b3-4dcb-9e56-8d1ff895c338" />

### Environment data

```console
PS Version: 7.5.4
PS HostName: ConsoleHost
PSReadLine Version: 2.4.5
PSReadLine EditMode: Emacs
OS: 10.0.26100.7306 (WinBuild.160101.0800)
BufferWidth: 120
BufferHeight: 32
```

### Steps to reproduce

1. Make your clipboard as:
```text
Set-PSReadLineOption -EditMode Emacs
[...]

PR #5102, opened 2026-03-08
Fix BackwardKillLine length calculation for multiline input

Snowy117:
## Summary
- fix BackwardKillLine to delete only from current logical line start to cursor (length = current - start)
- add a regression test in KillYankTest.BackwardKillLine for multiline input to prevent ArgumentOutOfRangeException
- addresses issue #5101


Snowy117:

andyleejordan (approved):
Wow, thanks for the fix. Verified the second argument is `length` so yes, the start definitely needs to be subtracted.

